PR #215, opened 2014-08-29
SPARKC-43: (previously #37) Create separate spark RDD partition for each key combintation 

artem-aliev:
included in request
Calculate TOKEN value for keys and suggest appropriate node to support data locality
Fixes #37.


bcantoni:
Superseded by #746

PR #475, opened 2014-12-10
wip-467: Moved several connection settings from system properties to Spa...

### LINKED ISSUE #467: Setting connector's system properties on the executors
I'm trying to implement the advice @jacek-lewandowski gave me on this discussion #465, but I'm having some difficult time finding the right way to set a system-wide property like 'spark.cassandra.connection.keep_alive_ms' on the executors.

I think that the spark way would be to use 'spark.executor.extraJavaOptions', like this:
```bash
spark.executor.extraJavaOptions = "-Dspark.cassandra.connection.keep_alive_ms=12000"
```
but it's illegal in Spark to set a property that starts by "-Dspark." [1]

I also tried the following, but without effect:
```bash
spark.executorEnv.spark.cassandra.connection.keep_alive_ms=12000
```
What would be the recommended way to setting these system-wide properties?

(not on the spark-conf file: we run on Mesos and use specific configuration for each job) 

-kr, Gerard.

[1] https://github.com/apache/spark/blob/master/core%2Fsrc%2Fmain%2Fscala%2Forg%2Fapache%2Fspark%2FSparkConf.scala#L269

jacek-lewandowski:
...rk configuration and renamed spark.cassandra.connection.keep_alive_ms property

spark.cassandra.connection.keep_alive_ms has been renamed to cassandra.connection.keep_alive_ms because one could not pass this setting with `-Dprop=value` otherwise.

Fixes #467

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 1.1.1 (unreleased)
  * Merged bug fix for #463 from b1.0
+ * Fixed the problem with passing additional connection settings to executors (#467)
 
 1.1.0
  * Switch to java driver 2.1.3 and Guava 14.0.1 (yay!).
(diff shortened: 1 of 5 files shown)


pkolaczk:
First, do we really need this in 1.1? Looks like there is an API change (name of the parameter) which could break some things. @jacek-lewandowski Can you rebase it to master?

pkolaczk:
There are no global connection parameters in master now, and this is a too big change to include in 1.1.x.

PR #488, opened 2014-12-17
wip-442: Token aware writes

jacek-lewandowski:


pkolaczk:
After thinking of it more:

1. I really don't like the fact we have different configs for writes and reads. This is IMHO wrong level to deal with this. There should be *one* config and the differentiation between reads and writes should happen at a lower level.

2. The PriorityQueue / Map could be better and I know how to do this. I'm little afraid the high worst-case complexity and high object allocation rate will bite us, and this code is on the critical path for writes. Hold on, I create another PR based on this one.

Other than those two major things, it is ok.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/Batch.scala
```diff
@@ -0,0 +1,76 @@
...
+  def add(stmt: BoundStatement): Unit
+
+  /** Returns `true` if the collected statements exceed the desired size of this batch. Rows based and
+    * bytes based implementations compute it in different ways. */
+  def isSizeExceeded: Boolean
+
+  def statements: Seq[BoundStatement]
+
+  /** Removes all the collected statements and resets this batch to the initial state. */
+  def reuse(): Unit
```
jacek-lewandowski: ok

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/Batch.scala
```diff
@@ -0,0 +1,76 @@
...
+      case BytesInBatch(bytes) => new SizeLimitedBatch(bytes)
+    }
+  }
+}
+
+private[writer] class RowLimitedBatch(val maxRows: Int) extends Batch {
+  val buf = new ArrayBuffer[BoundStatement](maxRows)
+
+  def add(stmt: BoundStatement): Unit = {
+    buf += stmt
```
jacek-lewandowski: This limit is not intended to be strict. This is only a hint. However, I found that I should allocate `maxRows+1` for `buf`.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/Batch.scala
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  override def isSizeExceeded: Boolean = buf.size > maxRows
+
+  override def compare(that: Batch): Int = that match {
+    case thatBatch: RowLimitedBatch => buf.size.compare(thatBatch.maxRows)
+    case _ => throw new ClassCastException("Not a RowLimitedBatch")
```
jacek-lewandowski: Ok, I'll refactor it a bit

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/Batch.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  override def reuse(): Unit = buf.clear()
+}
+
+private[writer] class SizeLimitedBatch(val maxBytes: Int) extends Batch {
+  val buf = new ArrayBuffer[BoundStatement](10)
+  var size = 0
+
+  def add(stmt: BoundStatement): Unit = {
+    buf += stmt
```
jacek-lewandowski: Same as above

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/GroupingBatchBuilder.scala
```diff
@@ -0,0 +1,227 @@
...
+        startPos = normalize(startPos + 1)
+        curSize -= 1
+        result
+      } else {
+        throw new NoSuchElementException("The queue is empty")
+      }
+    }
+
+    /** Adds a value the end of the queue and then moves it toward the head if needed. */
+    def add(value: V): Box = {
```
jacek-lewandowski: ok

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/GroupingBatchBuilder.scala
```diff
@@ -0,0 +1,227 @@
...
+    @inline
+    private[this] def normalize(x: Int): Int =
+      if (x < 0)
+        capacity + (x % capacity)
+      else
+        x % capacity
+
+    @tailrec
+    private[this] def upliftIfNeeded(realIdx: Int): Box = {
+      normalize(realIdx) match {
```
jacek-lewandowski: ok

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/GroupingBatchBuilder.scala
```diff
@@ -0,0 +1,227 @@
...
+
+    /** Updates a box with the provided value. Then, it moves the box towards the
+      * head of the queue if needed. Box indexes are updated accordingly. */
+    def update(box: Box): Box = {
+      val b = setUnchecked(normalize(box._1), box._2)
+      upliftIfNeeded(b._1)
+    }
+
+    /** Removes and returns the first element of the queue. */
+    def remove(): Box = {
```
jacek-lewandowski: ok

jacek-lewandowski:
The case you mentioned is in fact quite probable. But - in such the case, using batches is pointless, isn't it? For different jobs we need different configurations. If the user wants to write long rows, and he has very few partitions, he will create buffer of the proper size - enough to store all partitions used in a single write job. On the other hand, if the user has much more partitions, he may want to change the batching level from partition to host - less efficient, but still he can benefit somehow. When the buffer size is not adjusted to the particular job, it will not be efficient anyway. Otherwise, the new batch will be added to the end or quite close to the end. 

I'm just saying that you've implemented an excellent general purpose PriorityHashMap and integrated it with this PR - I suggest contributing it to Scala :). However, I don't agree that my solution is far worse than your one for the purpose it was designed. I measured it a little bit for the case where batches has 5 elements on average and see no significant difference. 

For the configuration - as we talked - we should discuss it more and eventually address it in a separate ticket.

pkolaczk:
I was not saying your solution was far worse :). I actually liked it very much. After longer thinking of it I was just worried about some edge cases. I know this could be well tuned. But I also want our software to behave in a predictable ways, even if not perfectly tuned (e.g. the user doesn't know what he's doing). Most of the users start with default settings, and batch level "per partition" was the default one.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/LocalNodeFirstLoadBalancingPolicy.scala
```diff
@@ -1,21 +1,27 @@
...
+import com.datastax.driver.core._
 import java.net.{InetAddress, NetworkInterface}
+
 import scala.collection.JavaConversions._
 import scala.util.Random
 
 import org.apache.spark.Logging
 
 /** Selects local node first and then nodes in local DC in random order. Never selects nodes from other DCs. */
-class LocalNodeFirstLoadBalancingPolicy(contactPoints: Set[InetAddress], localDC: Option[String] = None) extends LoadBalancingPolicy with Logging {
```
pkolaczk: Would be nice to reflect token-awareness in ScalaDoc here.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/GroupingBatchBuilder.scala
```diff
@@ -0,0 +1,122 @@
...
+      None
+    } else {
+      batchMap.remove(batchKey)
+      Some(batch)
+    }
+  }
+
+  private def addBatchToMap(batchKey: Any, newStatement: BoundStatement): Option[Batch] = {
+    if (batchMap.size == maxBatches) {
+      Some(batchMap.dequeue())
```
pkolaczk: newStatement is ignored if this branch is taken; isn't it a bug?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/GroupingBatchBuilder.scala
```diff
@@ -0,0 +1,122 @@
...
+  final override def computeNext(): Statement = {
+    if (lastStatement == null && data.hasNext) {
+      lastStatement = batchStatementBuilder.bind(data.next())
+      lastKey = batchKeyGenerator(lastStatement)
+    }
+
+    if (lastStatement != null) {
+      processStatement(lastKey, lastStatement) match {
+        case Some(batch) =>
+          createStmtAndReleaseBatch(batch)
```
pkolaczk: This code is quite non-obvious and I remember the previous one was much easier to understand. E.g. it is quite confusing for me that `processStatement` may fail and not add the statement, and the method `createStmtAndReleaseBatch` doesn't do anything with the not-added statement, yet the statement processing is (silently) postponed to the next iteration by *not setting* `lastStatement` to null. So it requires me to actually analyze not just what was written, but also what was *not written*.

I wonder, why can't we do it similar to as it was before:
1. grab the row from `data` and create the statement
2. create a batch key
3. find a batch or create a new one
4. add a statement to the batch
5. update priority map (call put)
6. the batch is full (exceeded) after addition? remove from the `PriorityMap` and return it
7. the number of batches is exceeded? take the first one from the `PriorityMap` and return it
8. nothing to return? continue `computeNext` tail call

pkolaczk:
+1

pkolaczk:
All tests green.

PR #501, opened 2015-01-02
SPARKC-5: PriorityHashMap for prioritizing batches.

pkolaczk:


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/PriorityHashMap.scala
```diff
@@ -0,0 +1,330 @@
...
+    find(key, hash(key))
+  }
+
+  /** Finds a key in the indexes array, starting at a given position in the indexes array
+    * Returns a position in the indexes array pointing to the found key or a position of
+    * the first empty index entry, if key was not found. */
+  @tailrec
+  private def find(key: K, pos: Int): Int = {
+    val i = _indexes(pos)
+    if (i < 0 || _keys(i) == key) pos
```
jacek-lewandowski: What do you think about comparing real hash codes of keys before comparing the keys?

pkolaczk: It would only make sense if we memorized key hashes. Otherwise, computing hashes for both keys is probably just as costly as comparing objects directly.

jacek-lewandowski: It is advised to use immutable classes for keys so precomputed hashes can be expected.

pkolaczk: Yes, but I wonder if it is worth the effort. It will require one additional array for storing hashes (of size capacity * 2), plus extra calls to save hashes (probably negligible) plus an extra compare + conditional jump in find() before comparing the key; which depending on the circumstances can be anything between ~2-4 CPU cycles (L1 cache hit,  branch predicted) through ~7-15 CPU cycles (cache hit, branch mispredicted) to 1000+ CPU cycles worstcase (L2 cache miss when referencing the hashcodes array; quite improbable though considering our size of the queue). Most keys are found *in the first iteration*, so most of the time you'd just pay that extra cost for nothing, because you have to finally compare key value anyway. Actually, in one of the first versions I had a bug, where capacity of the hashtable was incorrectly 4 times bigger than should be, and the test coverage showed that this condition in line 93 was never false during the whole test suite (which led me to discover that bug with sizing, and added another test for that). 

BTW: There is one bigger thing I found by profiling that influences performance - there are some wasteful calls to `find` in `siftDown` and `siftUp`, because the same key is being moved ("bubble" like) over and over again, instead of moving it only in the last step when its final position is known.

PR #504, opened 2015-01-10
Added support for InSet predicate

granturing:
Spark 1.1.1 introduced a new InSet expression which is generated by the optimizer for "IN" filters where the values are literals. This PR was originally made for Spark 1.1.1 but is valid for Spark 1.2 as well. I've tested this on both Spark 1.1.1 and Spark 1.2, using the most recent commits. I added a simple integration test to ensure the predicate is pushed down as part of the CassandraTableScan versus SparkSQL filter.


pkolaczk:
+1

PR #505, opened 2015-01-12
SPARKC-2: Efficient grouping of Cassandra rows by partition key. Useful ...

pkolaczk:
...for working with "wide rows", e.g. time-series data.

This commit introduces new spanBy and spanByKey operations. For data coming out of Cassandra, they can be used as faster equivalent of groupBy / groupByKey, because they take advantage of the fact that data coming out from Cassandra is already grouped by partition key and ordered by clustering columns. Therefore, costly data shuffling can be avoided and it is possible to group several tens of millions of objects per second on a single core. However, those methods are not general replacements for groupBy, as they rely on data order to work correctly. This commit addresses only Scala API.

### DIFF
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/CassandraRDDSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/CassandraRDDSpec.scala
@@ -18,6 +18,7 @@ import scala.reflect.runtime.universe.typeTag
 case class KeyValue(key: Int, group: Long, value: String)
 case class KeyValueWithConversion(key: String, group: Int, value: Long)
 case class CustomerId(id: String)
+case class Key(key: Int)
 case class KeyGroup(key: Int, group: Int)
 case class Value(value: String)
 case class WriteTimeClass(id: Int, value: String, writeTimeOfValue: Long)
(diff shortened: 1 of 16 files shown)


spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/CassandraRDDSpec.scala
```diff
@@ -394,6 +395,41 @@ class CassandraRDDSpec extends FlatSpec with Matchers with SharedEmbeddedCassand
...
+
+    val results = sc
+      .cassandraTable[(Key, (Int, Int, String))]("read_test", "wide_rows")
+      .select("key", "group", "value")
+      .spanByKey
+      .collect()
+      .toMap
+
+    results should have size 2
+    results.map(_._1) should contain (Key(10))
```
helena: results.keys

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/SpannedByKeyRDD.scala
```diff
@@ -0,0 +1,24 @@
+package com.datastax.spark.connector.rdd
+
+import com.datastax.spark.connector.util.SpanningIterator
```
helena: i'd just move this below the 3rd party imports as well

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/SpannedRDD.scala
```diff
@@ -0,0 +1,34 @@
+package com.datastax.spark.connector.rdd
+
+import com.datastax.spark.connector.util.{SpanningIterator, BufferedIterator2}
```
helena: I'd move this below language and third party imports

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/SpannedByKeyRDD.scala
```diff
@@ -0,0 +1,24 @@
...
+private[connector] class SpannedByKeyRDD[K, V](parent: RDD[(K, V)]) extends RDD[(K, Seq[V])](parent) {
+
+  override protected def getPartitions = parent.partitions
+
+  @DeveloperApi
+  override def compute(split: Partition, context: TaskContext) = {
+    val parentIterator = parent.iterator(split, context)
+    def keyFunction(item: (K, V)) = item._1
+    def extractValues(group: (K, Seq[(K, V)])) = (group._1, group._2.map(_._2))
+    new SpanningIterator(parentIterator, keyFunction).map(extractValues)
```
helena: clean!

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/BufferedIterator2.scala
```diff
@@ -0,0 +1,56 @@
...
+  private[this] var headDefined: Boolean = false
+  private[this] var headElement: T = advance()
+
+  def head =
+    if (headDefined) headElement
+    else throw new NoSuchElementException("Head of empty iterator")
+
+  def headOption = headElement
+
+  private def advance(): T = {
```
helena: hrm, why the parents if it is not a Unit ?
should it be 

    private def advance: T

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/BufferedIterator2.scala
```diff
@@ -0,0 +1,56 @@
...
+import scala.collection.mutable.ArrayBuffer
+
+/** Serves the same purpose as `BufferedIterator` in Scala, but its `takeWhile` method 
+  * properly doesn't consume the next element. */
+class BufferedIterator2[T](iterator: Iterator[T]) extends Iterator[T] {
+
+  // Instead of a pair of T and Boolean we could use Option, but
+  // this would allocate a new Option object per each item, which might be
+  // too big overhead in tight loops using this iterator.
+  private[this] var headDefined: Boolean = false
```
helena: Personally I would not use this and would prefer to see

    if (iterator.hasNext) advance

It is clearer and removes the need for a var

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/BufferedIterator2.scala
pkolaczk: It has a side-effect. The style rules are that side-effecting methods should have parens, even if not returning Unit.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/BufferedIterator2.scala
pkolaczk: It would be clearer, but it wouldn't work.

helena: +1

helena:
@pkolaczk other than the minutia I commented on, +1

PR #521, opened 2015-01-22
Added support for SSL connections

rstrickland:


pkolaczk:
@jacek-lewandowski  can you take a look at this?

jacek-lewandowski:
please rebase it

rstrickland:
All the relevant properties are exposed by default: http://docs.oracle.com/cd/B19306_01/java.102/b14355/sslthin.htm#BABJIHFJ.  The driver uses the default options when you call the default SSLOptions constructor, which results in the JVM respecting these system JSSE properties.

rstrickland:
Also, do you mean for me to rebase against master?

jacek-lewandowski:
@rstrickland yes - rebase against the master

For the SSL settings - so if you set them on the client host you will have to pass them also as `spark.executor.extraJavaOptions`. Don't you think that adding them to Spark configuration and distributing them this way is better?

Also - what about Thrift based connections?

rstrickland:
I went down this road initially, but then it seemed like wheel reinvention.  But I can see it both ways.  I'm happy to add spark options if you'd like to go that route.

jacek-lewandowski:
@rstrickland IMHO that would be better. Especially because we've recently moved all the connection parameters into Spark configuration. This has also one more benefit - it allows us to introduce namespace in Spark configuration to setup connections to different clusters.

jacek-lewandowski:
@rstrickland are you going to work on this?

rstrickland:
@jacek-lewandowski it's on my radar, but I've been distracted by other stuff lately.

jacek-lewandowski:
@rstrickland could you create a ticket in Jira for this ticket and add a link to this branch?

RussellSpitzer:
I'm pretty sure this is done (since we have a lot of ssl options :P) but just incase i'm missing something to we still need this?

jacek-lewandowski:
Closing this - we already have SSL support and client auth is covered by SPARKC-359

PR #527, opened 2015-01-28
SPARKC-25: Initial work to allow RDD's to be used as Keys for C*

jacek-lewandowski:
Added function keyByReplica
    For keying RDD's based on their cassandra replica placement
Added function partitionByReplica
    Which keys an RDD based on replica placement then creates
    new partitions based on their replicas

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
@@ -1,13 +1,20 @@
 package com.datastax.spark.connector
 
-import com.datastax.spark.connector.cql.{TableDef, CassandraConnector}
+
+import java.net.InetAddress
+
+import com.datastax.spark.connector.cql._
 import com.datastax.spark.connector.mapper.ColumnMapper
-import com.datastax.spark.connector.rdd.SpannedRDD
-import com.datastax.spark.connector.writer._
+import com.datastax.spark.connector.rdd.partitioner.{CassandraPartitionedRDD, ReplicaPartitioner}
+import com.datastax.spark.connector.rdd.reader._
+import com.datastax.spark.connector.rdd.{CassandraJoinRDD, SpannedRDD, ValidRDDType}
+import com.datastax.spark.connector.writer.{ReplicaMapper, _}
 import org.apache.spark.SparkContext
+import org.apache.spark.SparkContext._
 import org.apache.spark.rdd.RDD
 
 import scala.reflect.ClassTag
+
 
 /** Provides Cassandra-specific methods on `RDD` */
 class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializable {
@@ -87,8 +94,71 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
   /** Applies a function to each item, and groups consecutive items having the same value together.
     * Contrary to `groupBy`, items from the same group must be already next to each other in the
     * original collection. Works locally on each partition, so items from different
-    * partitions will never be placed in the same group.*/
+    * partitions will never be placed in the same group. */
   def spanBy[U](f: (T) => U): RDD[(U, Iterable[T])] =
     new SpannedRDD[U, T](rdd, f)
+
+  /**
+   * Uses the data from `RDD` to join with a Cassandra table without retrieving the entire table.
+   * Any RDD which can be used to saveToCassandra can be used to joinWithCassandra as well as any
+   * RDD which only specifies the partition Key of a Cassandra Table. This method executes single
+   * partition requests against the Cassandra Table and accepts the functional modifiers that a
+   * normal [[com.datastax.spark.connector.rdd.CassandraTableScanRDD]] takes.
+   *
+   * By default this method only uses the Partition Key for joining but any combination of columns
+   * which are acceptable to C* can be used in the join. Specify columns using joinColumns as a parameter
+   * or the on() method.
+   *
+   * Example With Prior Repartitioning: {{{
+   * val source = sc.parallelize(keys).map(x => new KVRow(x))
+   * val repart = source.repartitionByCassandraReplica(keyspace, tableName, 10)
+   * val someCass = repart.joinWithCassandraTable(keyspace, tableName)
+   * }}}
+   *
+   * Example Joining on Clustering Columns: {{{
+   * val source = sc.parallelize(keys).map(x => (x, x * 100))
+   * val someCass = source.joinWithCassandraTable(keyspace, wideTable).on(SomeColumns("key", "group"))
+   * }}}
+   **/
+  def joinWithCassandraTable[R](keyspaceName: String, tableName: String,
+                                selectedColumns: ColumnSelector = AllColumns,
+                                joinColumns: ColumnSelector = PartitionKeyColumns)
+                               (implicit connector: CassandraConnector = CassandraConnector(sparkContext.getConf),
+                                newType: ClassTag[R], rrf: RowReaderFactory[R], ev: ValidRDDType[R],
+                                currentType: ClassTag[T], rwf: RowWriterFactory[T]): CassandraJoinRDD[T, R] = {
+    new CassandraJoinRDD[T, R](rdd, keyspaceName, tableName, connector, columnNames = selectedColumns, joinColumns = joinColumns)
+  }
+
+
+  /**
+   * Repartitions the data (via a shuffle) based upon the replication of the given `keyspaceName` and `tableName`. 
+   * Calling this method before using joinWithCassandraTable will ensure that requests will be coordinator
+   * local. `partitionsPerHost` Controls the number of Spark Partitions that will be created in this repartitioning
+   * event.
+   * The calling RDD must have rows that can be converted into the partition key of the given Cassandra Table.
+   **/
+  def repartitionByCassandraReplica(keyspaceName: String, tableName: String, partitionsPerHost: Int = 10)
+                                   (implicit connector: CassandraConnector = CassandraConnector(sparkContext.getConf),
+                                    currentType: ClassTag[T], rwf: RowWriterFactory[T]) = {
+    val part = new ReplicaPartitioner(partitionsPerHost, connector)
+    val repart = rdd.keyByCassandraReplica(keyspaceName, tableName).partitionBy(part)
+    val output = repart.mapPartitions(_.map(_._2), preservesPartitioning = true)
+    new CassandraPartitionedRDD[T](output)
+  }
+
+  /**
+   * Key every row in the RDD by with the IP Adresses of all of the Cassandra nodes which a contain a replica
+   * of the data specified by that row.
+   * The calling RDD must have rows that can be converted into the partition key of the given Cassandra Table.
+   */
+  def keyByCassandraReplica(keyspaceName: String, tableName: String)
+                           (implicit connector: CassandraConnector = CassandraConnector(sparkContext.getConf),
+                            currentType: ClassTag[T], rwf: RowWriterFactory[T]): RDD[(Set[InetAddress], T)] = {
+    val converter = ReplicaMapper[T](connector, keyspaceName, tableName)
+    rdd.mapPartitions(primaryKey =>
+      converter.keyByReplicas(primaryKey)
+    )
+  }
+
 
 }
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
@@ -1,0 +1,180 @@
+package com.datastax.spark.connector.rdd
+
+import com.datastax.driver.core.Session
+import com.datastax.spark.connector._
+import com.datastax.spark.connector.cql._
+import com.datastax.spark.connector.metrics.InputMetricsUpdater
+import com.datastax.spark.connector.rdd.reader._
+import com.datastax.spark.connector.util.CqlWhereParser.{EqPredicate, InListPredicate, InPredicate, RangePredicate}
+import com.datastax.spark.connector.util.{CountingIterator, CqlWhereParser}
+import com.datastax.spark.connector.writer._
+import org.apache.spark.rdd.RDD
+import org.apache.spark.{Partition, TaskContext}
+
+import scala.reflect.ClassTag
+
+/**
+ * An RDD that will do a selecting join between `prev:RDD` and the specified Cassandra Table
+ * This will perform individual selects to retrieve the rows from Cassandra and will take
+ * advantage of RDD's that have been partitioned with the [[com.datastax.spark.connector.rdd.partitioner.ReplicaPartitioner]]
+ */
+class CassandraJoinRDD[Left, Right] private[connector](prev: RDD[Left],
+                                                val keyspaceName: String,
+                                                val tableName: String,
+                                                val connector: CassandraConnector,
+                                                val columnNames: ColumnSelector = AllColumns,
+                                                val joinColumns: ColumnSelector = PartitionKeyColumns,
+                                                val where: CqlWhereClause = CqlWhereClause.empty,
+                                                val limit: Option[Long] = None,
+                                                val clusteringOrder: Option[ClusteringOrder] = None,
+                                                val readConf: ReadConf = ReadConf())
+                                                      (implicit oldTag: ClassTag[Left], val rct: ClassTag[Right],
+                                                       @transient val rwf: RowWriterFactory[Left], @transient val rtf: RowReaderFactory[Right])
+  extends CassandraRDD[(Left, Right)](prev.sparkContext, prev.dependencies) with CassandraTableRowReaderProvider[Right] {
+
+  //Make sure copy operations make new CJRDDs and not CRDDs
+  override protected def copy(columnNames: ColumnSelector = columnNames,
+                              where: CqlWhereClause = where,
+                              limit: Option[Long] = limit,
+                              clusteringOrder: Option[ClusteringOrder] = None,
+                              readConf: ReadConf = readConf,
+                              connector: CassandraConnector = connector) =
+    new CassandraJoinRDD[Left, Right](prev, keyspaceName, tableName, connector, columnNames, joinColumns, where, limit, clusteringOrder, readConf).asInstanceOf[this.type]
+
+  lazy val joinColumnNames: Seq[NamedColumnRef] = joinColumns match {
+    case AllColumns => throw new IllegalArgumentException("Unable to join against all columns in a Cassandra Table. Only primary key columns allowed.")
+    case PartitionKeyColumns => tableDef.partitionKey.map(col => col.columnName: NamedColumnRef).toSeq
+    case SomeColumns(cs@_*) => {
+      checkColumnsExistence(cs)
+      cs.map {
+        case c: NamedColumnRef => c
+        case _ => throw new IllegalArgumentException("Unable to join against unnamed columns. No CQL Functions allowed.")
+      }
+    }
+  }
+
+  override def count(): Long = {
+    columnNames match {
+      case SomeColumns(_) => logWarning("You are about to count rows but an explicit projection has been specified.")
+      case _ =>
+    }
+    new CassandraJoinRDD[Left, Long](prev, keyspaceName, tableName, connector, SomeColumns(RowCountRef), joinColumns, where, limit, clusteringOrder, readConf).map(_._2).reduce(_ + _)
+  }
+
+  /**
+   * This method will create the RowWriter required before the RDD is serialized. This is called during getPartitions
+   */
+  protected def checkValidJoin(): Seq[NamedColumnRef] = {
+    val regularColumnNames = tableDef.regularColumns.map(_.columnName).toSet
+    val partitionKeyColumnNames = tableDef.partitionKey.map(_.columnName).toSet
+    val colNames = joinColumnNames.map(_.columnName).toSet
+
+    //Initialize RowWriter and Query to be used for accessing Cassandra
+    rowWriter.columnNames
+    singleKeyCqlQuery.size
+
+    def checkSingleColumn(column: NamedColumnRef): Unit = {
+      if (regularColumnNames.contains(column.columnName))
+        throw new IllegalArgumentException(s"Can't pushdown join on column $column because it is not part of the PRIMARY KEY")
+    }
+
+    //Make sure we have all of the clustering indexes between the 0th position and the max requested in the join
+    val chosenClusteringColumns = tableDef.clusteringColumns
+      .filter(cc => colNames.contains(cc.columnName))
+    if (!tableDef.clusteringColumns.startsWith(chosenClusteringColumns)) {
+      val maxCol = chosenClusteringColumns.last
+      val maxIndex = maxCol.componentIndex.get
+      val requiredColumns = tableDef.clusteringColumns.takeWhile(_.componentIndex.get <= maxIndex)
+      val missingColumns = requiredColumns.toSet -- chosenClusteringColumns.toSet
+      if (!missingColumns.isEmpty)
+        throw new IllegalArgumentException(s"Can't pushdown join on column $maxCol without also specifying [ $missingColumns ]")
+      }
+    val missingPartitionKeys = partitionKeyColumnNames -- colNames
+    if (missingPartitionKeys.size != 0) {
+      throw new IllegalArgumentException(s"Can't join without the full partition key. Missing: [ ${missingPartitionKeys} ]")
+    }
+    joinColumnNames.foreach { case c: NamedColumnRef => checkSingleColumn(c)}
+    joinColumnNames
+  }
+
+  lazy val rowWriter = implicitly[RowWriterFactory[Left]].rowWriter(
+    tableDef,
+    joinColumnNames.map {
+      _.columnName
+    },
+    joinColumns.aliases
+  )
+
+  def on(joinColumns: ColumnSelector): CassandraJoinRDD[Left, Right] = {
+    new CassandraJoinRDD[Left, Right](prev, keyspaceName, tableName, connector, columnNames, joinColumns, where, limit, clusteringOrder, readConf).asInstanceOf[this.type]
+  }
+
+  //We need to make sure we get selectedColumnRefs before serialization so that our RowReader is
+  //built
+  lazy val singleKeyCqlQuery: (String) = {
+    val whereClauses = where.predicates.flatMap(CqlWhereParser.parse)
+    val partitionKeys = tableDef.partitionKey.map(_.columnName)
+    val partitionKeyPredicates = whereClauses.collect {
+      case EqPredicate(c, _) if partitionKeys.contains(c) => c
+      case InPredicate(c) if partitionKeys.contains(c) => c
+      case InListPredicate(c, _) if partitionKeys.contains(c) => c
+      case RangePredicate(c, _, _) if partitionKeys.contains(c) => c
+    }.toSet
+    require(partitionKeyPredicates.isEmpty, s"No partition keys allowed in where on joins with Cassandra. Found : $partitionKeyPredicates")
+
+    logDebug("Generating Single Key Query Prepared Statement String")
+    logDebug(s"SelectedColumns : $selectedColumnRefs -- JoinColumnNames : $joinColumnNames")
+    val columns = selectedColumnRefs.map(_.cql).mkString(", ")
+    val joinWhere = joinColumnNames.map(_.columnName).map(name => s"${quote(name)} = :$name")
+    val limitClause = limit.map(limit => s"LIMIT $limit").getOrElse("")
+    val orderBy = clusteringOrder.map(_.toCql(tableDef)).getOrElse("")
+    val filter = (where.predicates ++ joinWhere).mkString(" AND ")
+    val quotedKeyspaceName = quote(keyspaceName)
+    val quotedTableName = quote(tableName)
+    val query = s"SELECT $columns FROM $quotedKeyspaceName.$quotedTableName WHERE $filter $limitClause $orderBy"
+    logDebug(s"Query : $query")
+    query
+  }
+
+  /**
+   * When computing a CassandraPartitionKeyRDD the data is selected via single CQL statements
+   * from the specified C* Keyspace and Table. This will be preformed on whatever data is
+   * avaliable in the previous RDD in the chain.
+   */
+  override def compute(split: Partition, context: TaskContext): Iterator[(Left, Right)] = {
+    val session = connector.openSession()
+    implicit val pv = protocolVersion(session)
+    val stmt = session.prepare(singleKeyCqlQuery).setConsistencyLevel(consistencyLevel)
+    val bsb = new BoundStatementBuilder[Left](rowWriter, stmt, pv)
+    val metricsUpdater = InputMetricsUpdater(context, 20)
+    val rowIterator = fetchIterator(session, bsb, prev.iterator(split, context))
+    val countingIterator = new CountingIterator(rowIterator, limit)
+    context.addTaskCompletionListener { (context) =>
+      val duration = metricsUpdater.finish() / 1000000000d
+      logDebug(f"Fetched ${countingIterator.count} rows from $keyspaceName.$tableName for partition ${split.index} in $duration%.3f s.")
+      session.close()
+    }
+    countingIterator
+  }
+
+  private def fetchIterator(session: Session, bsb: BoundStatementBuilder[Left], lastIt: Iterator[Left]): Iterator[(Left, Right)] = {
+    val columnNamesArray = selectedColumnRefs.map(_.selectedFromCassandraAs).toArray
+    implicit val pv = protocolVersion(session)
+    for (leftSide <- lastIt;
+         rightSide <- {
+           val rs = session.execute(bsb.bind(leftSide))
+           val iterator = new PrefetchingResultSetIterator(rs, fetchSize)
+           iterator.map(rowReader.read(_, columnNamesArray))
+         }) yield (leftSide, rightSide)
+  }
+
+  override protected def getPartitions: Array[Partition] = {
+    verify()
+    checkValidJoin()
+    prev.partitions
+  }
+
+  override def getPreferredLocations(split: Partition): Seq[String] = prev.preferredLocations(split)
+
+  override def toEmptyCassandraRDD(): EmptyCassandraRDD[(Left, Right)] = new EmptyCassandraRDD[(Left, Right)](prev.sparkContext, keyspaceName, tableName, columnNames, where, limit, clusteringOrder, readConf)
+}
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraRDD.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraRDD.scala
@@ -1,148 +1,61 @@
 package com.datastax.spark.connector.rdd
 
-import java.io.IOException
-
-import com.datastax.spark.connector.rdd.ClusteringOrder.{Descending, Ascending}
-
-import scala.reflect.ClassTag
-import scala.collection.JavaConversions._
-import scala.language.existentials
-
-import com.datastax.spark.connector.metrics.InputMetricsUpdater
-import org.apache.spark.rdd.RDD
-import org.apache.spark.{Logging, Partition, SparkContext, TaskContext}
-
-import com.datastax.driver.core._
-import com.datastax.spark.connector.{SomeColumns, AllColumns, ColumnSelector}
 import com.datastax.spark.connector.cql._
-import com.datastax.spark.connector.rdd.partitioner.{CassandraRDDPartitioner, CassandraPartition, CqlTokenRange}
-import com.datastax.spark.connector.rdd.partitioner.dht.TokenFactory
+import com.datastax.spark.connector.rdd.ClusteringOrder.{Ascending, Descending}
 import com.datastax.spark.connector.rdd.reader._
-import com.datastax.spark.connector.types.{ColumnType, TypeConverter}
-import com.datastax.spark.connector.util.CountingIterator
-import com.datastax.spark.connector._
+import com.datastax.spark.connector.types.TypeConverter
+import com.datastax.spark.connector.{ColumnSelector, NamedColumnRef, SomeColumns, _}
+import org.apache.spark.rdd.RDD
+import org.apache.spark.{Dependency, SparkContext}
 
+import scala.language.existentials
+import scala.reflect.ClassTag
 
-/** RDD representing a Cassandra table.
-  * This class is the main entry point for analyzing data in Cassandra database with Spark.
-  * Obtain objects of this class by calling [[com.datastax.spark.connector.SparkContextFunctions#cassandraTable cassandraTable]].
-  *
-  * Configuration properties should be passed in the `SparkConf` configuration of `SparkContext`.
-  * `CassandraRDD` needs to open connection to Cassandra, therefore it requires appropriate connection property values
-  * to be present in `SparkConf`. For the list of required and available properties, see
-  * [[com.datastax.spark.connector.cql.CassandraConnector CassandraConnector]].
-  *
-  * `CassandraRDD` divides the dataset into smaller partitions, processed locally on every cluster node.
-  * A data partition consists of one or more contiguous token ranges.
-  * To reduce the number of roundtrips to Cassandra, every partition is fetched in batches. The following
-  * properties control the number of partitions and the fetch size:
-  *
-  *   - spark.cassandra.input.split.size:        approx number of Cassandra partitions in a Spark partition, default 100000
-  *   - spark.cassandra.input.page.row.size:     number of CQL rows fetched per roundtrip, default 1000
-  *
-  * A `CassandraRDD` object gets serialized and sent to every Spark executor.
-  *
-  * By default, reads are performed at ConsistencyLevel.LOCAL_ONE in order to leverage data-locality and minimize network traffic.
-  * This read consistency level is controlled by the following property:
-  *
-  *   - spark.cassandra.input.consistency.level: consistency level for RDD reads, string matching the ConsistencyLevel enum name.
-  *
-  * If a Cassandra node fails or gets overloaded during read, queries are retried to a different node.
-  */
-class CassandraRDD[R] private[connector] (
-    @transient val sc: SparkContext,
-    val connector: CassandraConnector,
-    val keyspaceName: String,
-    val tableName: String,
-    val columnNames: ColumnSelector = AllColumns,
-    val where: CqlWhereClause = CqlWhereClause.empty,
-    val empty: Boolean = false,
-    val limit: Option[Long] = None,
-    val clusteringOrder: Option[ClusteringOrder] = None,
-    val readConf: ReadConf = ReadConf())(
-  implicit
-    ct : ClassTag[R], @transient val rtf: RowReaderFactory[R])
-  extends RDD[R](sc, Seq.empty) with Logging {
+abstract class CassandraRDD[R](_sc: SparkContext, dep: Seq[Dependency[_]])(implicit ct: ClassTag[R]) extends RDD[R](_sc, dep) {
+
+  protected def keyspaceName: String
+
+  protected def tableName: String
+
+  protected def columnNames: ColumnSelector
+
+  protected def where: CqlWhereClause
+
+  protected def readConf: ReadConf
+
+  protected def limit: Option[Long]
 
   require(limit.isEmpty || limit.get > 0, "Limit must be greater than 0")
 
-  private def fetchSize = readConf.fetchSize
-  private def splitSize = readConf.splitSize
-  private def consistencyLevel = readConf.consistencyLevel
+  protected def clusteringOrder: Option[ClusteringOrder]
 
-  private def copy(columnNames: ColumnSelector = columnNames,
-                   clusteringOrder: Option[ClusteringOrder] = clusteringOrder,
-                   where: CqlWhereClause = where,
-                   empty: Boolean = empty,
-                   limit: Option[Long] = limit,
-                   readConf: ReadConf = readConf,
-                   connector: CassandraConnector = connector): CassandraRDD[R] = {
-    require(sc != null,
-      "RDD transformation requires a non-null SparkContext. Unfortunately SparkContext in this CassandraRDD is null. " +
-      "This can happen after CassandraRDD has been deserialized. SparkContext is not Serializable, therefore it deserializes to null." +
-      "RDD transformations are not allowed inside lambdas used in other RDD transformations.")
-    new CassandraRDD(sc, connector, keyspaceName, tableName, columnNames, where, empty, limit, clusteringOrder, readConf)
-  }
+  protected def connector: CassandraConnector
 
-  /** Returns a copy of this Cassandra RDD with specified connector */
-  def withConnector(connector: CassandraConnector): CassandraRDD[R] =
-    copy(connector = connector)
-
-  /** Adds a CQL `WHERE` predicate(s) to the query.
-    * Useful for leveraging secondary indexes in Cassandra.
-    * Implicitly adds an `ALLOW FILTERING` clause to the WHERE clause, however beware that some predicates
-    * might be rejected by Cassandra, particularly in cases when they filter on an unindexed, non-clustering column.*/
-  def where(cql: String, values: Any*): CassandraRDD[R] = {
-    copy(where = where and CqlWhereClause(Seq(cql), values))
-  }
-
-  /** Adds a CQL `ORDER BY` clause to the query.
-    * It can be applied only in case there are clustering columns and primary key predicate is
-    * pushed down in `where`.
-    * It is useful when the default direction of ordering rows within a single Cassandra partition
-    * needs to be changed. */
-  def clusteringOrder(order: ClusteringOrder): CassandraRDD[R] = {
-    copy(clusteringOrder = Some(order))
-  }
-
-  def withAscOrder: CassandraRDD[R] = clusteringOrder(Ascending)
-  def withDescOrder: CassandraRDD[R] = clusteringOrder(Descending)
+  def toEmptyCassandraRDD(): EmptyCassandraRDD[R]
 
   /** Allows to set custom read configuration, e.g. consistency level or fetch size. */
   def withReadConf(readConf: ReadConf) = {
     copy(readConf = readConf)
   }
 
-  /** Produces the empty CassandraRDD which has the same signature and properties, but it does not
-    * perform any validation and it does not even try to return any rows. */
-  def toEmptyCassandraRDD = {
-    copy(empty = true)
+  protected def copy(columnNames: ColumnSelector = columnNames,
+                     where: CqlWhereClause = where,
+                     limit: Option[Long] = limit,
+                     clusteringOrder: Option[ClusteringOrder] = None,
+                     readConf: ReadConf = readConf,
+                     connector: CassandraConnector = connector): CassandraRDD[R]
+
+  /** Returns a copy of this Cassandra RDD with specified connector */
+  def withConnector(connector: CassandraConnector): CassandraRDD[R] = {
+    copy(connector = connector)
   }
 
-  /** Throws IllegalArgumentException if columns sequence contains unavailable columns */
-  private def checkColumnsAvailable(columns: Seq[SelectableColumnRef], availableColumns: Seq[SelectableColumnRef]) {
-    val availableColumnsSet = availableColumns.toSet
-
-    val notFound = columns.find {
-      case column => !availableColumnsSet.contains(column)
-    }
-
-    if (notFound.isDefined)
-      throw new IllegalArgumentException(
-        s"Column not found in selection: ${notFound.get}. " +
-          s"Available columns: [${availableColumns.mkString(",")}].")
-  }
-
-  /** Filters currently selected set of columns with a new set of columns */
-  private def narrowColumnSelection(columns: Seq[SelectableColumnRef]): Seq[SelectableColumnRef] = {
-    columnNames match {
-      case SomeColumns(cs @ _*) =>
-        checkColumnsAvailable(columns, cs)
-      case AllColumns =>
-        // we do not check for column existence yet as it would require fetching schema and a call to C*
-        // columns existence will be checked by C* once the RDD gets computed.
-    }
-    columns
+  /** Adds a CQL `WHERE` predicate(s) to the query.
+    * Useful for leveraging secondary indexes in Cassandra.
+    * Implicitly adds an `ALLOW FILTERING` clause to the WHERE clause, however beware that some predicates
+    * might be rejected by Cassandra, particularly in cases when they filter on an unindexed, non-clustering column. */
+  def where(cql: String, values: Any*): CassandraRDD[R] = {
+    copy(where = where and CqlWhereClause(Seq(cql), values))
   }
 
   /** Narrows down the selected set of columns.
@@ -161,14 +74,6 @@ class CassandraRDD[R] private[connector] (
     copy(columnNames = SomeColumns(narrowColumnSelection(columns): _*))
   }
 
-  override def count(): Long = {
-    columnNames match {
-      case SomeColumns(_) => logWarning("You are about to count rows but an explicit projection has been specified.")
-      case _ =>
-    }
-    new CassandraRDD[Long](sc, connector, keyspaceName, tableName, SomeColumns(RowCountRef), where, empty, limit, clusteringOrder, readConf).reduce(_ + _)
-  }
-
   /** Adds the limit clause to CQL select statement. The limit will be applied for each created
     * Spark partition. In other words, unless the data are fetched from a single Cassandra partition
     * the number of results is unpredictable.
@@ -180,6 +85,35 @@ class CassandraRDD[R] private[connector] (
     copy(limit = Some(rowLimit))
   }
 
+  /** Adds a CQL `ORDER BY` clause to the query.
+    * It can be applied only in case there are clustering columns and primary key predicate is
+    * pushed down in `where`.
+    * It is useful when the default direction of ordering rows within a single Cassandra partition
+    * needs to be changed. */
+  def clusteringOrder(order: ClusteringOrder): CassandraRDD[R] = {
+    copy(clusteringOrder = Some(order))
+  }
+
+  def withAscOrder: CassandraRDD[R] = clusteringOrder(Ascending)
+
+  def withDescOrder: CassandraRDD[R] = clusteringOrder(Descending)
+
+  override def take(num: Int): Array[R] = {
+    limit match {
+      case Some(_) => super.take(num)
+      case None => limit(num).take(num)
+    }
+  }
+
+  protected def narrowColumnSelection(columns: Seq[SelectableColumnRef]): Seq[SelectableColumnRef]
+
+  // Needed to be public for JavaAPI
+  val selectedColumnRefs: Seq[SelectableColumnRef]
+
+  //convertTo must be implmented for classes which wish to support `.as`
+  protected def convertTo[B](implicit ct: ClassTag[B], rrf: RowReaderFactory[B]): CassandraRDD[B] =
+    throw new NotImplementedError(s"convertTo not implemented for this class")
+
   /** Maps each row into object of a different type using provided function taking column value(s) as argument(s).
     * Can be used to convert each row to a tuple or a case class object:
     * {{{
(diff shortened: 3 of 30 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/DefaultRowWriter.scala
```diff
@@ -9,9 +9,15 @@ import com.datastax.spark.connector.types.TypeConverter
 import scala.collection.{Map, Seq}
 import scala.collection.JavaConversions._
 
+trait CheckSetting
+object CheckLevel{
```
jacek-lewandowski: Why do you need this?

RussellSpitzer: So i can use the RowWriter factory without it complaining that i'm missing keys to get values to put in the driver

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraPartitionKeyRDD.scala
```diff
@@ -0,0 +1,125 @@
...
+  //We need to make sure we get selectedColumnNames before serialization so that our RowReader is
+  //built
+  private val singleKeyCqlQuery: (String) = {
+    logDebug("Generating Single Key Query Prepared Statement String")
+    val columns = selectedColumnNames.map(_.cql).mkString(", ")
+    val partitionWhere = tableDef.partitionKey.map(_.columnName).map(name => s"$name = :$name")
+    val filter = (where.predicates ++ partitionWhere).mkString(" AND ")
+    val quotedKeyspaceName = quote(keyspaceName)
+    val quotedTableName = quote(tableName)
+    logDebug(s"SELECT $columns FROM $quotedKeyspaceName.$quotedTableName WHERE $filter")
```
jacek-lewandowski: maybe you can assign the query to some val instead of repeating it?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraPartitionKeyRDD.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  private val converter = ReplicaMapper[O](connector, keyspaceName, tableName)
+
+  //We need to make sure we get selectedColumnNames before serialization so that our RowReader is
+  //built
+  private val singleKeyCqlQuery: (String) = {
+    logDebug("Generating Single Key Query Prepared Statement String")
+    val columns = selectedColumnNames.map(_.cql).mkString(", ")
+    val partitionWhere = tableDef.partitionKey.map(_.columnName).map(name => s"$name = :$name")
+    val filter = (where.predicates ++ partitionWhere).mkString(" AND ")
```
jacek-lewandowski: Here is a bug. Where clause can already include a partition key predicate. In this case that predicate is duplicated and Cassandra complains about it.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraPartitionKeyRDD.scala
```diff
@@ -0,0 +1,125 @@
...
+    new CassandraPartitionKeyRDD[O,N](prev, keyspaceName, tableName, connector, columnNames, where, readConf)
+
+  private val converter = ReplicaMapper[O](connector, keyspaceName, tableName)
+
+  //We need to make sure we get selectedColumnNames before serialization so that our RowReader is
+  //built
+  private val singleKeyCqlQuery: (String) = {
+    logDebug("Generating Single Key Query Prepared Statement String")
+    val columns = selectedColumnNames.map(_.cql).mkString(", ")
+    val partitionWhere = tableDef.partitionKey.map(_.columnName).map(name => s"$name = :$name")
```
jacek-lewandowski: i wonder whether you should quote the column names here

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraPartitionKeyRDD.scala
```diff
@@ -0,0 +1,125 @@
...
+   * avaliable in the previous RDD in the chain.
+   * @param split
+   * @param context
+   * @return
+   */
+  override def compute(split: Partition, context: TaskContext): Iterator[N] = {
+    connector.withSessionDo { session =>
+      logDebug(s"Query::: $singleKeyCqlQuery")
+      val stmt = session.prepare(singleKeyCqlQuery).setConsistencyLevel(consistencyLevel)
+      fetchIterator(session, stmt, prev.iterator(split, context))
```
jacek-lewandowski: i think that the session can be closed before the iterator is exhausted - take a look how it is done in CassandraRDD

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraPartitionKeyRDD.scala
```diff
@@ -0,0 +1,125 @@
...
+      }
+    }
+
+  def fetchIterator(session:Session, stmt: PreparedStatement, lastIt:Iterator[O]): Iterator[N] = {
+    converter.bindStatements(lastIt, stmt).flatMap { request => //flatMap Because we may get multiple results for a single query
+      implicit val pv = protocolVersion(session)
+      val columnNamesArray = selectedColumnNames.map(_.selectedAs).toArray
```
jacek-lewandowski: nit: columnNamesArray can be defined outside the loop

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
```diff
@@ -34,4 +38,28 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
...
+   * an order valid for Cassandra.
+   *
+   * The flag repartition (default: true) tells Spark whether or not it should shuffle and repartition the data
+   * so that the queries are preformed on a coordinator for the data they request. When preforming inner-joins between
+   * Cassandra tables this is not neccessary so set this to false so no data shuffling occurs.
+   */
+  def fetchFromCassandra[R](keyspaceName: String, tableName: String, repartition: Boolean = true, partitionsPerReplicaSet:Int = 10)
+                           (implicit connector: CassandraConnector = CassandraConnector(sparkContext.getConf),
+                            newType: ClassTag[R], rrf: RowReaderFactory[R], ev: ValidRDDType[R],
+                            currentType: ClassTag[T], rwf: RowWriterFactory[T]): CassandraRDD[R] = {
```
jacek-lewandowski: we could avoid the requirement for `currentType: ClassTag[T]` because this class tag is already in rdd wrapped by this `RDDFunctions`. The way to obtain it is `rdd.toJavaRDD().classTag`.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
```diff
@@ -34,4 +38,28 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
...
 
+  /**
+   * Uses the data from `RDD` to get data from a Cassandra table. This function is slightly more liberal than
+   * saveToCassandra as you are not required to specify the entire Primary Key. You must specify
+   * the entire partition key but clustering columns are optional. If clustering columns are included they must be in
+   * an order valid for Cassandra.
+   *
+   * The flag repartition (default: true) tells Spark whether or not it should shuffle and repartition the data
+   * so that the queries are preformed on a coordinator for the data they request. When preforming inner-joins between
+   * Cassandra tables this is not neccessary so set this to false so no data shuffling occurs.
```
jacek-lewandowski: IntelliJ reports typo: neccessary -> necessary ;-)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
jacek-lewandowski: it would be useful to add here `ReadConf` parameter

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
```diff
@@ -0,0 +1,34 @@
...
+
+
+case class EndpointPartition(index: Int, endpoint: Set[InetAddress]) extends Partition
+
+class ReplicaPartitioner(partitionsPerReplicaSet: Int, connector:CassandraConnector) extends Partitioner {
+    val hosts = connector.hosts
+    val hostMap = hosts.zipWithIndex.toMap
+    /*TODO We Need JAVA-312 to get sets of replicas instead of single endpoints. Once we have that we'll be able to
+    build a map of Set[ip,ip,...] => Index before looking at our data. */
+    val indexMap = (0 to partitionsPerReplicaSet).flatMap(offset => hostMap.map { case (hosts, index) => (index + offset, Set(hosts))}).toMap
```
jacek-lewandowski: if you expect it to iterate over values between 0 and partitionsPerReplicaSet-1, you should use `until` instead of `to`

jacek-lewandowski: in `case (hosts, index)` you really should use different name for `hosts` because (1) it masks `hosts` value and (2) it is really a single host.

jacek-lewandowski: other than that - i cannot understand what it really does - i guess it produces a map which has all values from range (0, partitionsPerReplicaSet.size + hosts.size) as keys mapped to sets of a single, random host

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
```diff
@@ -0,0 +1,34 @@
+package com.datastax.spark.connector.rdd.partitioner
+
+import java.net.InetAddress
+
+import com.datastax.spark.connector.cql.CassandraConnector
+import org.apache.spark.{Partition, Partitioner}
+
+
+case class EndpointPartition(index: Int, endpoint: Set[InetAddress]) extends Partition
```
jacek-lewandowski: Would you make a trait `ICassandraPartition` or something which would be extended by `CassandraPartition` and `EndpointPartition`? they both share `index` and `endpoint` fields

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
```diff
@@ -0,0 +1,34 @@
...
+
+class ReplicaPartitioner(partitionsPerReplicaSet: Int, connector:CassandraConnector) extends Partitioner {
+    val hosts = connector.hosts
```
RussellSpitzer: Discussion with Jacek, This needs to be simplified as the current code is far too complicated for the task at hand

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
RussellSpitzer: Index is actually inherited from Partition so i'll just make a Endpoints trait

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraPartitionKeyRDD.scala
RussellSpitzer: I also realize this doesn't allow you to push in clustering column values, not sure if that's worth adding ... looking into it

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
RussellSpitzer: I'm not sure how to do the rdd.toJavaRDD().classTag code that you are asking about

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraPartitionKeyRDD.scala
RussellSpitzer: It looked like the session isn't closed until the iterator is exhausted in CassandraRDD. Maybe i'm misreading?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraRDD.scala
```diff
@@ -21,64 +20,49 @@ import com.datastax.spark.connector.util.{Logging, CountingIterator}
 import com.datastax.spark.connector._
 
 
+
 /** RDD representing a Cassandra table.
   * This class is the main entry point for analyzing data in Cassandra database with Spark.
-  * Obtain objects of this class by calling [[com.datastax.spark.connector.SparkContextFunctions#cassandraTable cassandraTable]].
+  * Obtain objects of this class by calling [[com.datastax.spark.connector.SparkContextFunctions# c a s s a n d r a T a b l e c a s s a n d r a T a b l e]].
```
RussellSpitzer: Note to myself to cleanup this spacing issue ...

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/BaseCassandraRDD.scala
```diff
@@ -0,0 +1,245 @@
...
+                                                          ct: ClassTag[R], rddt: ClassTag[P],
+                                                          @transient rtf: RowReaderFactory[R])
+  extends RDD[P](sc, dependecyList) with Logging {
+
+  /* Logging classes inheritance conflict fix. */
+  override def log = super[Logging].log
+
+  override def logName = super[Logging].logName
+
+  override def logInfo(msg: => String) = super[Logging].logInfo(msg)
```
jacek-lewandowski: Why all this logging stuff here?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/BaseCassandraRDD.scala
```diff
@@ -0,0 +1,245 @@
...
+
+  override def isTraceEnabled() = super[Logging].isTraceEnabled()
+
+  protected def fetchSize = readConf.fetchSize
+
+  protected def splitSize = readConf.splitSize
+
+  protected def consistencyLevel = readConf.consistencyLevel
+
+  protected def copy(columnNames: ColumnSelector = columnNames,
```
jacek-lewandowski: I suppose that you may want to try using `this.type` as a return type. Take a look [here](http://scalada.blogspot.com/2008/02/thistype-for-chaining-method-calls.html)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/BaseCassandraRDD.scala
jacek-lewandowski: I cleaned it a bit - take a look https://datastax-oss.atlassian.net/browse/SPARKC-54

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraTableScan.scala
```diff
@@ -10,27 +10,25 @@ import org.apache.spark.sql.execution.LeafNode
...
-      rdd = rdd.select(attributes.map(a => relation.columnNameByLowercase(a.name): NamedColumnRef): _*)
+      rdd.select(attributes.map(a => relation.columnNameByLowercase(a.name): NamedColumnRef): _*)
     if (pushdownPred.nonEmpty) {
-      val(cql, values) = whereClause(pushdownPred)
-      rdd = rdd.where(cql, values: _*)
+      val (cql, values) = whereClause(pushdownPred)
+      rdd.where(cql, values: _*)
     }
-
-    rdd
```
jacek-lewandowski: By removing this you made the result of this method unpredictable.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/BaseCassandraRDD.scala
RussellSpitzer: Yeah this is just from the old code, I"ll rebase to your mods if SPARKC-54 gets merged first

spark-cassandra-connector-java/src/main/java/com/datastax/spark/connector/japi/rdd/CassandraJavaPairRDD.java
```diff
@@ -44,7 +45,7 @@ public CassandraJavaPairRDD(CassandraRDD<Tuple2<K, V>> rdd, ClassTag<K> keyClass
     public CassandraJavaPairRDD<K, V> select(String... columnNames) {
         // explicit type argument is intentional and required here
         //noinspection RedundantTypeArguments
-        CassandraRDD<Tuple2<K, V>> newRDD = rdd().select(JavaApiHelper.<NamedColumnRef>toScalaSeq(CassandraJavaUtil.convert(columnNames)));
+        CassandraRDD<Tuple2<K, V>> newRDD = (CassandraRDD<Tuple2<K, V>>) rdd().select(JavaApiHelper.<NamedColumnRef>toScalaSeq(CassandraJavaUtil.convert(columnNames)));
```
pkolaczk: Hmm, this casting is not very elegant. :(

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/BaseCassandraRDD.scala
```diff
@@ -0,0 +1,219 @@
...
+                                                          val connector: CassandraConnector,
+                                                          val keyspaceName: String,
+                                                          val tableName: String,
+                                                          val columnNames: ColumnSelector = AllColumns,
+                                                          val where: CqlWhereClause = CqlWhereClause.empty,
+                                                          val readConf: ReadConf = ReadConf(),
+                                                          val dependecyList: Seq[Dependency[_]] = Seq.empty)(
+                                                          implicit
+                                                          ct: ClassTag[R], rddt: ClassTag[P],
+                                                          @transient rtf: RowReaderFactory[R])
```
pkolaczk: rtf is *not used* in this BaseClass.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/BaseCassandraRDD.scala
```diff
@@ -0,0 +1,219 @@
...
+  * A `CassandraRDD` object gets serialized and sent to every Spark executor.
+  *
+  * By default, reads are performed at ConsistencyLevel.LOCAL_ONE in order to leverage data-locality and minimize network traffic.
+  * This read consistency level is controlled by the following property:
+  *
+  * - spark.cassandra.input.consistency.level: consistency level for RDD reads, string matching the ConsistencyLevel enum name.
+  *
+  * If a Cassandra node fails or gets overloaded during read, queries are retried to a different node.
+  */
+abstract class BaseCassandraRDD[R, P] private[connector](
```
pkolaczk: R type is not really used. Am I missing something?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/BaseCassandraRDD.scala
pkolaczk: There is a this.type now here, but Java side seems to ignore this...

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/DefaultRowWriter.scala
pkolaczk: How is it related to joins?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/ReplicaMapper.scala
```diff
@@ -0,0 +1,95 @@
...
+import java.net.InetAddress
+
+import com.datastax.driver.core._
+import com.datastax.spark.connector.cql._
+import com.datastax.spark.connector.util.Logging
+
+import scala.collection.JavaConversions._
+import scala.collection._
+
+class ReplicaMapper[T] private(
```
pkolaczk: Documentation needed.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RowWriterFactory.scala
```diff
@@ -13,7 +13,7 @@ trait RowWriterFactory[T] {
   /** Creates a new `RowWriter` instance.
     * @param table target table the user wants to write into
     * @param columnNames columns selected by the user; the user might wish to write only a subset of columns */
-  def rowWriter(table: TableDef, columnNames: Seq[String]): RowWriter[T]
+  def rowWriter(table: TableDef, columnNames: Seq[String], checkColumns: CheckSetting): RowWriter[T]
```
pkolaczk: Whenever I need to add something like this to optionally disable  something somewhere, I think whether the thing I want to disable was in the right place at all. Maybe rowWriters should not perform those checks? Maybe we should just move them somewhere else (TableWriter?) where they do not collide with the function you need this for. This might be reminiscent of ancient pre 1.0 times when there was no RowWriters at all and that logic was part of the TableWriter; then RowWrites have been extracted from TableWriter with those checks.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
```diff
@@ -0,0 +1,51 @@
...
+
+case class ReplicaPartition(index: Int, endpoints: Set[InetAddress]) extends EndpointPartition
+
+/**
+ * The replica partitioner will work on an RDD which is keyed on sets of InetAdresses representing Cassandra
+ * Hosts . It will group keys which share a common IP address into partitionsPerReplicaSet Partitions.
+ * @param partitionsPerReplicaSet The number of Spark Partitions to make Per Unique Endpoint
+ * @param connector
+ */
+class ReplicaPartitioner(partitionsPerReplicaSet: Int, connector: CassandraConnector) extends Partitioner {
```
pkolaczk: Honestly I don't get how this is supposed to collocate Spark partitions with Cassandra replica locations. As I understand, this object can split RDD into partitions having different sets of replicas. However, I'm missing the part which tells Spark which nodes those partitions should be really placed on, so they are close to Cassandra. E.g. a similar behavior like we achieve with getPreferredLocations in RDD.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RowWriterFactory.scala
RussellSpitzer: Thats good point, I'll extract this check out so that we can do column checking in the RDD implementation rather than in the rowWriter construction

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
RussellSpitzer: This code doesn't collocate with replicas, it performs a shuffle which groups rows by replica. The CassandraJoinRDD has the preferredLocation method which uses the parent RDD's partitioner to determine where to execute the partition. 

The only way to make this method actually colocate in place would be by defining a custom shuffler I think ... I'll see if I can determine a more elegant method

RussellSpitzer: I think what I can do is create another RDD type which just defines the getPreferredLocation/partition method and have the final action of this code to be transform to that RDD. There will still be an additional move (from after the shuffle) but any RDD which references this one should pick up the correct location...

RussellSpitzer:
Other than my love of white text on black backgrounds this looks good,
+1 Code
+1 Unittests/It Tests

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraReader.scala
```diff
@@ -0,0 +1,166 @@
...
+import com.datastax.spark.connector._
+import com.datastax.spark.connector.cql.{CassandraConnector, Schema}
+import com.datastax.spark.connector.rdd.reader._
+
+import scala.reflect.ClassTag
+
+trait CassandraReader[R] {
+
+
+  protected def connector: CassandraConnector
```
pkolaczk: I guess many of those protected methods here are just declared in order to get access to their implementations in CassandraRDD. Additionally, this trait is used as the only building block for CassandraRDD subclasses, it doesn't make sense to use it outside of CassandraRDD hierarchy. Therefore, you can use this nice Scala feature:
```scala    
   trait CassandraReader[R] {
      this: CassandraRDD[_] => 
```

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraReader.scala
```diff
@@ -0,0 +1,166 @@
...
+import java.io.IOException
+
+import com.datastax.driver.core.{ProtocolVersion, Session}
+import com.datastax.spark.connector._
+import com.datastax.spark.connector.cql.{CassandraConnector, Schema}
+import com.datastax.spark.connector.rdd.reader._
+
+import scala.reflect.ClassTag
+
+trait CassandraReader[R] {
```
pkolaczk: This needs better documentation what it does. Seems like it deals mostly with checking schema / columns, but doesn't do any *reading*. Can you think of a better name?

pkolaczk:
I haven't looked into details yet, but overall design looks much better now :)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/EmptyCassandraRDD.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  @DeveloperApi
+  override def compute(split: Partition, context: TaskContext): Iterator[R] = throw new UnsupportedOperationException("Cannot call compute on an EmptyRDD")
+
+  lazy val selectedColumnNames: Seq[NamedColumnRef] = {
+    val providedColumnNames =
+      columnNames match {
+        case AllColumns => Seq()
+        case PartitionKeyColumns => Seq()
+        case SomeColumns(cs@_*) => (cs)
```
pkolaczk: Unnecessary parantheses

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/EmptyCassandraRDD.scala
```diff
@@ -0,0 +1,41 @@
...
+      }
+    providedColumnNames
+  }
+
+  override protected def connector: CassandraConnector = throw new UnsupportedOperationException("Empty Cassandra RDD's Don't Have Connections to Cassandra")
+
+  override def toEmptyCassandraRDD(): EmptyCassandraRDD[R] = copy()
```
pkolaczk: No need for this method to have parentheses because it is stateless.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraReader.scala
```diff
@@ -0,0 +1,166 @@
...
+  protected def quote(name: String) = "\"" + name + "\""
+
+  def protocolVersion(session: Session): ProtocolVersion = {
+    session.getCluster.getConfiguration.getProtocolOptions.getProtocolVersionEnum
+  }
+
+  /** Checks for existence of keyspace, table, columns and whether the number of selected columns corresponds to
+    * the number of the columns expected by the target type constructor.
+    * If successful, does nothing, otherwise throws appropriate `IOException` or `AssertionError`. */
+  def verify() = {
```
pkolaczk: Oh, actually this is better now as a method. :)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableScanRDD.scala
```diff
@@ -0,0 +1,174 @@
...
+    new CassandraTableScanRDD[R](sc, connector, keyspaceName, tableName, columnNames, where, readConf).asInstanceOf[this.type]
+  }
+
+  override protected def convertTo[B](implicit ct: ClassTag[B], rrf: RowReaderFactory[B]): CassandraRDD[B] = {
+    new CassandraTableScanRDD[B](sc, connector, keyspaceName, tableName, columnNames, where, readConf)
+  }
+
+
+  override def getPartitions: Array[Partition] = {
+    verify // let's fail fast
```
pkolaczk: And now this needs parens. :)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -0,0 +1,152 @@
...
+    case AllColumns => throw new IllegalArgumentException("Unable to join against all columns in a Cassandra Table. Only primary key columns allowed")
+    case PartitionKeyColumns => tableDef.partitionKey.map(col => col.columnName: NamedColumnRef).toSeq
+    case SomeColumns(cs@_*) => {
+      checkColumnsExistence(cs)
+      checkValidJoin(cs)
+    }
+  }
+
+  protected def checkValidJoin(columns: Seq[NamedColumnRef]): Seq[NamedColumnRef] = {
+    val allColumnNames = tableDef.allColumns.map(_.columnName).toSet
```
pkolaczk: Unused

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -0,0 +1,152 @@
...
+
+    def checkSingleColumn(column: NamedColumnRef) = {
+      if (regularColumnNames.contains(column.columnName))
+        throw new IllegalArgumentException(s"Can't pushdown join on column $column because it is not part of the PRIMARY KEY")
+      column
+    }
+
+    //Make sure we have all of the clustering indexes between the 0th position and the max requested in the join
+    val chosenClusteringColumns = tableDef.clusteringColumns
+      .filter(tableCol => columns.exists(joinCol => tableCol.columnName == joinCol.columnName))
```
pkolaczk: Isn't it equivalent to `tableDef.clusteringColumns.filter(c => joinColumnNames.contains(c.columnName))`

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -0,0 +1,152 @@
...
+    }
+  }
+
+  protected def checkValidJoin(columns: Seq[NamedColumnRef]): Seq[NamedColumnRef] = {
+    val allColumnNames = tableDef.allColumns.map(_.columnName).toSet
+    val regularColumnNames = tableDef.regularColumns.map(_.columnName).toSet
+    val partitionKeyColumnNames = tableDef.partitionKey.map(_.columnName).toSet
+    val joinColumnNames = columns.map(_.columnName).toSet
+
+    def checkSingleColumn(column: NamedColumnRef) = {
```
pkolaczk: The primary purpose of this method is to eventually throw an exception. Therefore why not make it return Unit?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -0,0 +1,152 @@
...
+          .filter(cc => missingIndexes.contains(cc.componentIndex.get))
+          .map(_.columnName).mkString(", ")
+        throw new IllegalArgumentException(s"Can't pushdown join on column ${maxCol} without also specifying [ ${otherCols} ]")
+      }
+    }
+    val missingPartitionKeys = partitionKeyColumnNames -- joinColumnNames
+    if (missingPartitionKeys.size != 0) {
+      throw new IllegalArgumentException(s"Can't join without the full partition key. Missing: [ ${missingPartitionKeys} ]")
+    }
+    columns.map(checkSingleColumn)
```
pkolaczk: ... and use columns.foreach(checkSingleColumn)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -0,0 +1,152 @@
...
+
+    //Make sure we have all of the clustering indexes between the 0th position and the max requested in the join
+    val chosenClusteringColumns = tableDef.clusteringColumns
+      .filter(tableCol => columns.exists(joinCol => tableCol.columnName == joinCol.columnName))
+    if (chosenClusteringColumns.size != 0) {
+      val chosenClusteringIndexes = chosenClusteringColumns.flatMap(col => col.componentIndex)
+      val maxIndex = chosenClusteringIndexes.max
+      val requiredIndexes = 0 to maxIndex
+      val missingIndexes = requiredIndexes.toSet -- chosenClusteringIndexes.toSet
+      if (missingIndexes.size != 0) {
```
pkolaczk: This code seems to check if a sequence of selected columns, ordered by component index is a prefix of the clustering columns sequence. The clustering columns are already ordered by component index. If so, this can be written in a much more explicit and readable way:

```scala
if (!tableDef.clusteringColumns.startsWith(chosenClusteringColumns)) {
  val requiredColumns = tableDef.clusteringColumns.takeWhile(_.componentIndex <= chosenClusteringColumns.last.componentIndex)
  val missingColumns = requiredColumns.toSet -- chosenClusteringColumns
  // report error
}
```
You also don't need to check if any clustering column was selected at all because empty sequence is a valid prefix of every sequence.

RussellSpitzer: That looks really nice! Much cleaner than what I had

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -0,0 +1,152 @@
...
+        val maxCol = tableDef.clusteringColumns
+          .filter(cc => cc.componentIndex.get == maxIndex)
+          .map(_.columnName).mkString(", ")
+        val otherCols = tableDef.clusteringColumns
+          .filter(cc => missingIndexes.contains(cc.componentIndex.get))
+          .map(_.columnName).mkString(", ")
+        throw new IllegalArgumentException(s"Can't pushdown join on column ${maxCol} without also specifying [ ${otherCols} ]")
```
pkolaczk: No need to enclose single variables in string interpolation in {}, $ alone is enough.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
RussellSpitzer: Are we guaranteed that the last element of chosenClusteringColumns will have the greatest component index? Or do we need to sort that on component index first

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -0,0 +1,152 @@
...
+  def on(joinColumns: ColumnSelector): CassandraJoinRDD[O, N] = {
+    new CassandraJoinRDD[O, N](prev, keyspaceName, tableName, connector, columnNames, joinColumns, where, readConf).asInstanceOf[this.type]
+  }
+
+  //We need to make sure we get selectedColumnNames before serialization so that our RowReader is
+  //built
+  private val singleKeyCqlQuery: (String) = {
+    require(tableDef.partitionKey.map(_.columnName).exists(
+      partitionKey => where.predicates.exists(_.contains(partitionKey))) == false,
+      "Partition keys are not allowed in the .where() clause of a Cassandra Join")
```
pkolaczk: Checking for substring existence is risky. What if there is a predicate accidentally containing partition key column name in a string, i.e. within single quotes? The string comparison would match, but the predicate is not really using the partition key.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -0,0 +1,152 @@
...
+    query
+  }
+
+  /**
+   * When computing a CassandraPartitionKeyRDD the data is selected via single CQL statements
+   * from the specified C* Keyspace and Table. This will be preformed on whatever data is
+   * avaliable in the previous RDD in the chain.
+   * @param split
+   * @param context
+   * @return
```
pkolaczk: Cosmetics: don't include a param / return in ScalaDoc if you don't intend to document it.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
pkolaczk: You might use a where predicate parser that Artem wrote some time ago to do it better.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -0,0 +1,152 @@
...
+   */
+  override def compute(split: Partition, context: TaskContext): Iterator[(O, N)] = {
+    logDebug(s"Query::: $singleKeyCqlQuery")
```
pkolaczk: It was already logged. Also, please keep logging messages consistent. Here it is prefixed with "Query", in the previous case it just logged naked query without any prefix.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -0,0 +1,152 @@
...
+
+import scala.reflect.ClassTag
+
+// O[ld] Is the type of the left Side RDD, N[ew] the type of the right hand side Results
+/**
+ * An RDD that will do a selecting join between `prev:RDD` and the specified Cassandra Table
+ * This will perform individual selects to retrieve the rows from Cassandra and will take
+ * advantage of RDD's that have been partitioned with the [[com.datastax.spark.connector.rdd.partitioner.ReplicaPartitioner]]
+ */
+class CassandraJoinRDD[O, N] private[connector](prev: RDD[O],
```
pkolaczk: O for a type parameter is not a very fortunate choice. It is easy to mistake it for 0 (zero).

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraReader.scala
RussellSpitzer: Yeah it is mostly the code to generate the RowReader for the given CassandraRDD, So maybe "CassandraRowReader[R]" This way it is written Class with CassandraRowReader[R]

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
RussellSpitzer: Left and Right would be my choice, I wasn't sure about style on type parameters.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -0,0 +1,152 @@
...
+    countingIterator
+  }
+
+  private def fetchIterator(session: Session, bsb: BoundStatementBuilder[O], lastIt: Iterator[O]): Iterator[(O, N)] = {
+    val columnNamesArray = selectedColumnNames.map(_.selectedAs).toArray
+    implicit val pv = protocolVersion(session)
+    lastIt.map(leftSide => (leftSide, bsb.bind(leftSide))).flatMap { case (leftSide, boundStmt) =>
+      val rs = session.execute(boundStmt)
+      val iterator = new PrefetchingResultSetIterator(rs, fetchSize)
+      val result = iterator.map(rightSide => (leftSide, rowTransformer.read(rightSide, columnNamesArray)))
```
pkolaczk: Don't call it rowTransformer. This is probably a leftover from some refactoring and I we must have not noticed this name wasn't properly updated. Call it rowReader (as the class).

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraReader.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  protected def consistencyLevel = readConf.consistencyLevel
+
+  /**
+   * rtf:RowReaderFactory and rct should be included as implict parameters in the constructor
+   * of the class implementing this trait see [[CassandraTableScanRDD]]
+   */
+  protected val rtf: RowReaderFactory[R]
+  protected val rct: ClassTag[R]
+  protected lazy val rowTransformer: RowReader[R] = rtf.rowReader(tableDef)
```
pkolaczk: Call it rowReader (I'm sorry we left it incorrectly named).

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraTableScan.scala
```diff
@@ -47,12 +57,12 @@ case class CassandraTableScan(
 
   private[this] def predicateOperator(predicate: Expression): String = {
     predicate match {
-      case _: EqualTo =>            "="
-      case _: LessThan =>           "<"
-      case _: LessThanOrEqual =>    "<="
-      case _: GreaterThan =>        ">"
+      case _: EqualTo => "="
```
pkolaczk: They were so nicely aligned... :(

RussellSpitzer: Sorry! I'll put them back

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -0,0 +1,152 @@
...
+      logDebug(f"Fetched ${countingIterator.count} rows from $keyspaceName.$tableName for partition ${split.index} in $duration%.3f s.")
+      session.close()
+    }
+    countingIterator
+  }
+
+  private def fetchIterator(session: Session, bsb: BoundStatementBuilder[O], lastIt: Iterator[O]): Iterator[(O, N)] = {
+    val columnNamesArray = selectedColumnNames.map(_.selectedAs).toArray
+    implicit val pv = protocolVersion(session)
+    lastIt.map(leftSide => (leftSide, bsb.bind(leftSide))).flatMap { case (leftSide, boundStmt) =>
```
pkolaczk: Code style: this is a nice candidate for a for comprehension.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/ReplicaMapper.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  private def quote(name: String): String =
+    "\"" + name + "\""
+
+  /**
+   * This query is only used to build a prepared statement so we can more easily extract
+   * partition tokens from tables. We prepare a statement of the form SELECT * FROM keyspace.table
+   * where x= .... This statement is never executed.
+   */
+  private lazy val querySelectUsingOnlyParititonKeys: String = {
```
pkolaczk: typo: Paritition -> Partition

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/ReplicaMapper.scala
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * Pairs each piece of data with the Cassandra Replicas which that data would be found on
+   * @param data A source of data which can be bound to a statement by BatchStatementBuilder
+   * @return an Iterator over the same data keyed by the replica's ip addresses
+   */
+  def keyByReplicas(data: Iterator[T]): Iterator[(scala.collection.immutable.Set[InetAddress], T)] = {
+    connector.withClusterDo { cluster =>
+      connector.withSessionDo { session =>
```
pkolaczk: very minor thing: no need to nest this way; session already has a reference to the cluster

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
```diff
@@ -0,0 +1,51 @@
...
+   */
+  override def getPartition(key: Any): Int = {
+    val replicaSet = key.asInstanceOf[Set[InetAddress]].toVector
+    val endpoint = replicaSet(rand.nextInt(replicaSet.size))
+    hostMap.getOrElse(endpoint, hostMap(hosts(rand.nextInt(numHosts))))(rand.nextInt(partitionsPerReplicaSet))
+  }
+
+  override def numPartitions: Int = partitionsPerReplicaSet * numHosts
+
+  def getEndpointParititon(partition: Partition): ReplicaPartition = {
```
pkolaczk: typo: Paritition

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
```diff
@@ -0,0 +1,51 @@
...
+ * @param connector
+ */
+class ReplicaPartitioner(partitionsPerReplicaSet: Int, connector: CassandraConnector) extends Partitioner {
+  /* TODO We Need JAVA-312 to get sets of replicas instead of single endpoints. Once we have that we'll be able to
+  build a map of Set[ip,ip,...] => Index before looking at our data and give the all options for the preferred location
+   for a partition*/
+  val hosts = connector.hosts.toVector
+  val numHosts = hosts.size
+  val partitionIndexes = (0 until partitionsPerReplicaSet * numHosts).grouped(partitionsPerReplicaSet).toList
+  val hostMap = hosts zip partitionIndexes toMap
```
pkolaczk: toMap is in postfix position here; postfix syntax is going to be deprecated (because it is errorprone). Preferred style:

```scala
(hosts zip partitionIndexes).toMap
```

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
```diff
@@ -0,0 +1,51 @@
...
+  // Ip1 -> (0,1,2,..), Ip2 -> (11,12,13...)
+  val indexMap = hostMap flatMap { case (ip, partitions) => partitions.map(partition => (partition, ip))}
```
pkolaczk: for comprehension should be clearer:

```scala
for ((ip, partitions) <- hostMap; p <- partitions) yield (p, ip)
```

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
```diff
@@ -0,0 +1,51 @@
...
+  val rand = new java.util.Random()
+
+  /**
+   * Given a set of endpoints, pick a random endpoint, and then a random partition owned by that endpoint. If the
+   * requested host doesn't exist chose another random host.
+   * @param key A Set[InetAddress] of replicas for this Cassandra Partition
+   * @return An integer between 0 and numPartitions
+   */
+  override def getPartition(key: Any): Int = {
+    val replicaSet = key.asInstanceOf[Set[InetAddress]].toVector
```
pkolaczk: Shouldn't we check if it really is the correct type and eventually throw a meaningful message?

RussellSpitzer: Not sure how we can do that with the type erasure here, I'll push a version with an unchecked match but thats all I can think of at the moment

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableRowReader.scala
```diff
@@ -0,0 +1,175 @@
...
+import scala.reflect.ClassTag
+
+/**
+ * Used to get a RowReader of type [R] for transforming the rows of a particular Cassandra table into
+ * scala objects. Performs necessary checking of the schema and output class to make sure they are
+ * compatible.
+ * @see [[CassandraTableScanRDD]]
+ * @see [[CassandraJoinRDD]]
+ */
+trait CassandraTableRowReader[R] {
```
pkolaczk: Hmm, this is still confusing, because the name suggests this object is a RowReader. While in fact it only provides a RowReader to the RDD implementation. Maybe call it `CassandraTableRowReaderProvider`?

pkolaczk:
One more thing - All implementations have keyspaceName and tableName, even the `EmptyCassandraRDD`. However, `CassandraRDD` has no `keyspaceName` and `tableName` abstract properties. I think we need to add them there (with no implementation, just an abstract def).

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/EmptyCassandraRDD.scala
```diff
@@ -0,0 +1,45 @@
...
+                           val keyspaceName: String,
+                           val tableName: String,
+                           val columnNames: ColumnSelector = AllColumns,
+                           val where: CqlWhereClause = CqlWhereClause.empty,
+                           val limit: Option[Long] = None,
+                           val clusteringOrder: Option[ClusteringOrder] = None,
+                           val readConf: ReadConf = ReadConf())
+                          (implicit val rct: ClassTag[R])
+  extends CassandraRDD[R](sc, Seq.empty) {
+  override protected def copy(columnNames: ColumnSelector = columnNames,
```
pkolaczk: Cosmetics: Can you move this line a little lower? Visually it looks as if it belonged to the class header.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/EmptyCassandraRDD.scala
```diff
@@ -0,0 +1,45 @@
...
+                              where: CqlWhereClause = where, limit: Option[Long] = limit, clusteringOrder: Option[ClusteringOrder] = None,
+                              readConf: ReadConf = readConf, connector: CassandraConnector = connector) =
+    new EmptyCassandraRDD[R](sc, keyspaceName, tableName, columnNames, where, limit, clusteringOrder, readConf).asInstanceOf[this.type]
```
pkolaczk: Additionally, if you have multiple parameters that don't fit in a single line, put one parameter per line.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/CountingIterator.scala
```diff
@@ -1,13 +1,16 @@
 package com.datastax.spark.connector.util
 
-/** Counts elements fetched form the underlying iterator. */
-class CountingIterator[T](iterator: Iterator[T]) extends Iterator[T] {
+/** Counts elements fetched form the underlying iterator. Limit causes iterator to terminate early */
+class CountingIterator[T](iterator: Iterator[T], limit: Option[Long] = None) extends Iterator[T] {
```
RussellSpitzer: I want to draw attention to this change, I added this to minimize over-requesting from C* when we have a known limit.

pkolaczk:
Two tests are failing:
```
 - should be lazy and must not throw IOException if the table was not found at the RDD initialization time *** FAILED *** (2 milliseconds)
[info]   java.io.IOException: Table not found: read_test.unknown_table
[info]   at com.datastax.spark.connector.rdd.CassandraTableRowReader$class.tableDef(CassandraTableRowReader.scala:48)
[info]   at com.datastax.spark.connector.rdd.CassandraTableScanRDD.tableDef$lzycompute(CassandraTableScanRDD.scala:47)
[info]   at com.datastax.spark.connector.rdd.CassandraTableScanRDD.tableDef(CassandraTableScanRDD.scala:47)
[info]   at com.datastax.spark.connector.rdd.CassandraTableRowReader$class.rowReader(CassandraTableRowReader.scala:43)
[info]   at com.datastax.spark.connector.rdd.CassandraTableScanRDD.rowReader$lzycompute(CassandraTableScanRDD.scala:47)
[info]   at com.datastax.spark.connector.rdd.CassandraTableScanRDD.rowReader(CassandraTableScanRDD.scala:47)
[info]   at com.datastax.spark.connector.rdd.CassandraTableRowReader$class.$init$(CassandraTableRowReader.scala:105)
[info]   at com.datastax.spark.connector.rdd.CassandraTableScanRDD.<init>(CassandraTableScanRDD.scala:47)
[info]   at com.datastax.spark.connector.SparkContextFunctions.cassandraTable(SparkContextFunctions.scala:50)
[info]   at com.datastax.spark.connector.rdd.CassandraRDDSpec$$anonfun$37.apply$mcV$sp(CassandraRDDSpec.scala:367)
[info]   at com.datastax.spark.connector.rdd.CassandraRDDSpec$$anonfun$37.apply(CassandraRDDSpec.scala:367)
[info]   at com.datastax.spark.connector.rdd.CassandraRDDSpec$$anonfun$37.apply(CassandraRDDSpec.scala:367)
[info]   at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.FlatSpecLike$$anon$1.apply(FlatSpecLike.scala:1647)
[info]   at org.scalatest.Suite$class.withFixture(Suite.scala:1122)
[info]   at org.scalatest.FlatSpec.withFixture(FlatSpec.scala:1683)
[info]   at org.scalatest.FlatSpecLike$class.invokeWithFixture$1(FlatSpecLike.scala:1644)
[info]   at org.scalatest.FlatSpecLike$$anonfun$runTest$1.apply(FlatSpecLike.scala:1656)
[info]   at org.scalatest.FlatSpecLike$$anonfun$runTest$1.apply(FlatSpecLike.scala:1656)
[info]   at org.scalatest.SuperEngine.runTestImpl(Engine.scala:306)
[info]   at org.scalatest.FlatSpecLike$class.runTest(FlatSpecLike.scala:1656)
[info]   at org.scalatest.FlatSpec.runTest(FlatSpec.scala:1683)
[info]   at org.scalatest.FlatSpecLike$$anonfun$runTests$1.apply(FlatSpecLike.scala:1714)
[info]   at org.scalatest.FlatSpecLike$$anonfun$runTests$1.apply(FlatSpecLike.scala:1714)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:413)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:401)
[info]   at scala.collection.immutable.List.foreach(List.scala:318)
[info]   at org.scalatest.SuperEngine.traverseSubNodes$1(Engine.scala:401)
[info]   at org.scalatest.SuperEngine.org$scalatest$SuperEngine$$runTestsInBranch(Engine.scala:390)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:427)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:401)
[info]   at scala.collection.immutable.List.foreach(List.scala:318)
[info]   at org.scalatest.SuperEngine.traverseSubNodes$1(Engine.scala:401)
[info]   at org.scalatest.SuperEngine.org$scalatest$SuperEngine$$runTestsInBranch(Engine.scala:396)
[info]   at org.scalatest.SuperEngine.runTestsImpl(Engine.scala:483)
[info]   at org.scalatest.FlatSpecLike$class.runTests(FlatSpecLike.scala:1714)
[info]   at org.scalatest.FlatSpec.runTests(FlatSpec.scala:1683)
[info]   at org.scalatest.Suite$class.run(Suite.scala:1424)
[info]   at org.scalatest.FlatSpec.org$scalatest$FlatSpecLike$$super$run(FlatSpec.scala:1683)
[info]   at org.scalatest.FlatSpecLike$$anonfun$run$1.apply(FlatSpecLike.scala:1760)
[info]   at org.scalatest.FlatSpecLike$$anonfun$run$1.apply(FlatSpecLike.scala:1760)
[info]   at org.scalatest.SuperEngine.runImpl(Engine.scala:545)
[info]   at org.scalatest.FlatSpecLike$class.run(FlatSpecLike.scala:1760)
[info]   at org.scalatest.FlatSpec.run(FlatSpec.scala:1683)
[info]   at org.scalatest.tools.Framework.org$scalatest$tools$Framework$$runSuite(Framework.scala:466)
[info]   at org.scalatest.tools.Framework$ScalaTestTask.execute(Framework.scala:677)
[info]   at sbt.ForkMain$Run$2.call(ForkMain.java:294)
[info]   at sbt.ForkMain$Run$2.call(ForkMain.java:284)
[info]   at java.util.concurrent.FutureTask.run(FutureTask.java:262)
[info]   at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
[info]   at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
[info]   at java.lang.Thread.run(Thread.java:745)
 - should allow to use empty RDD on undefined table *** FAILED *** (1 millisecond)
[info]   java.io.IOException: Table not found: unknown_ks.unknown_table
[info]   at com.datastax.spark.connector.rdd.CassandraTableRowReader$class.tableDef(CassandraTableRowReader.scala:48)
[info]   at com.datastax.spark.connector.rdd.CassandraTableScanRDD.tableDef$lzycompute(CassandraTableScanRDD.scala:47)
[info]   at com.datastax.spark.connector.rdd.CassandraTableScanRDD.tableDef(CassandraTableScanRDD.scala:47)
[info]   at com.datastax.spark.connector.rdd.CassandraTableRowReader$class.rowReader(CassandraTableRowReader.scala:43)
[info]   at com.datastax.spark.connector.rdd.CassandraTableScanRDD.rowReader$lzycompute(CassandraTableScanRDD.scala:47)
[info]   at com.datastax.spark.connector.rdd.CassandraTableScanRDD.rowReader(CassandraTableScanRDD.scala:47)
[info]   at com.datastax.spark.connector.rdd.CassandraTableRowReader$class.$init$(CassandraTableRowReader.scala:105)
[info]   at com.datastax.spark.connector.rdd.CassandraTableScanRDD.<init>(CassandraTableScanRDD.scala:47)
[info]   at com.datastax.spark.connector.SparkContextFunctions.cassandraTable(SparkContextFunctions.scala:50)
[info]   at com.datastax.spark.connector.rdd.CassandraRDDSpec$$anonfun$63.apply$mcV$sp(CassandraRDDSpec.scala:640)
[info]   at com.datastax.spark.connector.rdd.CassandraRDDSpec$$anonfun$63.apply(CassandraRDDSpec.scala:639)
[info]   at com.datastax.spark.connector.rdd.CassandraRDDSpec$$anonfun$63.apply(CassandraRDDSpec.scala:639)
[info]   at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.FlatSpecLike$$anon$1.apply(FlatSpecLike.scala:1647)
[info]   at org.scalatest.Suite$class.withFixture(Suite.scala:1122)
[info]   at org.scalatest.FlatSpec.withFixture(FlatSpec.scala:1683)
[info]   at org.scalatest.FlatSpecLike$class.invokeWithFixture$1(FlatSpecLike.scala:1644)
[info]   at org.scalatest.FlatSpecLike$$anonfun$runTest$1.apply(FlatSpecLike.scala:1656)
[info]   at org.scalatest.FlatSpecLike$$anonfun$runTest$1.apply(FlatSpecLike.scala:1656)
[info]   at org.scalatest.SuperEngine.runTestImpl(Engine.scala:306)
[info]   at org.scalatest.FlatSpecLike$class.runTest(FlatSpecLike.scala:1656)
[info]   at org.scalatest.FlatSpec.runTest(FlatSpec.scala:1683)
[info]   at org.scalatest.FlatSpecLike$$anonfun$runTests$1.apply(FlatSpecLike.scala:1714)
[info]   at org.scalatest.FlatSpecLike$$anonfun$runTests$1.apply(FlatSpecLike.scala:1714)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:413)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:401)
[info]   at scala.collection.immutable.List.foreach(List.scala:318)
[info]   at org.scalatest.SuperEngine.traverseSubNodes$1(Engine.scala:401)
[info]   at org.scalatest.SuperEngine.org$scalatest$SuperEngine$$runTestsInBranch(Engine.scala:390)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:427)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:401)
[info]   at scala.collection.immutable.List.foreach(List.scala:318)
[info]   at org.scalatest.SuperEngine.traverseSubNodes$1(Engine.scala:401)
[info]   at org.scalatest.SuperEngine.org$scalatest$SuperEngine$$runTestsInBranch(Engine.scala:396)
[info]   at org.scalatest.SuperEngine.runTestsImpl(Engine.scala:483)
[info]   at org.scalatest.FlatSpecLike$class.runTests(FlatSpecLike.scala:1714)
[info]   at org.scalatest.FlatSpec.runTests(FlatSpec.scala:1683)
[info]   at org.scalatest.Suite$class.run(Suite.scala:1424)
[info]   at org.scalatest.FlatSpec.org$scalatest$FlatSpecLike$$super$run(FlatSpec.scala:1683)
[info]   at org.scalatest.FlatSpecLike$$anonfun$run$1.apply(FlatSpecLike.scala:1760)
[info]   at org.scalatest.FlatSpecLike$$anonfun$run$1.apply(FlatSpecLike.scala:1760)
[info]   at org.scalatest.SuperEngine.runImpl(Engine.scala:545)
[info]   at org.scalatest.FlatSpecLike$class.run(FlatSpecLike.scala:1760)
[info]   at org.scalatest.FlatSpec.run(FlatSpec.scala:1683)
[info]   at org.scalatest.tools.Framework.org$scalatest$tools$Framework$$runSuite(Framework.scala:466)
[info]   at org.scalatest.tools.Framework$ScalaTestTask.execute(Framework.scala:677)
[info]   at sbt.ForkMain$Run$2.call(ForkMain.java:294)
[info]   at sbt.ForkMain$Run$2.call(ForkMain.java:284)
[info]   at java.util.concurrent.FutureTask.run(FutureTask.java:262)
[info]   at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
[info]   at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
[info]   at java.lang.Thread.run(Thread.java:745)
```

pkolaczk:
+1
All tests green.
There is a small compile warning related to unchecked pattern match - I fixed this in a separate PR. 
https://github.com/datastax/spark-cassandra-connector/pull/579
Please squash commits before merge into master.

PR #536, opened 2015-02-04
SPARKC-27: Refactored the read path and the write path in order to provide read/write metrics for the Spark metrics system

jacek-lewandowski:


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/metrics/InputMetricsUpdater.scala
```diff
@@ -0,0 +1,80 @@
...
+    cnt += 1
+    if (cnt == groupSize)
+      update()
+    row
+  }
+
+  @inline
+  private def update(): Unit = {
+    CassandraConnectorSource.readRowMeter.mark(cnt)
+    CassandraConnectorSource.readByteMeter.mark(metrics.bytesRead - dataLength)
```
RussellSpitzer: I'm a little lost on what's happening here. The way i'm reading this I see the following happening:

datalength is set to the current number of bytes read
datalength is increased to by the size of each row
when cnt has reached groupSize:
     we update the read byteMeter by the bytesRead - (previously obtained)bytesRead+(the size of all the rows we read before) 
     then we reset datalength to (current) bytesRead

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/metrics/OutputMetricsUpdater.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  def batchSucceeded(stmt: RichStatement, submissionTimestamp: Long, executionTimestamp: Long): Unit = {
+    val t = System.nanoTime()
+    CassandraConnectorSource.writeBatchTimer.update(t - executionTimestamp, TimeUnit.NANOSECONDS)
+    CassandraConnectorSource.writeBatchWaitTimer.update(executionTimestamp - submissionTimestamp, TimeUnit.NANOSECONDS)
+    CassandraConnectorSource.writeRowMeter.mark(stmt.rowsCount)
+    CassandraConnectorSource.writeByteMeter.mark(stmt.bytesCount)
+    CassandraConnectorSource.writeSuccessCounter.inc()
+    mutex.acquire()
+    outputMetrics.bytesWritten += stmt.bytesCount
```
RussellSpitzer: I wonder why this isn't atomic in the Spark API ...?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/metrics/OutputMetricsUpdater.scala
```diff
@@ -0,0 +1,73 @@
...
+  lazy val detailedMetricsEnabled =
+    SparkEnv.get.conf.getBoolean("spark.cassandra.output.metrics", defaultValue = true)
+
+  def apply(taskContext: TaskContext): OutputMetricsUpdater = {
+    CassandraConnectorSource.ensureInitialized
+
+    if (detailedMetricsEnabled) {
+      val tm = taskContext.taskMetrics()
+      if (tm.outputMetrics.isEmpty || tm.outputMetrics.get.writeMethod != DataWriteMethod.Hadoop)
+        tm.outputMetrics = Some(new OutputMetrics(DataWriteMethod.Hadoop))
```
RussellSpitzer: This is going to cause us no-end of support calls, is there any way we can extend the enumeration and put in a C* option on our end?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/metrics/InputMetricsUpdater.scala
```diff
@@ -0,0 +1,80 @@
...
+object InputMetricsUpdater {
+  lazy val detailedMetricsEnabled =
+    SparkEnv.get.conf.getBoolean("spark.cassandra.input.metrics", defaultValue = true)
+
+  def apply(taskContext: TaskContext, groupSize: Int): InputMetricsUpdater = {
+    CassandraConnectorSource.ensureInitialized
+
+    if (detailedMetricsEnabled) {
+      val tm = taskContext.taskMetrics()
+      if (tm.inputMetrics.isEmpty || tm.inputMetrics.get.readMethod != DataReadMethod.Hadoop)
```
RussellSpitzer: Same comment with the DataWriteMethod. I'd rather not if at any point we make any statement containing "hadoop" to end users

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/PrefetchingResultSetIterator.scala
```diff
@@ -8,16 +12,27 @@ import com.datastax.driver.core.{Row, ResultSet}
...
   override def hasNext = iterator.hasNext
 
   private[this] def maybePrefetch(): Unit = {
-    if (!resultSet.isFullyFetched && resultSet.getAvailableWithoutFetching < prefetchWindowSize)
-      resultSet.fetchMoreResults()
+    if (!resultSet.isFullyFetched && resultSet.getAvailableWithoutFetching < prefetchWindowSize) {
+      val t0 = System.nanoTime()
+      val future: ListenableFuture[Void] = resultSet.fetchMoreResults()
+      if (timer.isDefined)
+        Futures.addCallback(future, new FutureCallback[Void] {
```
RussellSpitzer: I feel a little strange about us recording request latencies and the driver also doing it at the same time (http://www.datastax.com/drivers/java/2.1/com/datastax/driver/core/Metrics.htm). I understand this lets us separate out read's from writes but maybe it would be beneficial if we try to get the separation in upstream in the Java Driver rather than redoing the work in the connector?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/metrics/OutputMetricsUpdater.scala
jacek-lewandowski: I suppose that this isn't atomic because this is a part of task metrics, which should regard a single task. This means - one thread. However, what we do in our writer - we use multiple threads to write - that is why we need additional synchronisation.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/metrics/OutputMetricsUpdater.scala
jacek-lewandowski: Unfortunately, AFAIK we can't :(
Also - Hadoop is the only value in DataWriteMethod enum (weird). That's why I used it for both read and write. Anyway, for writes, it isn't mentioned anywhere that this is Hadoop. It is just "output". For reads, Spark UI writes "hadoop" in the parentheses.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/PrefetchingResultSetIterator.scala
jacek-lewandowski: Well, we can try at least. But as far as I can see, it is more difficult to add something to Java Driver than to Spark ;)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/metrics/InputMetricsUpdater.scala
jacek-lewandowski: ahh, i confused something during refactoring - in the for loop there should be metrics.bytesRead += ... instead of dataLength += ....

jacek-lewandowski: good catch

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/metrics/InputMetricsUpdater.scala
jacek-lewandowski: explained in the other comment

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/PrefetchingResultSetIterator.scala
RussellSpitzer: I'm fine with putting that off for later, and keeping our own metrics for now.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/metrics/OutputMetricsUpdater.scala
RussellSpitzer: Yeah that UI is gonna come back to haunt us, I guarantee it. But since we can't really do anything without patching Spark lets go with this and maybe open a spark ticket to add Cassandra :)

RussellSpitzer:
+1

PR #537, opened 2015-02-04
SPARKC-29: Improve error message when SparkContext is null in CassandraR...

pkolaczk:
...DD after deserialization and the user wants to transform it.


helena:
LGTM

pkolaczk:
Merged into b1.1 and master as well.

PR #541, opened 2015-02-09
SPARKC-30: Upgrade to Spark 1.2.1

jacek-lewandowski:
- changed Spark dependency version
- updated CassandraCatalog
- unlocked CassandraSQL tests which where ignored before
- increased max perm size for test


alexliu68:
+1

PR #549, opened 2015-02-13
SPARKC-38: Save RDD to a new table.

pkolaczk:


spark-cassandra-connector/src/test/scala/com/datastax/spark/connector/cql/TableDefSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+            |  "c2" varchar,
+            |  "c3" varchar,
+            |  PRIMARY KEY (("c1"), "c2")
+            |)""".stripMargin
+        )
+      }
+
+      "it contains compound partition key and multiple clustering columns" in {
+        val column1 = ColumnDef("c1", PartitionKeyColumn, IntType)
+        val column2 = ColumnDef("c2", RegularColumn, VarCharType)
```
RussellSpitzer: This test is confusing to me, shouldn't this throw an exception since c2 is marked as a regular column but then used in the Partition key?

RussellSpitzer:
The columndef type being ignored by the TableDef Constructor but I think thats ok, I do think we should change the it tests so usage is clear though.  +1

pkolaczk:
All tests good.

PR #550, opened 2015-02-15
[SPARKC-60] Drop dependency on scala-compiler in published POM

redbaron:


pkolaczk:
Scala compiler is no longer on the dependency list.

PR #552, opened 2015-02-17
SPARKC-9: Allow to specify column to property mapping without explicit creation of ColumnMapper

jacek-lewandowski:


doc/4_mapper.md
```diff
@@ -58,6 +58,29 @@ class WordCount extends Serializable {
...
+```
+
+`as` method can be used for any type of projected value: normal column, TTL or write time:
+
+```scala
+sc.cassandraTable[SomeClass]("test", "table").select(
+    "no_alias",
+    "simple" as "simpleProp",
+    "simple".ttl as "simplePropTTL",
+    "simple"writeTime as "simpleWriteTime")
```
pkolaczk: missing "."

doc/4_mapper.md
```diff
@@ -58,6 +58,29 @@ class WordCount extends Serializable {
...
+Example:
+Say, we have a table with columns `word TEXT` and `num INT`. We would like to map rows from this
+table to the objects of class with fields `word: String` and `count: Int`:
+
+```scala
+case class WordCount(word: String, count: Int)
+val result = sc.cassandraTable[WordCount]]("test", "words").select("word", "num" as "count").collect()
+```
+
+`as` method can be used for any type of projected value: normal column, TTL or write time:
```
pkolaczk: The `as` method...

doc/7_java_api.md
```diff
@@ -138,6 +138,30 @@ is also one overloaded `mapRowTo` methods which allows to specify a custom `Colu
...
+Since 1.2, it is possible to easily provide custom column name to property name translation by
+`select` method.
+
+Example:
+Say we have a table `people2` with columns `id INT`, `last_name TEXT`, `date_of_birth TIMESTAMP` and
+we want to map the rows of this table to objects of `Person` class.
+
+```java
+CassandraJavaRDD<Person> rdd = javaFunctions(sc).cassandraTable("ks", "people2", mapRowTo(Person.class)).select(
+        plain("id"),
```
pkolaczk: Why is it called "plain"? Little surprising to me. Wouldn't it be more readable to call it "column" ?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/ColumnSelector.scala
```diff
@@ -2,9 +2,19 @@ package com.datastax.spark.connector
...
 
-sealed trait ColumnSelector
-case object AllColumns extends ColumnSelector
-case class SomeColumns(columns: NamedColumnRef*) extends ColumnSelector
+sealed trait ColumnSelector {
+  def toAliasMap: Map[String, String]
+}
+
+case object AllColumns extends ColumnSelector {
+  override def toAliasMap: Map[String, String] = Map.empty
```
pkolaczk: shouldn't it return a Map with default identity? So always `toAliasMap(x) == x`

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/ColumnSelector.scala
```diff
@@ -2,9 +2,19 @@ package com.datastax.spark.connector
 
 import scala.language.implicitConversions
 
-sealed trait ColumnSelector
-case object AllColumns extends ColumnSelector
-case class SomeColumns(columns: NamedColumnRef*) extends ColumnSelector
+sealed trait ColumnSelector {
+  def toAliasMap: Map[String, String]
```
pkolaczk: I'd call it `aliasMap`. I don't think alias map preserves all the information that is the ColumnSelector

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/ColumnRef.scala
```diff
@@ -24,24 +27,30 @@ object NamedColumnRef {
 }
 
 /** References a column by name. */
-case class ColumnName(columnName: String) extends NamedColumnRef {
+case class ColumnName(columnName: String, alias: Option[String] = None) extends NamedColumnRef {
   val cql = s""""$columnName""""
   val selectedAs = columnName
```
pkolaczk: If an alias is set, shouldn't selectedAs just return the alias?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/ColumnRef.scala
```diff
@@ -17,6 +17,9 @@ sealed trait NamedColumnRef extends ColumnRef {
   /** Returns a name of the selection as it is seen in the result set. Most likely this is going to be
     * used when providing custom column name to field name mapping. */
   def selectedAs: String
```
pkolaczk: `selectedAs` and `alias` are confusing. At first I thought that `selectedAs` applies to CassandraRDD select. However, in fact, this is how we identify this column in the ResultSet coming from Cassandra. So maybe rename `selectedAs` to something more unambiguous, like `resultSetColumnName` or `selectedFromCassandraAs`?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/ColumnRef.scala
```diff
@@ -17,6 +17,9 @@ sealed trait NamedColumnRef extends ColumnRef {
...
+
+  /** Returns an alias for the referenced column */
+  def alias: Option[String]
```
pkolaczk: I don't like the fact it is an `Option`. It doesn't tell me, what to do if it is not set. Maybe it should be just a `String` set to `selectedAs` if alias was not given?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/mapper/ColumnMapper.scala
```diff
@@ -23,7 +23,7 @@ import scala.reflect.ClassTag
   * }}}
   */
 trait ColumnMapper[T] extends Serializable {
-  def columnMap(tableDef: TableDef): ColumnMap
+  def columnMap(tableDef: TableDef, aliases: Map[String, String] = Map.empty): ColumnMap
```
pkolaczk: I'm not sure applying aliases to table should be really a concern of a `ColumnMapper`. If we're selecting a subset of columns from a table, why not pass a projection of the original table as a tableDef param, i.e. apply selection + aliases to the original tableDef and then pass it to the column mapper to obtain the column map? This way we wouldn't need to put aliasing logic in all ColumnMappers, and ColumnMappers would be simpler.

jacek-lewandowski: It would be nice, but I don't see how it would work in the case when functions are used - functions, like TTL, write time or count are not columns.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/ColumnSelector.scala
jacek-lewandowski: I could write something like `Map.empty.withDefault(x => x)` but this is not the same. Later in the code, there is a test whether the alias is defined for the field. If it is not, the convention is used. If I implemented the default identity, I wouldn't be able to fallback to convention.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/ColumnRef.scala
jacek-lewandowski: This is an `Option` because alias is optional. And if it is not used, something else is tried - look at my above comment related to `AllColumns`.

pkolaczk:
+1, please rebase & squash

pkolaczk:
All tests green.

jacek-lewandowski:
Thanks

PR #554, opened 2015-02-17
SPARKC-32: Added support for pushing down order by clause

jacek-lewandowski:
With this patch:
- added withAscOrder and withDescOrder methods to CassandraRDD, CassandraJavaRDD and CassandraJavaPairRDD
- updated CQL generation method in CassandraRDD so that it can use the order by clause
- added test cases
- updated documentation and CHANGES.txt


pkolaczk:
+1

PR #555, opened 2015-02-17
SPARKC-57: Write throughput limiting

pkolaczk:


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/TableWriter.scala
```diff
@@ -97,42 +97,53 @@ class TableWriter[T] private (
...
+        bs.getRoutingKey.duplicate()
+    }
+  }
+
+  def newRateLimiter = {
+    if (!OutputMetricsUpdater.detailedMetricsEnabled && writeConf.throughputMBPS < WriteConf.DefaultThroughputMBPS)
+      logError(
+        s"Detailed metrics not enabled but write throughput limiting requested at ${writeConf.throughputMBPS} MB/s." +
+          "Throughput limiting will not function properly. Please set spark.cassandra.output.metrics to true.")
+    new RateLimiter(writeConf.throughputMBPS * 1000000, 1000000)
```
RussellSpitzer: I think we'll end up with some issues if we use the Decimal Byte counters for those watching the system, 1048576 may be better to mesh with OS monitoring tools.

pkolaczk: Hmm, ok, so you suggest switching to MiB/s ?
I wonder what C* is using as MB definition. 1024 kB or 1000 kB ? ;)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/TableWriter.scala
```diff
@@ -97,42 +97,53 @@ class TableWriter[T] private (
...
+        if (bs.getRoutingKey == null) {
+          bs.setRoutingKey(routingKeyGenerator(bs))
+        }
+        bs.getRoutingKey.duplicate()
+    }
+  }
+
+  def newRateLimiter = {
+    if (!OutputMetricsUpdater.detailedMetricsEnabled && writeConf.throughputMBPS < WriteConf.DefaultThroughputMBPS)
+      logError(
```
RussellSpitzer: Maybe we should just have writeConf.throughputMBPS automatically enable detailed metrics? (Or enable metrics and warn that it needed to enable metrics?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/TableWriter.scala
RussellSpitzer: https://github.com/apache/cassandra/search?utf8=%E2%9C%93&q=1024+*+1024&type=Code
Lots of bibytes :)

RussellSpitzer:
Other than my comments, looks good.

+1 Unit Tests / IT Tests

qqibrow:
I am looking for solution for write throttling and got here. A simple question here, have you ever considered using [RateLimiter](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/util/concurrent/RateLimiter.html) from Guava package?

pkolaczk:
Spark cassandra connector has builtin write throttling.
See the docs:
https://github.com/datastax/spark-cassandra-connector/blob/master/doc/reference.md#write-tuning-parameters

PR #556, opened 2015-02-18
SPARKC-52: Support for native Cassandra count

jacek-lewandowski:
With this patch:
- Added one more trait (SelectableColumnRef) in ColumnRef hierarchy to reflect columns without names
- Added CountColumn case class
- Refactored CassandraRDD, CassandraJavaRDD and CassandraJavaPairRDD
- added test cases
- CHANGES.txt

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,7 @@
  * Added support for write throughput limiting (SPARKC-57)
  * Added EmptyCassandraRDD (SPARKC-37)
  * Exposed authConf in CassandraConnector
+ * Overridden count() implementation in CassandraRDD which uses native Cassandra count (SPARKC-52)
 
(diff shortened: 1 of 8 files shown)


RussellSpitzer:
I think for clarity this should log a warning, or throw an exception if you try to call count on a method with a selectStatement already state.

{code}
scala> sc.cassandraTable("test","fun").select("c").count
{code}
Should alert the user, or forbid setting the select("c") since it semantically will not do anything.

RussellSpitzer:
maybeAdd Warning when trying to use select in conjunction with count
+1 On Code, 
+1 Integration Tests, Unit tests

pkolaczk:
I agree with @RussellSpitzer on select and count. Besides that, the code looks good to me. +1

PR #557, opened 2015-02-18
SPARKC-37: Added EmptyCassandraRDD implementation

jacek-lewandowski:
With this patch:
- Implemented EmptyCassandraRDD subclass of CassandraRDD
- Added a method to create EmptyCassandraRDD to CassandraRDD, CassandraJavaRDD and CassandraJavaPairRDD
- added test cases
- updated docs and CHANGES.txt


spark-cassandra-connector-java/src/main/java/com/datastax/spark/connector/japi/rdd/CassandraJavaRDD.java
```diff
@@ -110,4 +110,13 @@ public CassandraJavaRDD(CassandraRDD<R> rdd, ClassTag<R> classTag) {
     public <K> JavaPairRDD<K, Iterable<R>> spanBy(Function<R, K> f, ClassTag<K> keyClassTag) {
         return CassandraJavaUtil.javaFunctions(rdd()).spanBy(f, keyClassTag);
     }
+
+    /**
+     * Produces the empty CassandraRDD which has the same signature and properties, but it does not
+     * perform any validation and it does not even try to return any rows.
+     */
+    public CassandraJavaRDD<R> toEmptyCassandraRDD() {
+        CassandraRDD<R> newRDD = rdd().toEmptyCassandraRDD();
```
helena: This should stay in line with in the spark API:

      def makeRDD[T: ClassTag](seq: Seq[T], numSlices: Int = defaultParallelism): RDD[T]  

So:

    def makeCassandraRDD[T: ClassTag](seq: Seq[T], numSlices: Int = defaultParallelism): CassandraRDD[T]

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraRDD.scala
```diff
@@ -105,6 +105,10 @@ class CassandraRDD[R] private[connector] (
     copy(readConf = readConf)
   }
 
+  /** Produces the empty CassandraRDD which has the same signature and properties, but it does not
+    * perform any validation and it does not even try to return any rows. */
+  def toEmptyCassandraRDD = new EmptyCassandraRDD[R](sc, connector, keyspaceName, tableName)
```
helena: def makeCassandraRDD[T: ClassTag](seq: Seq[T], numSlices: Int = defaultParallelism): CassandraRDD[T] = {
             ....
       }

doc/2_loading.md
```diff
@@ -39,6 +39,22 @@ rdd.toArray.foreach(println)
...
+exists. To create one, just initialize `CassandraRDD` as usual and then call `toEmptyCassandraRDD`
+method on it. The method will return the `EmptyCassandraRDD` instance.
+
+Example:
+
+```scala
+// validation is deferred, so it is not triggered during rdd creation
+val rdd = sc.cassandraTable[SomeType]("ks", "not_existing_table")
+val emptyRDD = rdd.toEmptyCassandraRDD
+```
```
pkolaczk: Why do I need to have an existing CassandraRDD to make an empty one? 
I should be supposed to write `CassandraRDD.empty[T]` or something similar.

doc/2_loading.md
```diff
@@ -39,6 +39,22 @@ rdd.toArray.foreach(println)
 // CassandraRow{word: foo, count: 20}   
 ```
 
+### Using fake CassandraRDD implementation
+
+In some cases there may be a need to use an empty RDD which is an instance of `CassandraRDD` but it
+does not perform any validation for the existence of the provided table and columns and it does not
+even try to fetch the results from the database. Such an RDD can represent a table which does not
+exists. To create one, just initialize `CassandraRDD` as usual and then call `toEmptyCassandraRDD`
```
pkolaczk: does not exist  (no "s" after exist)

doc/2_loading.md
pkolaczk: Also, beware that calling any "where" or "select" or other functions returning a CassandraRDD on EmptyCassandraRDD must return an EmptyCassandraRDD.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/EmptyCassandraRDD.scala
```diff
@@ -0,0 +1,21 @@
...
+class EmptyCassandraRDD[R](@transient sc: SparkContext,
+                           connector: CassandraConnector,
+                           keyspaceName: String,
+                           tableName: String)
+                          (implicit ct: ClassTag[R], @transient rtf: RowReaderFactory[R])
+
+  extends CassandraRDD[R](sc, connector, keyspaceName, tableName) {
+
+  override def getPartitions: Array[Partition] = Array.empty
+
```
pkolaczk: What if someone calls `where`? It must not return a normal `CassandraRDD` then.

doc/2_loading.md
jacek-lewandowski: Good catch for these methods. I think I'll remove the `EmptyCassandraRDD` class and make a param which will denote whether the RDD is empty or not.

For the first comment - this empty RDD is supposed to be CassandraRDD, which includes such properties as keyspace and table name. 
I'll provide a method to create it directly but I think that the user still needs to provide keyspace and table names, even if they do not exists.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraRDD.scala
jacek-lewandowski: I decided to remove `EmptyCassandraRDD`.

However - `makeRDD` is a method of `SparkContext`, isn't it? If we decided to do such changes, we should rename `cassandraTable` to `makeCassandraRDD` and add `makeEmptyCassandraRDD`. I'm not convinced for this change in this ticket.

RussellSpitzer:
We may want to reword the loading doc, I put a sample clarification on https://github.com/datastax/spark-cassandra-connector/pull/558

+1 on the code

jacek-lewandowski:
Rebased

helena:
@jacek-lewandowski actually I just noticed the private[spark] EmptyRDD.. had not seen that before.

PR #559, opened 2015-02-19
fix to report the correct row count

chanderspechetty:
row count is reported as 0 currently. As the number of rows is logged at
info level, this can be misleading. This fix passes the CountingIterator
to the batch builder.


jacek-lewandowski:
+1

PR #560, opened 2015-02-19
SPARKC-61 bump to sbt assembly 0.12

redbaron:


helena:
@redbaron I checked out your branch and ran ;clean ;assembly and get many of these errors:

[error] java.io.FileNotFoundException: /Users/helena/.sbt/boot/scala-2.10.4/lib/scala-reflect.jar (Interrupted system call)
[error] java.io.FileNotFoundException: /Users/helena/.ivy2/cache/org.datanucleus/datanucleus-api-jdo/jars/datanucleus-api-jdo-3.2.6.jar (Interrupted system call)
etc

Which means there is something not right in your build changes. Perhaps you have something in your environment making your successful assembly runs work that others would not have.

helena:
In my experience with this particular assembly plugin error the fix is hard to find but simple.

pkolaczk:
Closing, because it is outdated. sbt assembly already works.

PR #561, opened 2015-02-19
Use hash ring as partition boundaries

redbaron:
This patch makes spark.cassandra.input.split.size=0 a special value
which makes partitioner to make split not using estimated number of rows
but instead use cassandra hashring as partition boundaries


RussellSpitzer:
We've now allowed a few other options for asking for discrete numbers of partitions as well as other options. Please open a new Jira if you believe we should reopen

PR #562, opened 2015-02-20
Upgrade to Scala 2.11 and Add cross build

helena:


helena:
For anyone wanting to test aside from our internal QA, https://datastax-oss.atlassian.net/browse/SPARKC-22

doc/12_building_and_artifacts.md
```diff
@@ -0,0 +1,85 @@
...
+The library package jars will be generated to:
+  - `spark-cassandra-connector/target/scala-{binary.version}/`
+  - `spark-cassandra-connector-java/target/scala-{binary.version}/`
+
+The documentation will be generated to:
+  - `spark-cassandra-connector/target/scala-{binary.version}/api/`
+  - `spark-cassandra-connector-java/target/scala-{binary.version}/api/`
+
+##### Build Tasks
+The easiest way to do this is to make the assembled connector jar using
```
RussellSpitzer: I would like to keep this in the quick use section just to show people how to get started with the Shell as quickly as possible but I definitely see the rational for moving it.

helena: I get it but we had duplicate build/artifact content, and over time I don't want to have us maintaining both. This way it is a link in both README and getting started documentation pages, where it is currently duplicated.

project/Settings.scala
```diff
@@ -138,12 +198,16 @@ object Settings extends Build {
     }
   )
 
-  /* By default, assembly is not enabled for the demos module, but it can be enabled with
-    `-Ddemos.assembly=true`. From the command line this would be:
-     sbt -Ddemos.assembly=true twitter/assembly */
-  lazy val demoAssemblySettings =
-    if (System.getProperty("demos.assembly", "true").toBoolean) sbtAssemblySettings
-    else Seq.empty
+  lazy val theme = PromptTheme(List(
```
RussellSpitzer: With style :)

spark-cassandra-connector/scala-2.11/src/main/scala/com/datastax/spark/connector/rdd/reader/AnyObjectFactory.scala
```diff
@@ -0,0 +1,194 @@
+package com.datastax.spark.connector.rdd.reader
```
RussellSpitzer: What is different in this class vs the 2.10 version?

spark-cassandra-connector/scala-2.11/src/main/scala/com/datastax/spark/connector/rdd/reader/AnyObjectFactory.scala
```diff
@@ -0,0 +1,194 @@
...
+  val argCount: Int = constructorParamTypes.length
+
+  val argOffset: Int = oneIfMemberClass(javaClass)
+
+  @transient
+  lazy val constructorParamTypes: Array[Type] = {
+    val requiredParamClasses = javaConstructor.getParameterTypes
+      .drop(AnyObjectFactory.oneIfMemberClass(javaClass))
+
+    tpe.decl(termNames.CONSTRUCTOR).asTerm.alternatives.map { term =>
```
helena: @RussellSpitzer this

spark-cassandra-connector/scala-2.11/src/main/scala/com/datastax/spark/connector/rdd/reader/ClassBasedRowReader.scala
```diff
@@ -0,0 +1,145 @@
...
+  private val tpe = implicitly[TypeTag[R]].tpe
+
+  // This must be  serialized:
+  val constructorArgConverters: Array[TypeConverter[_]] =
+    factory.constructorParamTypes.map(t => TypeConverter.forType(t))
+
+  @transient
+  private val setterTypes: Map[String, Type] = {
+    def argType(name: String) = {
+      val methodSymbol = tpe.decl(TermName(name)).asMethod
```
helena: @RussellSpitzer this is the other

project/Settings.scala
RussellSpitzer: This doesn't really respect default terminal colors so on my screen it's unreadable

image

helena: Hrm, did not even consider that obviously. I'll see if the plugin can handle that even.

RussellSpitzer:
Minor Other than my love of white text on black backgrounds this looks good,
+1 Code
+1 Unittests/It Tests

helena:
@RussellSpitzer fixed shell prompt for user defaults.

helena:
@pkolaczk Can you review this week? Thanks

spark-cassandra-connector/scala-2.11/src/main/scala/com/datastax/spark/connector/rdd/reader/AnyObjectFactory.scala
pkolaczk: I'd certainly like to *not* have to maintain this class separately for scala 2.10 and 2.11. Seems there is exactly *one* single line change in this file - the line to get the constructor declaration. Please extract the method for getting constructor into external object, that can be coded differently for 2.10 and 2.11.

spark-cassandra-connector/scala-2.11/src/main/scala/com/datastax/spark/connector/rdd/reader/AnyObjectFactory.scala
pkolaczk: Yes, please extract it to ReflectionUtil. 
The same applies to ClassBasedRowReader. There is no point in forking the whole class if the only thing that requires special treating is a well defined operation for getting the constructor definition.

spark-cassandra-connector/scala-2.11/src/main/scala/com/datastax/spark/connector/rdd/reader/ClassBasedRowReader.scala
```diff
@@ -0,0 +1,145 @@
+package com.datastax.spark.connector.rdd.reader
```
pkolaczk: As stated above, no need to have this separated.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraRelation.scala
```diff
@@ -9,7 +9,7 @@ import org.apache.spark.sql.catalyst.plans.logical.{Statistics, LeafNode}
 import org.apache.spark.sql.{StructField, catalyst}
 
 private[cassandra] case class CassandraRelation
-  (tableDef: TableDef, alias: Option[String])(@transient cc: CassandraSQLContext)
+  (tableDef: TableDef, alias: Option[String])(@transient val cc: CassandraSQLContext)
```
pkolaczk: Why need val here?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/GroupingBatchBuilder.scala
```diff
@@ -78,7 +78,7 @@ class GroupingBatchBuilder[T](batchStatementBuilder: BatchStatementBuilder[T],
 
   /** Creates a statement from the given batch and cleans the batch so that it can be reused. */
   @inline
-  private def createStmtAndReleaseBatch(batch: Batch): RichStatement = {
+  protected def createStmtAndReleaseBatch(batch: Batch): RichStatement = {
```
pkolaczk: Why protected?

spark-cassandra-connector/scala-2.11/src/main/scala/com/datastax/spark/connector/rdd/reader/AnyObjectFactory.scala
helena: Agreed. I brought up the idea of creating the primary trait and branch the changed functions in the 2 branches. I'll push that soon.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraRelation.scala
helena: Does not compile in 2.11

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/GroupingBatchBuilder.scala
helena: It broke in compile but I retested and it compiles in both now. Reverted.

pkolaczk:
There are two separate compiles, once with picking files from 2.10 and another time with 2.11; and the rest of files is common. Am I right?
So I can see no problem here. The code for AnyObjectFactory and ClassBasedRowReader needs to be in the *common part*. For reflection, you call some ReflectionUtil object which has a common API, but different implementetation (separated into 2.10 and 2.11).

helena:
@pkolaczk it can't be in the common part, the actual class names but all the code for the common (now abstractions) is in the shared source. The file names must be the same for the 2 builds, which is how I have it.

helena:
I think having a shared abstract class in shared src containing all the code for those 2 but the one line difference per scala version, and the one line implementation for 2.10 and 2.11 is clean vs adding a new reflection util. Adding *Util* to me smells like poor design.

pkolaczk:
Honestly, adding more inheritance smells like poor design more than adding an Util to me, particularly *implementation inheritance*. The main (only?) reason for an abstract class is when there are going to be 2 or more concrete implementations (or you at least predict there will be more than one in the future) to be switched at runtime, differing with something important. In this case, logically we've got only one, however non-portable. 

A widely applied and very successful pattern for dealing with lack of portability is extracting the non-portable bits (and only them) to external compatibility layer exposing a common API. This way you don't break SRP: the class is still responsible only for its core functionality, while the compatibility layer is responsible for portability and dealing with whatever differences there might be between Scala versions. Clear separation of concerns. And btw, if you don't like Utils, just call it `ScalaReflectionCompatibilityLayer` This way everybody knows what this class is for. Scala itself comes with plenty of singletons and *stateless* singletons are harmless.  :)

helena:
Are you saying you want an object vs class?
Note that for an object you can not lazy val this: implicitly[TypeTag[T]].tpe, it has to decorate a class but I can make it a function.

helena:
@pkolaczk updated PR.

pkolaczk:
+1, please rebase / squash the commits, as there were a few back-and-forth moves.

pkolaczk:
All tests green, both for 2.10 and 2.11.
sbt assembly ok for both as well.

helena:
LGTM?

helena:
@piotr will do. Thanks for pushing me to find a cleaner way on this.

helena:
Closing for the rebased/squashed version for another round due to issues with java api. See https://github.com/datastax/spark-cassandra-connector/pull/573

PR #566, opened 2015-02-27
Update 2_loading.md

twistedpair:
Misleading typo. Empty RDD's are for tables that don't exist, not tables that do exist.


RussellSpitzer:
Good catch!

PR #567, opened 2015-02-28
SPARKC-72 more spark sql predicate pushdown and tests

alexliu68:


spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLPredicatePushdownSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  it should "not allow to push down none-primary column IN predicate. " +
+    " e=3 is pushed down" in {
+    val result = cc.sql("select * from test_pd where e=3 and j in (1,2,3)").collect()
+  }
+
+  it should "not allow to push down none-primary column predicates if there are no indexed column predicates. " +
+    " a=1 and b=1 and c=1 and d=3 are pushed down" in {
+    val result = cc.sql("select * from test_pd where a=1 and b=1 and c=1 and d=3 and k=4").collect()
```
jacek-lewandowski: I know this would be a pain but... would you test the results for the correct content?

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLPredicatePushdownSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+  }
+}
```
jacek-lewandowski: One more thing - we have some metrics available which include the number of fetched rows. Do you think we may use them to really check whether the predicates are really pushed down?

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLPredicatePushdownSpec.scala
alexliu68: Most of them returns empty results, so I don't check the result. The  logging print out shows which predicates have been pushed down. Because it's in the middle of the execution, we can't compare it to the correct push down predicate.  We need manually check the logging printout to see whether the  push down is correct or not.

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLPredicatePushdownSpec.scala
alexliu68: Does the metrics returns the number of fetched rows from the CQL query under hood? If that's the case, we can use it to check something. 

But It doesn't help that much. check my comments above. Spark sql filters the result from C* with the pushed down predicates by remaining predicates. All we have to do is check the correct predicates have been pushed down. The remaining predicates are all predicates minus push down predicates. 

Only check number of fetched row doesn't guaranteed  the correct predicates have been pushed down.

jacek-lewandowski: Actually we measure the number of rows which were fetched from Cassandra, not the number of rows processed by Spark - so there shall be the less number of rows fetched if the predicate is pushed down, am I right?

jacek-lewandowski: Disregard what I wrote. I found another way to test it well, look at this:
When we expect the predicate to be pushed down, and no Spark-side filtering is used, we can do the following:
```scala
val plan = cc.sql("SELECT * FROM ks.tab WHERE id = 1").queryExecution.sparkPlan
plan.nodeName shouldBe "CassandraTableScan"
plan.children should have size 0
```

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraStrategies.scala
```diff
@@ -121,7 +121,8 @@ private[cassandra] trait CassandraStrategies {
         val eqPredicates = eqColumns.flatMap(eqPredicatesByName)
         val optionalNonEqPredicate = for {
           c <- otherColumns.headOption.toSeq
```
jacek-lewandowski: no need to convert it to seq

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraStrategies.scala
```diff
@@ -121,7 +121,8 @@ private[cassandra] trait CassandraStrategies {
...
-          p <- firstNonEmptySeq(rangePredicatesByName(c), inPredicatesByName(c).filter(_ => eqColumns.nonEmpty))
+          p <- firstNonEmptySeq(rangePredicatesByName(c), inPredicatesByName(c).filter(
+            _ => eqColumns.nonEmpty && c==clusteringColumns.last))
```
jacek-lewandowski: why `eqColumns.nonEmpty`? shouldn't be there something like `eqColumns.size == clusteringColumns.size - 1` ?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraStrategies.scala
```diff
@@ -138,7 +139,9 @@ private[cassandra] trait CassandraStrategies {
         val eqIndexedColumns = indexedColumns.filter(eqPredicatesByName.contains)
         val eqIndexedPredicates = eqIndexedColumns.flatMap(eqPredicatesByName)
         val nonIndexedPredicates = for {
-          c <- allColumns if !eqIndexedColumns.contains(c)
+          c <- allColumns if !partitionKeyPredicatesToPushDown.isEmpty && !eqIndexedColumns.contains(c) ||
```
jacek-lewandowski: these conditions are not clear to me - could you add some comments?

jacek-lewandowski:
Reviewed, could you take a look at my new comments @alexliu68 ?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraStrategies.scala
alexliu68: I remove eqColumns.nonEmpty  condition. We only need c==clusteringColumns.last condition, so that only if the last cluster column with IN clause can be pushed down.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraStrategies.scala
alexliu68: I remove seq, it doesn't compile. Any idea how to remove it?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraStrategies.scala
alexliu68: sure

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLPredicatePushdownSpec.scala
alexliu68: I'll add those easy check for tests without spark sql side filtering.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraStrategies.scala
jacek-lewandowski: Well - I'm not sure - I thought that you can use `in` on the clustering column if eq predicates are defined on the preceding clustering columns. I tried this in cqlsh and got the following result:
```
cqlsh:ks> select * from users2 where cid2 in (4, 5) allow filtering;

InvalidRequest: code=2200 [Invalid query] message="PRIMARY KEY column "cid2" cannot be restricted (preceding column "cid1" is either not restricted or by a non-EQ relation)"
```

alexliu68: c <- otherColumns.headOption.toSeq only returns the first other column, so all previous columns have Eq predicates if there eqColumns are not empty.

Only the last cluster column are allowed to have IN predicate. I will add some more tests.

alexliu68:
@jacek-lewandowski  update code as reviewing.

jacek-lewandowski:
+1

jacek-lewandowski:
Please update CHANGES files

PR #569, opened 2015-03-04
Sparkc-73 support cross cluster table join for Spark SQL Cassandra integration

alexliu68:
For integration tests, set 127.0.0.2 as a loopback alias address.

### DIFF
--- a/spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/EmbeddedCassandra.scala
+++ b/spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/EmbeddedCassandra.scala
@@ -9,6 +9,7 @@ import java.net.InetAddress
 trait EmbeddedCassandra {
 
   def cassandraHost = EmbeddedCassandra.cassandraHost
+  def secondCassandraHost = SecondEmbeddedCassandra.cassandraHost
 
   /** Implementation hook. */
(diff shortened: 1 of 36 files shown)


jacek-lewandowski:
You should also take a look into `RDDFunctions` - I think these methods should also get `cluster` parameter.

jacek-lewandowski:
Also either add support for this in Java API or create a separate ticket to make sure we won't forget about it before release.

jacek-lewandowski:
For SharedEmbeddedCassandra - I'd really like it to be a kind of flexible array rather than hardcoded two objects. And we should have a single configuration yaml with more configurable fields. Your change will greatly improve our ability to do integration tests. However, if you make it more generic, we will benefit even more.

jacek-lewandowski:
First round of review done

alexliu68:
Can we open another ticket to improve the testing infrastructure(It will be big enough for a separate ticket :))? Let the function implementation be done in this PR?

jacek-lewandowski:
So, at least please eliminate use of the second yaml file (add more placeholder to the one we use now) and please create the other tickets to cover those missing fixes.

jacek-lewandowski:
Also, please make sure to not run the tests which require another localhost address to be configured. Just show some warning in that case and ignore the test. 

After rebasing, there will be a script {{dev/run-tests.sh}}. You can add the configuration of additional loopback address in that file.

PR #572, opened 2015-03-04
SPARKC-70: Fixed reading and writing objects with inherited fields

jacek-lewandowski:
With this patch:
- Added test cases for CassandraRDD, TableWriter, CassandraJavaRDD and CassandraJavaUtil
- Updted test cases in ReflectionUtilSpec
- Fixed ClassBasedRowReader, DefaultRowWriter and ReflectionUtils so that they use member methods rather than declared methods


pkolaczk:
+1, looks good.

PR #576, opened 2015-03-11
document is on the same folder level

snackycracky:


RussellSpitzer:
Thanks for the fix! +1

PR #581, opened 2015-03-13
SPARKC-85: Fix compiler warnings related to incorrect pattern matching o...

pkolaczk:
...n ColumnRefs. Move column selection code to TableDef.


alexliu68:
+1

helena:
These are some of the compiler warnings I have on a related ticket that we need to fix. I think this may resolve that. Would be great.

PR #586, opened 2015-03-16
SPARKC-84: Rename batch.level to batch.grouping.key. Rename batch.buffer...

pkolaczk:
....size to batch.grouping.buffer.size. Rename batch grouping key value "all" to "none" and "replicaset" to "replica_set".

### DIFF
--- a/doc/5_saving.md
+++ b/doc/5_saving.md
@@ -120,9 +120,9 @@ The following properties set in `SparkConf` can be used to fine-tune the saving 
   - `spark.cassandra.output.batch.size.rows`: number of rows per single batch; default is 'auto' which means the connector 
      will adjust the number of rows based on the amount of data in each row  
   - `spark.cassandra.output.batch.size.bytes`: maximum total size of the batch in bytes; defaults to 16 kB.
-  - `spark.cassandra.output.batch.level`: determines how insert statements are grouped into batches; available values are:
-     - `all`: all statements are appended to a single batch until it reaches its maximum size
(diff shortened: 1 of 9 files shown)


spark-cassandra-connector-demos/simple-demos/src/main/scala/com/datastax/spark/connector/demo/BasicWriteDemo.scala
```diff
@@ -0,0 +1,21 @@
...
+    .set("spark.cassandra.connection.host", "localhost")
+    .set("spark.cassandra.output.concurrent.writes", "1")
+    .set("spark.cassandra.output.batch.level", "replica_set")
+    .setMaster("local[1]")
+
+  val sc = new SparkContext(conf)
+
+  val col = for (i <- 1 to 1000000) yield (i, "foo" + i)
+  val rdd = sc.parallelize(col, 20)
+  rdd.saveToCassandra("test", "kv")
```
RussellSpitzer: Might want to have saveAsCassandraTable here or create the table beforehand

jacek-lewandowski:
+1

spark-cassandra-connector-demos/simple-demos/src/main/scala/com/datastax/spark/connector/demo/BasicWriteDemo.scala
pkolaczk: Hmm, I actually never wanted to commit this file. But indeed, let's create a table beforehand.

PR #588, opened 2015-03-18
SPARKC-91: Fixes in metrics

jacek-lewandowski:
- enabling/disabling Codahale metrics via metric.properties file
- changed locking to LongAdder in output metrics (dozens times faster)
- removed one useless metric
- added several unit tests and some integration tests
- updated docs and changes

### DIFF
--- a/doc/11_metrics.md
+++ b/doc/11_metrics.md
@@ -19,18 +19,18 @@ driver. To access these metrics add a new source called cassandra-connector in y
 `metrics.properties` file. Example:
 
 ```
-cassandra-connector.sink.csv.class=org.apache.spark.metrics.sink.CsvSink
-cassandra-connector.sink.csv.period=5
-cassandra-connector.sink.csv.unit=seconds
-cassandra-connector.sink.csv.directory=/tmp/spark/sink
+executor.source.cassandra-connector.class=org.apache.spark.metrics.CassandraConnectorSource
+driver.source.cassandra-connector.class=org.apache.spark.metrics.CassandraConnectorSource
 ```
 
 ### Performance impact
 While there should be a minimal performance effect from collecting metrics, Metric collection can be
-disabled by setting the following options in the Spark configuration:
+disabled. Codahale metrics are not collected if CassandraConnectorSource is not specified in the
+metrics configuration file. In order to disable task metrics, use these properties in Spark
+configuration:
 
-- `spark.cassandra.input.metrics` - set to `false` to disable collection of input metrics
-- `spark.cassandra.output.metrics` - set to `false` to disable collection of output metrics
+- `spark.cassandra.input.metrics` - set to `false` to disable collection of input task metrics
+- `spark.cassandra.output.metrics` - set to `false` to disable collection of output task metrics
 
 ### Available metrics
 Metric name            | Unit description
@@ -44,7 +44,11 @@ write-success-counter  | Number successfully written batches
 write-failure-counter  | Number of failed batches
 read-byte-meter        | Number of bytes read from Cassandra
 read-row-meter         | Number of rows read from Cassandra
-read-page-wait-timer   | The Time spent by the driver waiting for rows to be paged in from C*
 read-task-timer        | Timer to measure time of reading a single partition
+
+### Compatibility
+Codahale based metrics should work with either Spark 1.1.x or Spark 1.2.x. However task metrics
+works only with Spark 1.2.x. Therefore, if this version of Spark Cassandra Connector is to be used with
+Spark 1.1.x, task based metrics have to be disabled.
 
 [Next - Building And Artifacts](doc/12_building_and_artifacts.md)
--- a/spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
+++ b/spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
@@ -1,24 +1,59 @@
 package com.datastax.spark.connector.embedded
 
-import java.net.InetAddress
-
-import org.apache.spark.{SparkEnv, SparkConf, SparkContext}
+import akka.actor.ActorSystem
+import org.apache.spark.{SparkConf, SparkContext, SparkEnv}
 
 trait SparkTemplate {
-  val conf = SparkTemplate.conf
+  /** Obtains the active [[org.apache.spark.SparkContext]] object. */
+  def sc: SparkContext = SparkTemplate.sc
+
+  /** Obtains the [[akka.actor.ActorSystem]] associated with the active [[org.apache.spark.SparkEnv]]. */
+  def actorSystem: ActorSystem = SparkTemplate.actorSystem
+
+  /** Ensures that the currently running [[org.apache.spark.SparkContext]] uses the provided configuration.
+    * If the configurations are different or force is `true` Spark context is stopped and started again
+    * with the given configuration. */
+  def useSparkConf(conf: SparkConf, force: Boolean = false): SparkContext =
+    SparkTemplate.useSparkConf(conf, force)
+
+  def defaultSparkConf = SparkTemplate.defaultConf
 }
 
 object SparkTemplate {
 
-  val conf = new SparkConf(true)
+  /** Default configuration for [[org.apache.spark.SparkContext]]. */
+  val defaultConf = new SparkConf(true)
     .set("spark.cassandra.connection.host", EmbeddedCassandra.getHost(0).getHostAddress)
     .set("spark.cleaner.ttl", "3600")
-    .set("spark.metrics.conf", this.getClass.getResource("/metrics.properties").getPath)
     .setMaster(sys.env.getOrElse("IT_TEST_SPARK_MASTER", "local[*]"))
-    .setAppName(getClass.getSimpleName)
+    .setAppName("Test")
 
-  System.err.println("The Spark configuration is as follows:\n" + conf.toDebugString)
+  private var _sc: SparkContext = _
 
-  lazy val actorSystem = SparkEnv.get.actorSystem
+  /** Ensures that the currently running [[org.apache.spark.SparkContext]] uses the provided configuration.
+    * If the configurations are different or force is `true` Spark context is stopped and started again
+    * with the given configuration. */
+  def useSparkConf(conf: SparkConf = SparkTemplate.defaultConf, force: Boolean = false): SparkContext = {
+    if (_sc.getConf.getAll.toMap != conf.getAll.toMap || force)
+      resetSparkContext(conf)
+    _sc
+  }
 
+  private def resetSparkContext(conf: SparkConf) = {
+    if (_sc != null) {
+      _sc.stop()
+    }
+
+    System.err.println("Starting SparkContext with the following configuration:\n" + defaultConf.toDebugString)
+    _sc = new SparkContext(conf)
+    _sc
+  }
+
+  /** Obtains the active [[org.apache.spark.SparkContext]] object. */
+  def sc: SparkContext = _sc
+
+  /** Obtains the [[akka.actor.ActorSystem]] associated with the active [[org.apache.spark.SparkEnv]]. */
+  def actorSystem: ActorSystem = SparkEnv.get.actorSystem
+
+  resetSparkContext(defaultConf)
 }
--- a/spark-cassandra-connector-java/src/it/scala/com/datastax/spark/connector/CassandraJavaUtilSpec.scala
+++ b/spark-cassandra-connector-java/src/it/scala/com/datastax/spark/connector/CassandraJavaUtilSpec.scala
@@ -12,6 +12,8 @@ import scala.collection.JavaConversions._
 class CassandraJavaUtilSpec extends SparkCassandraITFlatSpecBase with BeforeAndAfter {
 
   useCassandraConfig(Seq("cassandra-default.yaml.template"))
+  useSparkConf(defaultSparkConf)
+
   val conn = CassandraConnector(Set(EmbeddedCassandra.getHost(0)))
 
   before {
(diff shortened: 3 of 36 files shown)


doc/11_metrics.md
```diff
@@ -44,7 +44,11 @@ write-success-counter  | Number successfully written batches
 write-failure-counter  | Number of failed batches
 read-byte-meter        | Number of bytes read from Cassandra
 read-row-meter         | Number of rows read from Cassandra
-read-page-wait-timer   | The Time spent by the driver waiting for rows to be paged in from C*
 read-task-timer        | Timer to measure time of reading a single partition
 
+### Compatibility
+Codahale based metrics should work with either Spark 1.1.x or Spark 1.2.x. However task metrics
+works only with Spark 1.2.x. Therefore, if this version of Spark Cassandra Connector is supposed to
```
pkolaczk: is to be used with Spark 1.1.x

spark-cassandra-connector-java/src/main/java/com/datastax/spark/connector/japi/RDDAndDStreamCommonJavaFunctions.java
```diff
@@ -177,7 +177,7 @@ public WriterBuilder withBatchSize(BatchSize batchSize) {
                 return withWriteConf(
                     new WriteConf(batchSize, writeConf.batchGroupingBufferSize(), writeConf.batchGroupingKey(),
                         writeConf.consistencyLevel(), writeConf.parallelismLevel(), writeConf.throughputMiBPS(),
-                        writeConf.ttl(), writeConf.timestamp()));
+                        writeConf.ttl(), writeConf.timestamp(), writeConf.taskMetricsEnabled()));
```
pkolaczk: Not now, but for the future: I think this whole fragment is too repeatable and written in Scala-style which doesn't play nice with Java. I think we need to implement those `withXYZ` methods in a standard Java way - that is by setting a private variable and then eventually having a build() method that creates a final `WriteConf`.

jacek-lewandowski: Yeah, I'm gonna work on cumulative updates and fixed in Java API, so I can address this

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
```diff
@@ -4,7 +4,16 @@ import org.apache.spark.{SparkEnv, SparkConf, SparkContext}
 
 trait SparkTemplate {
   val conf = SparkTemplate.conf
-  val sc = SparkTemplate.sc
+  def sc = SparkTemplate.sc
+
+  def resetSparkContext(conf: SparkConf = conf): Unit = {
```
pkolaczk: If there is at most one SparkContext, then this method really belongs to SparkTemplate *object*.

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
```diff
@@ -18,8 +27,8 @@ object SparkTemplate {
 
   System.err.println("The Spark configuration is as follows:\n" + conf.toDebugString)
 
-  val sc = new SparkContext(conf)
+  var sc = new SparkContext(conf)
```
pkolaczk: I think this should be private and modifiable only from resetSparkContext

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
jacek-lewandowski: It is easier to use it here because you do not have to import companion object.

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
jacek-lewandowski: ok

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
pkolaczk: So just import companion object. Methods modifying vars should be close to vars. If var is declared on the companion object, all methods manipulating that state should be on the same object.

pkolaczk:
Code style: in multiple places, lines are too long. We should try to stay within 100-105 chars (GitHub width), unless we have a really valid reason to exceed that.

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
jacek-lewandowski: I know that rules. However what do we need SparkTemplate for then?

spark-cassandra-connector/src/main/scala/org/apache/spark/metrics/CassandraConnectorSource.scala
```diff
@@ -20,8 +20,25 @@ object CassandraConnectorSource extends Source {
 
   val readByteMeter = metricRegistry.meter("read-byte-meter")
   val readRowMeter = metricRegistry.meter("read-row-meter")
-  val readPageWaitTimer = metricRegistry.timer("read-page-wait-timer")
   val readTaskTimer = metricRegistry.timer("read-task-timer")
 
-  lazy val ensureInitialized = SparkEnv.get.metricsSystem.registerSource(this)
+  CassandraConnectorSource._instance = Some(this)
+  CassandraConnectorSource._env = SparkEnv.get
```
pkolaczk: This is something potentially very dangerous. There can be multiple `CassandraConnectorSource` instances fighting for the shared "_instance" mutable field, overwriting it on their own instantiation. I believe you assume there is going to be only one, but the code doesn't force it in any way.

pkolaczk:
I haven't finished reviewing yet, but this ticket certainly needs better documentation at the class level. It is hard to analyze the dependency graph, as often mutable shared state is involved and dependencies are hidden. You need to document role of each class - what is its responsibility and why it is there.

spark-cassandra-connector/src/main/scala/org/apache/spark/metrics/CassandraConnectorSource.scala
jacek-lewandowski: This class is instantiated by `SparkEnv` so there can be only one active instance, just like `SparkEnv`. I will make the constructor of this class package private for Spark metrics package. This should be enough. However the other thing I can think about is to make sure that when then new instance is created, either there is no instance associated with `_instance` or the current env is different than `_env`.

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
```diff
@@ -1,25 +1,56 @@
...
+    if (_sc != null) {
+      _sc.stop()
+    }
+
+    System.err.println("Starting SparkContext with the following configuration:\n" + defaultConf.toDebugString)
+    _sc = new SparkContext(conf)
+    _sc
+  }
+
+  private var _sc: SparkContext = resetSparkContext(defaultConf)
```
pkolaczk: Fields should be declared before methods using them. This is to avoid accidental referencing a field before it is initialized.

spark-cassandra-connector-java/src/it/scala/com/datastax/spark/connector/CassandraJavaUtilSpec.scala
```diff
@@ -13,6 +13,8 @@ import scala.collection.JavaConversions._
 class CassandraJavaUtilSpec extends FlatSpec with Matchers with BeforeAndAfter with SharedEmbeddedCassandra with SparkTemplate {
 
   useCassandraConfig("cassandra-default.yaml.template")
+  useSparkConf()
```
pkolaczk: I think it would be much more readable to explicitly state which SparkConf to use. If the only purpose of this method is to make sure the default SparkConf is used, `useSparkConf(SparkTemplate.default)` would communicate this intent much more clearly.

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
```diff
@@ -1,25 +1,56 @@
...
+  val defaultConf = new SparkConf(true)
+                    .set("spark.cassandra.connection.host", EmbeddedCassandra.cassandraHost.getHostAddress)
+                    .set("spark.cleaner.ttl", "3600")
+                    .setMaster(sys.env.getOrElse("IT_TEST_SPARK_MASTER", "local[*]"))
+                    .setAppName("Test")
+
+  /** Ensures that the currently running [[SparkContext]] uses the provided configuration. If the
+    * configurations are different or force is `true` Spark context is stopped and started again with
+    * the given configuration. */
+  def useSparkConf(conf: SparkConf = SparkTemplate.defaultConf, force: Boolean = false): SparkContext = {
```
pkolaczk: Conf shouldn't have a default value here. If you insist on not giving the default argument explicitly, a special `useDefaultSparkConf` would be better.

spark-cassandra-connector/src/main/scala/org/apache/spark/metrics/CassandraConnectorSource.scala
```diff
@@ -4,24 +4,68 @@ import com.codahale.metrics
...
+  */
+class CassandraConnectorSource extends Source {
   override val sourceName = "cassandra-connector"
 
   override val metricRegistry = new metrics.MetricRegistry
 
-  val writeByteMeter = metricRegistry.meter("write-byte-meter")
-  val writeRowMeter = metricRegistry.meter("write-row-meter")
-  val writeBatchTimer = metricRegistry.timer("write-batch-timer")
+  val writeByteMeter      = metricRegistry.meter("write-byte-meter")
```
pkolaczk: Don't align like this. I know it looks more readable *now*, but on the other hand whenever you change a name of a field, the alignment gets broken. And you might not notice it when using automatic refactoring. And if do notice, and you fix alignment, you're only making probability of conflicts higher and adding a diff noise.

spark-cassandra-connector/src/test/scala/com/datastax/driver/core/RowMock.scala
```diff
@@ -0,0 +1,114 @@
+package com.datastax.driver.core
```
pkolaczk: Can't we use ScalaMock for this? It is already on the classpath. In this case reating a mock with just a few methods would be much simpler with ScalaMock.

spark-cassandra-connector/src/main/scala/org/apache/spark/metrics/CassandraConnectorSource.scala
```diff
@@ -4,24 +4,68 @@ import com.codahale.metrics
...
 import org.apache.spark.metrics.source.Source
 
-object CassandraConnectorSource extends Source {
+/** This class is a Source implementation for Cassandra Connector related Codahale metrics. It is
+  * automatically instantiated and registered by Spark metrics system if this source is specified in
+  * metrics configuration file.
+  *
+  * Spark instantiates this class when `SparkEnv` is started. There can be only a single instance of
+  * `SparkEnv` so there can be at most a single active instance of `CassandraConnectorSource`.
+  * The active instance is assigned to [[CassandraConnectorSource._instance]] so that it can be retrieved
```
pkolaczk: link not found by ScalaDoc

spark-cassandra-connector/src/main/scala/org/apache/spark/metrics/CassandraConnectorSource.scala
```diff
@@ -4,24 +4,68 @@ import com.codahale.metrics
...
+}
+
+object CassandraConnectorSource {
+  @volatile
+  private var _instance: Option[CassandraConnectorSource] = None
+  @volatile
+  private var _env     : SparkEnv                         = null
+
+  /** Returns an active instance of [[CassandraConnectorSource]] if it has been associated with the
+    * current [[SparkEnv]] and the current [[SparkEnv]] is running.
```
pkolaczk: Links to SparkEnv not found here.

pkolaczk:
Many links in ScalaDoc are not found because of a limitation of ScalaDoc: https://issues.scala-lang.org/browse/SI-8734

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/metrics/InputMetricsUpdater.scala
```diff
@@ -1,80 +1,149 @@
...
-    CassandraConnectorSource.readRowMeter.mark(cnt)
-    CassandraConnectorSource.readByteMeter.mark(metrics.bytesRead - dataLength)
-    dataLength = metrics.bytesRead
-    cnt = 0
-  }
+    override def updateMetrics(row: Row): Row = {
+      var rowLength = 0
+      for (i <- 0 until row.getColumnDefinitions.size() if !row.isNull(i))
+        rowLength += row.getBytesUnsafe(i).remaining()
+
```
pkolaczk: This makes a good function: `def rowLength(row: Row): Long`

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/metrics/InputMetricsUpdater.scala
```diff
@@ -1,80 +1,149 @@
...
-    for (i <- 0 until row.getColumnDefinitions.size() if !row.isNull(i))
-      metrics.bytesRead += row.getBytesUnsafe(i).remaining()
+    require(groupSize > 0)
 
-    cnt += 1
-    if (cnt == groupSize)
-      update()
-    row
-  }
+    private var cnt        = 0
```
pkolaczk: alignment

pkolaczk:
+1
Overall looks good. You can fix the things I found and move to testing / merging.

spark-cassandra-connector/src/test/scala/com/datastax/driver/core/RowMock.scala
jacek-lewandowski: I don't think it would be much simpler. I think it would be opposite because I would need to mock `getBytesUnsafe` and `isNull` in every case. However, I added inheritance from `AbstractGettableData` so that there are much less methods.

pkolaczk: Mocks can be shared.

PR #592, opened 2015-03-19
SPARKC-96: Allow RateLimiter to use an alternative time source and sleep...

pkolaczk:
... method. Make the test independent from the wall-clock time nd JVM performance hiccups.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RateLimiter.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RateLimiter.scala
@@ -15,8 +15,18 @@ import scala.annotation.tailrec
   * amount of time proportional to the amount of the overflow.
   *
   * This class is thread safe and lockless.
+  *
+  * @param rate maximum allowed long-term rate per 1000 units of time
+  * @param bucketSize maximum acceptable "burst"
+  * @param time source of time; typically 1 unit = 1 ms
+  * @param sleep a function to call to slow down the calling thread;
+  *              must use the same time units as `time`
   */
-class RateLimiter(rate: Long, bucketSize: Long) {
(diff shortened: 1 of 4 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RateLimiter.scala
```diff
@@ -16,7 +16,11 @@ import scala.annotation.tailrec
   *
   * This class is thread safe and lockless.
   */
-class RateLimiter(rate: Long, bucketSize: Long) {
+class RateLimiter(
+    rate: Long,
+    bucketSize: Long,
+    time: () => Long = System.currentTimeMillis,
+    sleep: Long => Any = Thread.sleep) {
```
pedjak: I would document the semantic of time() result and sleep() arg

spark-cassandra-connector/src/test/scala/com/datastax/spark/connector/writer/RateLimiterSpec.scala
```diff
@@ -4,44 +4,36 @@ import org.scalatest.{FlatSpec, Matchers}
...
-    val referenceElapsed = end1 - start1
-
-    val start2 = System.currentTimeMillis()
-    for (i <- 1 to 1000000)
-      limiter.maybeSleep(1000)
-    val end2 = System.currentTimeMillis()
-    val testElapsed = end2 - start2
-    testElapsed should be <= referenceElapsed + allowedSystemInducedPause
+    var now: Long = 0
+    def sleep(delay: Long) = throw new Exception("Sleep must never be called in this test")
```
pedjak: why not using ScalaMocks here instead? It is cleaner to me - you can check than assert that no interaction happened.

PR #598, opened 2015-03-24
SPARKC-99: Code formatting and cleanup.

pkolaczk:
Limit line length to 105 columns. No logic changed.


RussellSpitzer:
+1 I do like this style, The "self" type stuff is also very nice

PR #599, opened 2015-03-24
SPARKC-102 Add more valid props to config check

RussellSpitzer:


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigCheck.scala
```diff
@@ -22,8 +24,11 @@ object ConfigCheck {
     WriteConf.Properties ++
     ReadConf.Properties ++
     CassandraConnectorConf.Properties ++
+    AuthConf.Properties ++
```
pkolaczk: This wouldn't compile, there is no Properties on AuthConf object any more.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigCheck.scala
```diff
@@ -22,8 +24,11 @@ object ConfigCheck {
...
+    InputMetricsUpdater.Properties ++
+    OutputMetricsUpdater.Properties ++
     CassandraConnectionFactory.Properties ++
-    AuthConfFactory.Properties
```
pkolaczk: You shouldn't remove this

RussellSpitzer:
+1

PR #603, opened 2015-03-26
SPARKC-105: Fix setting custom connection factory which was blocked by t...

pkolaczk:
...he new configuration check system. Provided a way to set custom auth config and connection factory properties.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/AuthConf.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/AuthConf.scala
@@ -35,7 +35,23 @@ case class PasswordAuthConf(user: String, password: String) extends AuthConf {
 
 /** Obtains authentication configuration by reading  `SparkConf` object. */
 trait AuthConfFactory {
+
   def authConf(conf: SparkConf): AuthConf
+
+  /** List of extra properties allowed in SparkConf passed to `authConf` method */
+  def properties: Set[String] = Set.empty
+}
+
+object AuthConfFactory {
+  val AuthConfFactoryProperty = "spark.cassandra.auth.conf.factory"
+  val Properties = Set(AuthConfFactoryProperty)
+
+  def fromSparkConf(conf: SparkConf): AuthConfFactory = {
+    conf
+      .getOption(AuthConfFactoryProperty)
+      .map(ReflectionUtil.findGlobalObject[AuthConfFactory])
+      .getOrElse(DefaultAuthConfFactory)
+  }
 }
 
 /** Default `AuthConfFactory` that supports no authentication or password authentication.
@@ -61,19 +77,9 @@ object DefaultAuthConfFactory extends AuthConfFactory {
 /** Entry point for obtaining `AuthConf` object from `SparkConf`, used when establishing connections to Cassandra.
   * The actual `AuthConf` creation is delegated to the [[AuthConfFactory]] pointed by `spark.cassandra.auth.conf.factory` property. */
 object AuthConf {
-  val AuthConfFactoryProperty = "spark.cassandra.auth.conf.factory"
-
-  val Properties = Seq(
-    AuthConfFactoryProperty
-  )
 
   def fromSparkConf(conf: SparkConf) = {
-    val authConfFactory = conf
-      .getOption(AuthConfFactoryProperty)
-      .map(ReflectionUtil.findGlobalObject[AuthConfFactory])
-      .getOrElse(DefaultAuthConfFactory)
-
-    authConfFactory.authConf(conf)
+    AuthConfFactory.fromSparkConf(conf).authConf(conf)
   }
 }
 
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectionFactory.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectionFactory.scala
@@ -25,6 +25,8 @@ trait CassandraConnectionFactory extends Serializable {
   /** Creates and configures native Cassandra connection */
   def createCluster(conf: CassandraConnectorConf): Cluster
 
+  /** List of allowed custom property names passed in SparkConf */
+  def properties: Set[String] = Set.empty
 }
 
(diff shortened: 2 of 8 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/AuthConf.scala
```diff
@@ -35,7 +35,23 @@ case class PasswordAuthConf(user: String, password: String) extends AuthConf {
...
+
   def authConf(conf: SparkConf): AuthConf
+
+  /** List of extra properties allowed in SparkConf passed to `authConf` method */
+  def properties: Set[String] = Set.empty
+}
+
+object AuthConfFactory {
+  val AuthConfFactoryProperty = "spark.cassandra.auth.conf.factory"
+  val Properties = Set(AuthConfFactoryProperty)
```
helena: Thank you for one-lining this. 
Shouldn't it be a final val, final val are capitalized

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectionFactory.scala
```diff
@@ -81,6 +83,7 @@ object DefaultConnectionFactory extends CassandraConnectionFactory {
   * used when establishing connections to Cassandra. */
 object CassandraConnectionFactory {
   val ConnectionFactoryProperty = "spark.cassandra.connection.factory"
+  val Properties = Set(ConnectionFactoryProperty)
```
helena: final val ... for both?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigCheck.scala
```diff
@@ -29,7 +32,11 @@ object ConfigCheck {
    * @param conf SparkConf object to check
    */
   def checkConfig(conf: SparkConf): Unit = {
-    val unknownProps = getUnknownProperties(conf)
+    val connectionFactory = CassandraConnectionFactory.fromSparkConf(conf)
```
helena: These should be an apply but i know you disagree.
Apply's are for creating, the signature of the applys is all one needs, not "fromSparkConf"

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigCheck.scala
```diff
@@ -40,7 +47,8 @@ object ConfigCheck {
     }
   }
 
-  def getUnknownProperties(conf: SparkConf): Seq[String] = {
+  def getUnknownProperties(conf: SparkConf, extraProps: Set[String] = Set.empty): Seq[String] = {
```
helena: who added a 'get*', this is not java

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigCheck.scala
```diff
@@ -56,9 +64,10 @@ object ConfigCheck {
    *
    * Fuzziness is determined by MatchThreshold
    */
-  def getSuggestedVars(unknownProp: String): Option[Seq[String]] = {
+  def getSuggestedVars(unknownProp: String, extraProps: Set[String] = Set.empty): Option[Seq[String]] = {
```
helena: I know you didn't write this, i'm just seeing it for the first time.
It is odd to return an Option of some collection when it should simply be Seq.empty in this case

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/AuthConf.scala
pkolaczk: What is the point of final val for an object? Objects are final by definition.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigCheck.scala
pkolaczk: Ok, we can address it in another ticket. On the other hand, multiple apply() overload with *default* parameters is not allowed, while naming them differently allows default params.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigCheck.scala
pkolaczk: Oh yes, it caught my eye during review, too, however it turned out there was an *intended* semantic difference. Now when I look at it for the second time, I think it *can* be simplified to Seq[String], it just needs some minor modification of the call-site.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigCheck.scala
pkolaczk: *nod*

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/AuthConf.scala
helena: true, i am thinking in terms of a standard Akka class of settings which use final val MyVal syntax vs an object

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigCheck.scala
pkolaczk: BTW: In defense of Javaism, I can say that having "get" in the beginning makes it somehow handy when naming things on the call site. Compare those two snippets:
```scala
def getUnknownProperties(...) = ???
// later in the code:
val unknownProperties = getUnknownProperties(...)  // obvious
```
```scala
def unknownProperties(...) = ???
val unknownProperties = unknownProperties(...)  // damn; now it requires either explicit type annotation, or a different name
```

helena: Just a naming change, I'd hate to see java beans-isms sneak in

helena:
LGTM

PR #606, opened 2015-03-27
SPARKC-107 Synchronize Reflection Utils

pkolaczk:


spark-cassandra-connector/src/test/scala/com/datastax/spark/connector/util/ReflectionUtilSpec.scala
```diff
@@ -11,6 +14,25 @@ class ReflectionUtilSpec extends FlatSpec with Matchers {
     factory should be(DefaultConnectionFactory)
   }
 
+  it should "be able to find DefaultConnectionFactory in a multi-threaded context" in {
+    val conductor = new Conductor
+    import conductor._
+
+    thread("ThreadA") {
```
RussellSpitzer: For whatever reason this doesn't fail for me if the ^ above test runs first which is why i had it using the custom class

pkolaczk: Ok, so we need to pick a different class, because the one  you used was not present in 1.1.

PR #608, opened 2015-03-27
SPARKC-93:  Remove use of Thrift describe_ring.

pkolaczk:
Remove describe_ring call and replace it with native Java Driver support for fetching token ranges with corresponding replica sets from the Cluster metadata. Removed CassandraNode class, because now an endpoint is represented by only one address, taken from the Java driver Host object.  This is the first step towards removing thrift dependency from the connector code.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ServerSideTokenRangeSplitter.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ServerSideTokenRangeSplitter.scala
@@ -7,10 +7,10 @@ import scala.collection.JavaConversions._
 import scala.util.{Failure, Success, Try}
 
 import org.apache.cassandra.thrift.CfSplit
+import org.apache.spark.Logging
 
 import com.datastax.spark.connector.cql.CassandraConnector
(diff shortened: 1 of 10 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ServerSideTokenRangeSplitter.scala
```diff
@@ -35,24 +39,27 @@ class ServerSideTokenRangeSplitter[V, T <: Token[V]](
         client.set_keyspace(keyspaceName)
         client
           .describe_splits_ex(tableName, startToken, endToken, splitSize.toInt)
-          .map(unthriftify(_, range.endpoints))
```
RussellSpitzer: Why are we leaving the thrift method here?

pkolaczk: Because this is going to be addressed by another ticket. And it requires upgrading to a new Cassandra.

pkolaczk: SPARKC-94

RussellSpitzer:
Sgtm +1

PR #610, opened 2015-03-30
SPARKC-108: Fix DefaultAuthConfFactory properties not overriden properly...

pkolaczk:
... and therefore ignored.


RussellSpitzer:
I was thinking that the Properites element should be a Properties and a final constant that is overidden, but I don't have strong feelings about this. So i'm +1 for this fix as well.

tymoteuszsielach:
+1, it blocks my upgrade to spark 1.2

PR #611, opened 2015-03-30
SPARKC-109: Fix column aliases being ignored when fetching RDDs of Cassa...

pkolaczk:
...ndraRow.


jacek-lewandowski:
LGTM

BTW I wonder whether we should address native aliasing in CQL in a separate ticket. WDYT?

jacek-lewandowski:
All tests passed on my local machine

PR #612, opened 2015-03-30
SPARKC-111: Fixed bug in CassandraConnectorSource which caused problems when during initialization in executor environment

jacek-lewandowski:


spark-cassandra-connector/src/main/scala/org/apache/spark/metrics/CassandraConnectorSource.scala
```diff
@@ -37,10 +38,7 @@ class CassandraConnectorSource extends Source {
 }
 
 object CassandraConnectorSource {
-  @volatile
-  private var _instance: Option[CassandraConnectorSource] = None
-  @volatile
-  private var _env: SparkEnv = null
+  private var _state: (Option[CassandraConnectorSource], SparkEnv) = (None, null)
```
pkolaczk: Why None and null in the same place?

spark-cassandra-connector/src/main/scala/org/apache/spark/metrics/CassandraConnectorSource.scala
```diff
@@ -58,16 +58,41 @@ object CassandraConnectorSource {
 
   /** This method sets the given instance of [[CassandraConnectorSource]] as the current active instance.
     * It is allowed to call this method only once for the same `SparkEnv` instance which is currently
-    * up and running.
+    * up and running or this is the executor environment.
     */
   private def maybeSetInstance(ccs: CassandraConnectorSource): Unit = synchronized {
     val curEnv = SparkEnv.get
+    val (_instance, _env) = _state
     assert(ccs != null, "ccs cannot be null")
```
pkolaczk: Use require for precondition checking.

pkolaczk:
Generally +1, however I don't like those singletons in Spark everywhere (SparkEnv and stuff). Smells bad design to me.

spark-cassandra-connector/src/main/scala/org/apache/spark/metrics/CassandraConnectorSource.scala
pkolaczk: You can use `Option.apply` to convert nulls / non-nulls to `None` / `Some` automatically.

spark-cassandra-connector/src/main/scala/org/apache/spark/metrics/CassandraConnectorSource.scala
jacek-lewandowski: I really meant assertion. This is not intended to be used internally only and valid use of it should never cause those cases to occur.

spark-cassandra-connector/src/main/scala/org/apache/spark/metrics/CassandraConnectorSource.scala
jacek-lewandowski: ok

jacek-lewandowski:
I don't like it either. However I have 0 ideas how to do this other way without modifying Spark.

PR #614, opened 2015-04-01
SPARKC-110: Use a different object/class in ReflectionUtilSpec multithre...

pkolaczk:
...aded test, to avoid using cached results from the Scala reflection subsystem and increase the likelihood of triggering race conditions.


RussellSpitzer:
+1, Test fails if findSingleton is unsynchronized and passes if it is.

PR #615, opened 2015-04-01
SPARKC-115: Use JSR166e implementation from Twitter instead of accessing package private class in Guava

jacek-lewandowski:


pkolaczk:
+1
And I love the fact there is even less code now!

pkallos:
:+1: totally

PR #616, opened 2015-04-01
SPARKC-106: Removed conversion method rom WriteOption which accepted object of Duration type from Spark Streaming

jacek-lewandowski:


helena:
LGTM

PR #620, opened 2015-04-02
SPARKC-98 upgrade to spark 1.3.0

alexliu68:


artem-aliev:
+1

rbraley:
will this be pushed to maven central as well?

alexliu68:
We will not publish it until we have 1.30 release candidate.

PR #621, opened 2015-04-03
Sparkc 94 - Remove Thrift dependency - replace use of describe_splits_ex with CQL access

pkolaczk:


jacek-lewandowski:
@pkolaczk I haven't finished the review yet, but one thing I'd suggest is to revert the name change from `nativePort` to `port`. This value is still the native port and it is named `native` in `cassandra.yaml` file. Other than that, I thing this is unnecessary change in the set of our configuration properties. Would you consider leaving `nativePort`?

pkolaczk:
It is named native in cassandra.yaml, because Cassandra supports both thrift (rpc) and native interface. However, it is named just "port" in the official DataStax drivers (see `ClusterBuilder` API). I left "native" in the docs.

sdelmas:
retest this please

sdelmas:
Piotr, I hope it's ok to use this one to test our PR integration real quick.

sdelmas:
retest this please

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableRowReaderProvider.scala
```diff
@@ -29,13 +29,13 @@ trait CassandraTableRowReaderProvider[R] {
 
   protected def readConf: ReadConf
 
-  protected def fetchSize: Int = readConf.fetchSize
+  protected def fetchSize: Int = readConf.fetchSizeInRows
```
jacek-lewandowski: We are using size in bytes, megabytes, rows in more places (like `BatchSize`). Why don't create something like trait `Size` with subclasses `SizeInBytes` and `SizeInRows`, and then use these types where appropriate?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/Murmur3PartitionerTokenRangeSplitter.scala
```diff
@@ -2,29 +2,33 @@ package com.datastax.spark.connector.rdd.partitioner
...
-      (for (i <- 0 until n) yield left + (rangeSize * i.toDouble / n).toLong) :+ right
+      (for (i <- 0 until n) yield left + (rangeTokenCount * i / n).toLong) :+ right
+
     for (Seq(l, r) <- splitPoints.sliding(2).toSeq) yield
       new TokenRange[Long, LongToken](
         new LongToken(l),
         new LongToken(r),
         range.replicas,
-        Some((estimatedRows / n).toInt))
+        Some((rangeSize / n).toInt))
```
jacek-lewandowski: So, is the `rowCount` field name in `TokenRange` still valid?

jacek-lewandowski:
I have a similar concern as @artem-aliev does - what about older Cassandra versions? If I understand it correctly, it won't split the token ranges at all in that case, right? If so, it may lead to OOMs for large datasets. Maybe we could consider one of the following ideas:
- add an optional parameter to the read configuration, for example `sparkPartitionsPerTokenRange`, which would allow to explicitly specify how to split each token range
- create a pluggable mechanism for Thrift - I mean, we could provide the old functionality as a separate module and therefore not include Thrift dependency to the core

besides, LGTM - please consider my comments if you like (one more thing, which I couldn't comment in the code is that `CassandraRDDPartitioner.TokenRangeSampleSize` seems to be unused now).

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableRowReaderProvider.scala
pkolaczk: That's a good idea but even a better idea would be to encode the unit in the type. And even "more better" idea is to not code it by ourselves, but use something already invented for this purpose e.g. squants http://www.squants.com/. However, we should use it in all places and not just this one, so let's do it in a separate ticket.

jacek-lewandowski: sgtm

pkolaczk:
Yes, I'm working on an optional parameter to set the number of partitions explicitly. It will be useful also if someone doesn't like the automatic split-size tuning or for some other reasons it is not accurate.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/Murmur3PartitionerTokenRangeSplitter.scala
pkolaczk: Indeed, it should be named differently.

jacek-lewandowski:
retest this please

jacek-lewandowski:
retest this please

jacek-lewandowski:
retest this please

pkolaczk:
retest this please

pkolaczk:
retest this please

pkolaczk:
retest this please

pkolaczk:
```
		::::::::::::::::::::::::::::::::::::::::::::::
		::          UNRESOLVED DEPENDENCIES         ::
		::::::::::::::::::::::::::::::::::::::::::::::
		:: org.scala-sbt#sbt;0.13.8: not found
```

WTF?

pkolaczk:
@bcantoni @helena can you look why it has problems downloading sbt?

pkolaczk:
retest this please

pkolaczk:
Rebased onto master.

pkolaczk:
Rebased onto master.

pkolaczk:
retest this please

pkolaczk:
@jacek-lewandowski can you please review the last two commits? I added setting fixed split count for those who are running on older Cassandra and I added fetching listen addresses from the peers table to improve data locality in cloud environments.

jacek-lewandowski:
+1

PR #623, opened 2015-04-08
Sparkc 76 - Fix compilation warnings

pkolaczk:


spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/EmbeddedCassandra.scala
```diff
@@ -43,60 +46,81 @@ trait EmbeddedCassandra {
...
+    def checkValueType(obj: Any): ValueType
+  }
+
+  trait IntProperty extends TypedProperty {
+    type ValueType = Int
+    def convertValueFromString(str: String) = str.toInt
+    def checkValueType(obj: Any) =
+      obj match {
+        case x: Int => x
+        case _ => throw new ClassCastException (s"Expected Int, got ${obj.getClass.getName}")
```
helena: Pet peeve: 'got' usage in error or log messages, it's poor grammar. How about this for the message format

s"Expected Int but found ${obj.getClass.getName}"
or
s"Expected: Int, actual: ${obj.getClass.getName}"

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/EmbeddedCassandra.scala
```diff
@@ -43,60 +46,81 @@ trait EmbeddedCassandra {
...
+  val nativePorts = getValueSeq(NativePortProperty)
+  val rpcPorts = getValueSeq(RpcPortProperty)
+
+  def getProperty(nodeProperty: NodeProperty): Option[String] =
+    sys.env.get(nodeProperty.propertyName)
+  
+  def getPropertyOrThrowIfNotFound(nodeProperty: NodeProperty): String =
+    getProperty(nodeProperty).getOrElse(
+      throw new ConfigurationException(s"Missing ${nodeProperty.propertyName} in system environment"))
 }
```
helena: I am not sure about this, a pattern where an env or system setting missing will throw an exception.
Settings should be options with sensible defaults if not found so that deploys can move through a chain of cluster environments like dev/test/staging etc with no code changes.

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/EmbeddedCassandra.scala
```diff
@@ -43,60 +46,81 @@ trait EmbeddedCassandra {
...
+      }
+  }
+
+  trait InetAddressProperty extends TypedProperty {
+    type ValueType = InetAddress
+    def convertValueFromString(str: String) = InetAddress.getByName(str)
+    def checkValueType(obj: Any) =
+      obj match {
+        case x: InetAddress => x
+        case _ => throw new ClassCastException (s"Expected InetAddress, got ${obj.getClass.getName}")
```
helena: validation should be in the creation, i.e. an apply, to fail fast

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/EmbeddedCassandra.scala
```diff
@@ -120,37 +144,31 @@ object EmbeddedCassandra {
...
+        case HostProperty if index < hosts.size => hosts(index)
+        case _ => throw new RuntimeException(s"$index index is overflow the size of ${hosts.size}")
+      }
     }
-  }
+  }   
 
   Runtime.getRuntime.addShutdownHook(new Thread(new Runnable {
-    override def run() = cassandraRunners.map(_.map(_.destroy()))
+    override def run() = cassandraRunners.foreach(_.foreach(_.destroy()))
```
helena: not being in the code itself, it looks like it should be done with a .flatten

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ByteBufferUtil.scala
```diff
@@ -0,0 +1,28 @@
...
+      System.arraycopy(src.array, offset, dest, 0, length)
+    } else {
+      src.duplicate.get(dest)
+    }
+    dest
+  }
+
+  /** Converts a byte buffer into an array.
+    * The buffer's position remains untouched. */
+  def getArray(buffer: ByteBuffer): Array[Byte] = {
```
helena: def toArray vs get because this is converting

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ByteBufferUtil.scala
```diff
@@ -0,0 +1,28 @@
...
+  def copyBuffer(src: ByteBuffer, dest: Array[Byte]): Array[Byte] = {
+    if (src.hasArray) {
+      val length = src.remaining
+      val offset =  src.arrayOffset + src.position
+      System.arraycopy(src.array, offset, dest, 0, length)
+    } else {
+      src.duplicate.get(dest)
+    }
+    dest
+  }
```
helena: Or maybe

```def toBytes(src: ByteBuffer, dest: Array[Byte]): Array[Byte] = {
    if (src.hasArray)  {
      val length = src.remaining
      src.arrayOffset + src.position match {
        case boff  if boff == 0 && length == src.array.length => src.array
        case boff => java.util.Arrays.copyOfRange(src.array, boff, boff + length)
  }} else  
      buffer.duplicate.get(dest) 
  }```

Arrays.copyOfRange does System.arraycopy slightly differently: first it validates, then
System.arraycopy(original, from, copy, 0, Math.min(original.length - from, newLength))

helena:
@pkolaczk just a few suggestions

pedjak:
retest this please

pedjak:
retest this please

pedjak:
retest this please

pedjak:
retest this please

pedjak:
retest this please

pedjak:
retest this please

pedjak:
retest this please

pedjak:
retest this please

pedjak:
retest this please

pedjak:
retest this please

jacek-lewandowski:
retest this please

jacek-lewandowski:
retest this please

jacek-lewandowski:
retest this please

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/EmbeddedCassandra.scala
pkolaczk: ok

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/EmbeddedCassandra.scala
pkolaczk: The property trait does not represent a property value.
apply of what class do you have in mind?

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/EmbeddedCassandra.scala
pkolaczk: I only refactored the code and retained the original behaviour. Additionally, in this case a reasonable default is not easy to find - what are the reasonable defaults for ports of 3 cassandra nodes, assuming the ports must be different? Sometimes failing with an error is better than using a bad / non-obvious default.

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/EmbeddedCassandra.scala
pkolaczk: indeed, flatten.foreach is better idea

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ByteBufferUtil.scala
pkolaczk: I see what you're trying to achieve here, but now you introduced another problem that the dest array is not used at all in one of the branches.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ByteBufferUtil.scala
pkolaczk: getArray was just the old name taken from C*, but I agree, toArray is better

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ByteBufferUtil.scala
pkolaczk: and if it is not used, then allocating a dest array on the callsite is just a waste then

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/EmbeddedCassandra.scala
pkolaczk: Also, this checking here is actually superfluous to make Scala compiler happy. If it had a smarter typechecker, we could just remove the {{checkValueType}} call and the code would still typecheck.

helena: I believe I was thinking of an apply for whatever the implementer of InetAddressProperty would be, because I saw something invoking 'checkValueType'. But now that I think of it, 'checkValueType' is more an 'unapply' behavior.

pkolaczk: Yes, exactly.

helena:
BTW @pkolaczk @pedjak this branch compiles cleanly for me, no warnings.

jacek-lewandowski:
@helena not only for you - Jenkins tested it for 100 times :)

pkolaczk:
Addressed reviewers comments

PR #639, opened 2015-04-18
Allow joinWithCassandraTable() to work with solr_query predicate by tran...

doanduyhai:
...sforming partition key predicate into token-based predicate

Right now, everytime I use a where clause containing a `solr_query` predicate, I got an **com.datastax.driver.core.exceptions.InvalidQueryException: CQL Solr queries do not support partition key bounds.**

The fix would be to transform the join on partition key `partitionKey = :partitionKey` into `token(partitionKey) = token(:partitionKey)` so that it works.

It is necessary also to add an **ALLOW FILTERING** clause when we have a join with clustering column, for example:

`WHERE token(partitionKey) = token(:partitionKey)  AND clustering1 = :clustering1 ALLOW FILTERING`

,or when we have a non-empty **WHERE** clause (2nd index for example):

`WHERE token(partitionKey) = token(:partitionKey)  AND myIndex = :myIndex ALLOW FILTERING`

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
@@ -182,17 +182,26 @@ class CassandraJoinRDD[L, R] private[connector](
 
     logDebug("Generating Single Key Query Prepared Statement String")
     logDebug(s"SelectedColumns : $selectedColumnRefs -- JoinColumnNames : $joinColumnNames")
+    val partitionKeyColumns = tableDef.partitionKey.map(_.columnName)
+    val clusteringColumns = tableDef.clusteringColumns.map(_.columnName)
+    val joinColumns = joinColumnNames.map(_.columnName)
     val columns = selectedColumnRefs.map(_.cql).mkString(", ")
-    val joinWhere = joinColumnNames.map(_.columnName).map(name => s"${quote(name)} = :$name")
+    val joinWhere = joinColumns.map(name => {
+      if(partitionKeyColumns.contains(name))
+        s"token(${quote(name)}) = token(:$name)"
+      else
+        s"${quote(name)} = :$name"
+    })
     val limitClause = limit.map(limit => s"LIMIT $limit").getOrElse("")
     val orderBy = clusteringOrder.map(_.toCql(tableDef)).getOrElse("")
     val filter = (where.predicates ++ joinWhere).mkString(" AND ")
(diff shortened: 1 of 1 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -182,17 +182,26 @@ class CassandraJoinRDD[L, R] private[connector](
...
     logDebug("Generating Single Key Query Prepared Statement String")
     logDebug(s"SelectedColumns : $selectedColumnRefs -- JoinColumnNames : $joinColumnNames")
+    val partitionKeyColumns = tableDef.partitionKey.map(_.columnName)
+    val clusteringColumns = tableDef.clusteringColumns.map(_.columnName)
+    val joinColumns = joinColumnNames.map(_.columnName)
     val columns = selectedColumnRefs.map(_.cql).mkString(", ")
-    val joinWhere = joinColumnNames.map(_.columnName).map(name => s"${quote(name)} = :$name")
+    val joinWhere = joinColumns.map(name => {
+      if(partitionKeyColumns.contains(name))
+        s"token(${quote(name)}) = token(:$name)"
```
RussellSpitzer: This will break equivalency in hash collisions

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -182,17 +182,26 @@ class CassandraJoinRDD[L, R] private[connector](
...
     val orderBy = clusteringOrder.map(_.toCql(tableDef)).getOrElse("")
     val filter = (where.predicates ++ joinWhere).mkString(" AND ")
     val quotedKeyspaceName = quote(keyspaceName)
     val quotedTableName = quote(tableName)
+    val allowFiltering = if(where.predicates.size > 0 || joinColumns.intersect(clusteringColumns).size>0) "ALLOW FILTERING" else ""
     val query =
       s"SELECT $columns " +
         s"FROM $quotedKeyspaceName.$quotedTableName " +
-        s"WHERE $filter $limitClause $orderBy"
+        s"WHERE $filter $limitClause $orderBy $allowFiltering"
```
RussellSpitzer: I'm not sure what the use case for allowFiltering is here since we already have our partitions picked out. This would in essence perform the anti pattern of allow filtering (hitting all nodes on a single partition) is this required for something else like the "solr_query" field?

pkolaczk:
This was fixed in DSE 4.7 and Spark connector does not need any changes.

PR #640, opened 2015-04-20
SPARKC-128: Parameterized Where in JWCassTable

RussellSpitzer:
This is a fix for doing parameterized where clauses with joinWithCassandraTable as noted on the ML by
Craig Ingram	(Thanks for finding the bug!)

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/BoundStatementBuilder.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/BoundStatementBuilder.scala
@@ -4,19 +4,34 @@ import com.datastax.driver.core._
 import com.datastax.spark.connector.types.ColumnType
 import org.apache.spark.Logging
 
+/**
+ * Class for binding row-like objects into prepared statements. prefixVals
+ * is used for binding constant values into each bound statement. This supports parametrized
+ * .where clauses in [[com.datastax.spark.connector.rdd.CassandraJoinRDD]]
+ */
 private[connector] class BoundStatementBuilder[T](
     val rowWriter: RowWriter[T],
     val preparedStmt: PreparedStatement,
-    val protocolVersion: ProtocolVersion) extends Logging {
+    val protocolVersion: ProtocolVersion,
+    val prefixVals: Seq[Any] = Seq.empty) extends Logging {
 
   private val columnNames = rowWriter.columnNames.toIndexedSeq
   private val columnTypes = columnNames.map(preparedStmt.getVariables.getType)
   private val converters = columnTypes.map(ColumnType.converterToCassandra(_)(protocolVersion))
   private val buffer = Array.ofDim[Any](columnNames.size)
 
+  private val prefixConverted = for {
+    prefixIndex: Int <- 0 until prefixVals.length
+    prefixVal = prefixVals(prefixIndex)
+    prefixType = preparedStmt.getVariables.getType(prefixIndex)
+    prefixConverter =  ColumnType.converterToCassandra(prefixType)(protocolVersion)
+  } yield prefixConverter.convert(prefixVal)
+
   /** Creates `BoundStatement` from the given data item */
   def bind(row: T): RichBoundStatement = {
     val boundStatement = new RichBoundStatement(preparedStmt)
+    boundStatement.bind(prefixConverted: _*)
+
     rowWriter.readColumnValues(row, buffer)
     var bytesCount = 0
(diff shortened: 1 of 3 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/BoundStatementBuilder.scala
```diff
@@ -4,19 +4,35 @@ import com.datastax.driver.core._
...
-    val protocolVersion: ProtocolVersion) extends Logging {
+    val protocolVersion: ProtocolVersion,
+    val prefixVals: Seq[Any] = Seq.empty) extends Logging {
 
   private val columnNames = rowWriter.columnNames.toIndexedSeq
   private val columnTypes = columnNames.map(preparedStmt.getVariables.getType)
   private val converters = columnTypes.map(ColumnType.converterToCassandra(_)(protocolVersion))
   private val buffer = Array.ofDim[Any](columnNames.size)
 
+  private val prefixTypes = (0 until prefixVals.length).map(preparedStmt.getVariables.getType)
```
cin: unused. can safely remove.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/BoundStatementBuilder.scala
```diff
@@ -4,19 +4,35 @@ import com.datastax.driver.core._
...
+  private val prefixConverted = for {
```
cin: I assume the current implementation was done to maximize performance. I find the following easier to read though probably less performant. Using a view is debatable (just depends on how big prefixVals is). Since prefixVals is normally going to be small, it should probably not be used.

```
  private val prefixConverted = prefixVals.view.zipWithIndex.map { case (prefixVal, i) =>
    val prefixType = preparedStmt.getVariables.getType(i)
    ColumnType.converterToCassandra(prefixType)(protocolVersion).convert(prefixVal)
  }
```

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/BoundStatementBuilder.scala
RussellSpitzer: Good call, I forgot to remove that!

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/BoundStatementBuilder.scala
RussellSpitzer: I'm mostly trying to follow the same pattern that is used in the CassandraTableScan and I tend to like the for yields better than the maps recently but we could still do that with the `zipWithIndex` if that is more readable. I think the connector should make some general style decesions on way or another in the near future.

RussellSpitzer:
Thanks again @cin for the review and comment's i'll have another commit up tomorrow.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/BoundStatementBuilder.scala
jacek-lewandowski: @cin I don't think using view in this case is a good idea. This is going to be used for every processed row so we need to tune the performance as much as possible. If if used view, it would look for the converter and then convert each prefix val for each row.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/BoundStatementBuilder.scala
```diff
@@ -4,19 +4,35 @@ import com.datastax.driver.core._
...
+    prefixIndex: Int <- (0 until prefixVals.length)
+    prefixVal = prefixVals(prefixIndex)
```
jacek-lewandowski: @RussellSpitzer follow IntelliJ suggestions, remove unnecessary parentheses :smile:

jacek-lewandowski:
@RussellSpitzer, @cin this looks good.

The only thing I have is that maybe we should control somehow whether the columns in "where" predicate do not overlap the columns in "join" predicate and then throw some exception with meaningful message. I'm afraid that if we let Cassandra to inform about the problem, the user can be confused. Wdyt?

Hmmm... this PR made me come up with the idea to optimise queries with "where" clauses - if we specify equality predicates, we may safely remove those columns from the projections and therefore transfer less data over the network!

RussellSpitzer:
i'll put in another check for where and on clauses overlapping

PR #641, opened 2015-04-21
SPARKC-130: Upgrade to Spark 1.2.2 and enable Java API for Scala 2.11

jacek-lewandowski:


project/Versions.scala
```diff
@@ -17,7 +17,7 @@ import scala.util.Properties
  */
 object Versions {
 
-  val crossScala = Seq("2.11.5", "2.10.4")
+  val crossScala = Seq("2.11.6", "2.10.5")
```
helena: We don't want this?

Seq("2.11.5", "2.11.6", "2.10.4", "2.10.5")

pkolaczk: Why would you want 2.11.6 *and* 2.11.5 as well as 2.10.5 *and* 2.10.4? They are binary compatible.

helena: We would not. After writing that I realized it would cause more travis builds. But I never deleted my above comment.

pkolaczk:
+1

helena:
+1

PR #648, opened 2015-04-22
SPARKC-112 Spark datasource API integration

alexliu68:


pkolaczk:
```
[warn] Class org.apache.cassandra.io.util.DataOutputPlus not found - continuing with a stub.
[warn] Class org.apache.cassandra.io.util.DataOutputPlus not found - continuing with a stub.
[warn] non-nullary method overrides nullary method
[warn] three warnings found
```
This looks like based on pretty old 1.2.0 code; also not easily rebase'able - can you rebase it to recent origin/master? Thanks.

spark-cassandra-connector/src/it/scala/org/apache/spark/sql/cassandra/CassandraSQLPredicatePushdownSpec.scala
```diff
@@ -50,29 +54,29 @@ class CassandraSQLPredicatePushdownSpec extends SparkCassandraITFlatSpecBase {
 
   it should "allow predicate pushdown if all partition columns have EQ predicate.  a=1 and b=1 and c=1 and d =1 are pushed down" in {
     val plan = cc.sql("select * from test_pd where a=1 and b=1 and c=1 and d=1").queryExecution.sparkPlan
-    plan.nodeName shouldBe "CassandraTableScan"
-    plan.children should have size 0
+    plan.nodeName shouldBe "Filter"
+    plan.children should have size 1
```
pkolaczk: Not part of this ticket, but there are tests in this file authored by you, which lack those kinds of checks. E.g. one test above this, you have a test:
```scala
 it should "not allow predicate pushdown if some partition column doesn't have predicate" in {
    val result = cc.sql("select * from test_pd where a=1 and b=1 and c=1").collect()
  }
```
I have no idea how this test validates what it promises to validate. It doesn't check if the predicate pushdown was or wasn't really applied and it doesn't even check the correctness of the result. This ticket, when you already touched those predicate pushdown tests, is a good occasion to improve those other tests without proper checks.

spark-cassandra-connector/src/it/scala/org/apache/spark/sql/cassandra/CassandraSQLClusterLevelSpec.scala
```diff
@@ -51,12 +50,22 @@ class CassandraSQLClusterLevelSpec extends SparkCassandraITFlatSpecBase {
...
+  }
+
+  override def afterAll() {
+    super.afterAll()
+    conn.withSessionDo { session =>
+      session.execute("DROP KEYSPACE sql_test1")
+    }
+    conn2.withSessionDo { session =>
+      session.execute("DROP KEYSPACE sql_test2")
+    }
```
pkolaczk: You'd better use a different keyspace in every test suite, rather than relying on dropping/recreating them. In the future we will want to parallelize integration tests, because they are already taking longer and longer with every release. Fortunately this is hardly as bad as in DSE, but we're heading in that direction. If every test suite used a different keyspace that would be generally safer.

spark-cassandra-connector/src/it/scala/org/apache/spark/sql/cassandra/CassandraSQLSpec.scala
```diff
@@ -314,8 +321,264 @@ class CassandraSQLSpec extends SparkCassandraITFlatSpecBase {
...
+      session.execute("insert into sql_test.\"Upper_Case_Table\"(\"KEY\",\"VALUE\", d) values (1, 2, 'two')")
+      session.execute("insert into sql_test.\"Upper_Case_Table\"(\"KEY\",\"VALUE\", d) values (1, 3, 'three')")
+    }
+    cc.cassandraSql("select KEY, VALUE from Upper_Case_Table where VALUE > 1").collect() should have length 2
+  }
+
+  it should "not find non-exist tables" in {
+    an [ExecutionException] should be thrownBy {
+      cc.tableExists(TableIdent("non_exist", "non_exist"))
+    }
```
pkolaczk: This test surprises me. The documentation on the `tableExists` method doesn't mention throwing exceptions. It returns a Boolean. So my guess would be that this code returned false for non-existent table, not threw an exception.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/TableIdent.scala
```diff
@@ -0,0 +1,6 @@
+package org.apache.spark.sql.cassandra
+
+
+/** Store table identity including table name, keyspace name and option cluster name */
+case class TableIdent(table: String, keyspace: String, cluster: Option[String] = None)
```
pkolaczk: Can we use `TableRef` name to keep it consistent with `ColumnRef`?

pkolaczk: Why cluster is an Option[String] and keyspace is only a String? Can't we have a default keyspace?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/MetaStore.scala
```diff
@@ -0,0 +1,573 @@
...
+
+  /** Only Store customized tables meta data in metastore */
+  def storeTable(
+      tableIdent: TableIdent,
+      source: String,
+      schema: Option[StructType],
+      options: Map[String, String]) : Unit
+
+  /** Update table options */
+  def setTableSchema(tableIdent: TableIdent, scheamJsonString: String) : Unit
```
pkolaczk: typo

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/MetaStore.scala
```diff
@@ -0,0 +1,573 @@
...
+      | AND keyspace_name = '${tableIdent.keyspace}'
+      | AND table_name = '${tableIdent.table}'
+    """.stripMargin.replaceAll("\n", " ")
+    metaStoreConn.withSessionDo {
+      session => session.execute(deleteQuery)
+    }
+  }
+
+  /** Set table schema */
+  override def setTableSchema(tableIdent: TableIdent, scheamJsonString: String) : Unit = {
```
pkolaczk: typo (the same typo found also in other places - I'm not listing all of them in the comments)

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/MetaStore.scala
```diff
@@ -0,0 +1,573 @@
...
+      s"""
+      |INSERT INTO ${getMetaStoreTableFullName}
+      | (cluster_name, keyspace_name, table_name)
+      | values (?, ?, ?)
+    """.stripMargin.replaceAll("\n", " ")
+
+    val clusterName = cluster.getOrElse(sqlContext.getCluster)
+    metaStoreConn.withSessionDo {
+      session =>
+        val preparedStatement = session.prepare(insertQuery)
```
pkolaczk: Incorrect use of preparedStatement. 
This is not the critical path, so no need for prepared statements.
But even if it was on the critical path, statements must not be prepared per every query execution. 
Use simple  `session.execute(query, param1, param2, ...)`

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/MetaStore.scala
```diff
@@ -0,0 +1,573 @@
...
+      schema: Option[StructType],
+      options: Map[String, String]): Unit = {
+    import collection.JavaConversions._
+
+    val cluster = tableIdent.cluster.getOrElse(sqlContext.getCluster)
+    if (schema.nonEmpty) {
+      metaStoreConn.withSessionDo {
+        session =>
+          val preparedStatement = session.prepare(InsertIntoMetaStoreQuery)
+          session.execute(
```
pkolaczk: This pattern is repeated in many places of code, which means it is a nice candidate for being abstracted out to a separate method. The call site should look like:
```scala
execute(
  InsertIntoMetaStoreQuery, 
  cluster, 
  tableIdent.keyspace, 
  tableIdent.table, 
  source,
  ...
```

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSQLContext.scala
```diff
@@ -36,199 +31,48 @@ import collection.mutable
   *
   * }}} */
 class CassandraSQLContext(sc: SparkContext) extends SQLContext(sc) {
```
pkolaczk: This is the second thing called Cassandra SQL context and now I'm quite confused. The other one is in `org.apache.spark.sql.cassandra` package, in a form of an implicit wrapper over Spark SQLContext.

```scala
 /**
   * Add a method, `cassandraTable`, to SQLContext that allows reading from and writing to Cassandra.
   * Add methods to access cluster level, keyspace level and table level Cassandra configuration settings.
   */
  implicit class CSQLContext(sqlContext: SQLContext) {
```
Some similarities:
- both need an SQLContext to work (one uses inheritance, another one composition)
- both offer methods to access Cassandra data as `DataFrame`
- both have methods to change config / e.g. set some table-level options; 
- they look like an "entry point" of the API
- and they have almost the same name `CSQLContext` and `CassandraSqlContext`

However, they offer different APIs, and one persists its state in memory, while another in Cassandra (in the metastore). 

As a user of this, which class should I use? Why are they two different "things"? What is the main purpose of each of them? I can see no clear separation in their functionality.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSQLContext.scala
```diff
@@ -36,199 +31,48 @@ import collection.mutable
...
-    addClusterLevelCassandraConnConf(cluster, CassandraConnectorConf(conf))
-  }
-
-  /** Add cluster level write configuration settings */
-  def addClusterLevelCassandraConnConf(cluster: String, conf: CassandraConnectorConf) = {
-    validateClusterName(cluster)
-    clusterCassandraConnConf += cluster -> conf
-    this
-  }
-
```
pkolaczk: Oh, actually these methods removed here, has been moved to that other CSQLContext class. Why?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSQLContext.scala
alexliu68: CSQContext(This name can be any name) is an implicit class holder to add additional methods to SQLContext(a more general use), the class is not used to be called directly . CassandraSqlContext is used for CassandraCatalog(only narrow used), it should be called directly. They are different.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSQLContext.scala
alexliu68: CassandraSqlContext extends SQLContext, and those methods are implicitly added to SQLContext, so we removed all of them here.

alexliu68:
@pkolaczk  Before we reviewing the code style here. can we first discuss the API and functions in the Jira ticket, so once we agree that part, we can move on into code style reviewing?

alexliu68:
We need get more people to review the API and functions, so we all settle down on the API and functions changes. Once that's done, code style and other issues are easy to be done. @jacek-lewandowski  If you have time can you also review the API changes.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -0,0 +1,308 @@
...
+      case _ => (row: Row) => throw new RuntimeException(s"Unknown RichFilter $filter")
+    }
+
+    // a filter combining all other filters
+    val richFilters = preservedFilters.map(richFilter)
+    val translatedFilters = richFilters.map(translateFilter)
+    def rowFilter(row: Row): Boolean = translatedFilters.forall(_(row))
+
+    val transformer = new RddTransformer(requiredColumns, pushdownFilters)
+    transformer.transform(baseRdd).filter(rowFilter)
```
pkolaczk: Calling `filter` on the result RDD is not needed here, because Spark docs says explicitly, Spark applies the filters by itself. There is no need to rewrite parts of Spark here. And once filter is not needed, 90% of the logic of this class can be removed - that is, all lines between L163 and L271. To double check, I removed these lines and it didn't cause any unit nor integration tests to fail.

pkolaczk: Quote from the Spark docs on PrunedFilteredScan:
> A BaseRelation that can eliminate unneeded columns and filter using selected
 predicates before producing an RDD containing all matching tuples as Row objects.
> The pushed down filters are currently purely an optimization as *they will all be evaluated
 again*.  This means it is safe to use them with methods that produce false positives such
 as filtering partitions based on a bloom filter.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -0,0 +1,308 @@
...
+    if (sourceOptions.pushdown) {
+      new PrunedFilteredScanRelationImpl(
+        tableIdent = tableIdent,
+        connector = connector,
+        readConf = readConf,
+        writeConf = writeConf,
+        userSpecifiedSchema = sourceOptions.schema,
+        sqlContext = sqlContext)
+    } else {
+      new PrunedScanRelationImpl(
```
pkolaczk: Both `PrunedScanRelationImpl` and `PrunedFilteredScanRelationImpl` are very similar in implementation; the only difference is that one of them is doing predicate pushdown, while another one is not. So why not merge them into a single class with optional predicate pushdown enabled by a parameter?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -0,0 +1,308 @@
...
+
+  protected[this] val baseRdd =
+    sqlContext.sparkContext.cassandraTable[CassandraSQLRow](
+      tableIdent.keyspace,
+      tableIdent.table)(
+        connector,
+        readConf,
+        implicitly[ClassTag[CassandraSQLRow]],
+        CassandraSQLRowReader,
+        implicitly[ValidRDDType[CassandraSQLRow]])
```
pkolaczk: There is no need to specify all those implicit  parameter values here. It is enough to define an `implicit val` for `connector` and `readConf` immediately before the `cassandraTable` call.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -0,0 +1,308 @@
...
+      AllColumns,
+      writeConf)(
+        connector,
+        SqlRowWriter.Factory)
+  }
+
+  override def sizeInBytes: Long = {
+    val keyspace = tableIdent.keyspace
+    val table = tableIdent.table
+    val size = sqlContext.conf.getConf(s"spark.cassandra.$keyspace.$table.size.in.bytes", null)
```
pkolaczk: This should be first read from the Cassandra system table, and only if unavailable, then taken from the settings.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
alexliu68: We are creating a a bloom filter here why those code is removed? Why push it back to Catalyst ?

spark-cassandra-connector/src/it/scala/org/apache/spark/sql/cassandra/CassandraSQLPredicatePushdownSpec.scala
alexliu68: manual check it the test log for now. It's hard to check it for now. I will create another ticket to better test pushdown. Idea is to only test Pushdown class by passing the whole sql query process

spark-cassandra-connector/src/it/scala/org/apache/spark/sql/cassandra/CassandraSQLClusterLevelSpec.scala
alexliu68: good idea.

spark-cassandra-connector/src/it/scala/org/apache/spark/sql/cassandra/CassandraSQLSpec.scala
alexliu68: ExecutionException is a RuntimeException, so it's not caught by the method, Boolean should return.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/TableIdent.scala
alexliu68: Why keyspace is a String is trying to force user to specify the keyspace/database name. We can add a default keyspace name of "default"

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
alexliu68: Does it really make code simpler? PrunedScanRelationImpl and PrunedFilteredScanRelationImpl class names tell what each does. Making them separate class is easy to read.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
alexliu68: ok scala tricks:)

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
alexliu68: good idea

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
alexliu68: Another reason is that if a method implementing both PrunedFilteredScan and PrunedScan, Catalyst only identifies one type at runtime, and it's not possible to switch type.

So each class can only implement one interface. I tried implementing both interfaces, it did work, so I created separate classes.

pkolaczk: There is *no need* to implement both interfaces in the first place. Where did you read that?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
pkolaczk: The quoted documentation says Catalyst does filtering again *anyway*. Therefore, in your version filters are evaluated *twice*: the first time by your code, the second time by Catalyst. That's why removing your code does not change semantics.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
alexliu68: I tested it before, I didn't read it.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraCatalog.scala
```diff
@@ -1,65 +1,242 @@
...
+    synchronized {
+      metaStore.removeTableSchema(tableRef)
+    }
+    cachedDataSourceTables.refresh(catalystTableIdentFrom(tableRef))
   }
 
-  override def getTables(databaseName: Option[String]): Seq[(String, Boolean)] = ???
+  /** Convert Catalyst tableIdentifier to TableRef */
+  def tableRefFrom(tableIdentifier: Seq[String]) : TableRef = {
+    val id = processTableIdentifier(tableIdentifier).reverse.lift
```
RussellSpitzer: Might be nice to make a version of this which returns a named Tuple so we don't have the magic number indexing which happens a few times in the code

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -0,0 +1,296 @@
...
+    userSpecifiedSchema.getOrElse(StructType(tableDef.allColumns.map(toStructField)))
+  }
+
+  implicit val cassandraConnector = connector
+  implicit val readconf = readConf
+  protected[this] val baseRdd =
+    sqlContext.sparkContext.cassandraTable[CassandraSQLRow](tableRef.keyspace, tableRef.table)
+
+  override def insert(data: DataFrame, overwrite: Boolean): Unit = {
+    if (overwrite) {
```
RussellSpitzer: I'm not sure this is semantically what a C* user would expect.  I think we should not allow an insert to do a truncate in any case. I think that still follows the letter of the interface definition since overwriting will still over write the data of the "relation" even if it doesn't overwrite other existing data ...

> A BaseRelation that can be used to insert data into it through the insert method.
> If overwrite in insert method is true, the old data in the relation should be overwritten with
> the new data. If overwrite in insert method is false, the new data should be appended.

alexliu68: It depends on how you define "old data" as rows or as a table? I think overwite is to write data into C*, append should be same as overwrite. It's @pkolaczk's idea to truncate the table.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraCatalog.scala
alexliu68: Can you elaborate it more. The method converts Seq[String] to TableRef(table, keyspace, cluster).

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
RussellSpitzer: Brian Hess disagrees with me so the truncate is probably ok :)

RussellSpitzer: I think the truncate is the right thing to do after checking the JDBC implementation in spark 1.4 which calls a truncate

https://github.com/apache/spark/blob/213a6f30fee4a1c416ea76b678c71877fd36ef18/sql/core/src/main/scala/org/apache/spark/sql/DataFrame.scala#L1750-L1763

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraCatalog.scala
RussellSpitzer: Ah sorry i meant to post this on line 39, this is exactly the method I was looking for :)
```scala
val id = processTableIdentifier(tableIdentifier).reverse.lift
val tableName = id(0).getOrElse(throw new IOException(s"Missing table name"))
```

PR #655, opened 2015-04-30
SPARKC-139: Fix overflow bugs in RateLimiter

RussellSpitzer:
Two possible overflows fixed here,

Elapsed time * rate can overflow when lastTime is 0 --> Fixed this by starting last time at time()
rate at it's default will overflow since it is Int.MaxValue * 1024 * 1024 --> Fixed this by changing default to 1000000mbps

Also I set elapsed time to always be positive or 0 incase of ntp sync disrupting the clock and sending
time backwards.


jacek-lewandowski:
Review comments on hipchat

RussellSpitzer:
Addressed Comments, set the default rate back to Int.MaxValue, but make sure we multiply against Longs so we don't overflow. 

Put some basic requires in the construct to avoid infinite delays with 0 or negative rates
Added a test which tries setting various rates and makes sure everything works properly and returns to
a stable state.

jacek-lewandowski:
+1

PR #660, opened 2015-05-07
SPARKC-4: Mapping UDTs to case classes when reading from Cassandra.

pkolaczk:


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ReflectionUtil.scala
```diff
@@ -99,5 +106,20 @@ object ReflectionUtil {
     setters(implicitly[TypeTag[T]].tpe)
   }
 
+  /** Creates a corresponding `TypeTag` for the given `Type`.
+    * Allows to use reflection-created objects in APIs expecting `TypeTag`. */
+  def typeToTypeTag[T](tpe: Type): TypeTag[T] = {
```
RussellSpitzer: Maybe Synchronized on TypeTag?

RussellSpitzer:
The new columnMappers switch around the "as" ordering correct? If so I think we should make sure to clean up all of the docs and examples as well as post a very large warning for upgrading from 1.2 -> 1.3

RussellSpitzer:
Everything looks good to me, I've walked through this a couple times today and it all seems like a good improvement. +1

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ReflectionUtil.scala
pkolaczk: Oh yes, good catch. I don't know how many things are computed lazily underneath and are not thread-safe in 2.10, so it won't hurt to enclose it all in synchronized.

pkolaczk: Fixed.

pkolaczk:
Rebased to recent master and wrapped commit messages.

pkolaczk:
Rebased and resolved conflicts. Updated CHANGES.txt.

pkolaczk:
Rebased and resolved conflicts. Applied @jacek-lewandowski suggestions.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/GettableDataToMappedTypeConverter.scala
```diff
@@ -0,0 +1,254 @@
...
+  /** Returns a converter that can convert
+    * a Cassandra value of given `ColumnType` to Scala type given by `Type` */
+  private def converter(columnType: ColumnType[_], tpe: Type): TypeConverter[_] = {
+    type U = u forSome { type u }
+    implicit val tt = ReflectionUtil.typeToTypeTag[U](tpe)
+    converter[U](columnType)
+  }
+
+  /** Reads the column name of the column pointed by given column reference. */
+  private def columnName(columnRef: ColumnRef): String = {
```
jacek-lewandowski: Isn't it too trivial to make a method?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/GettableDataToMappedTypeConverter.scala
```diff
@@ -0,0 +1,254 @@
...
+      case ColumnName(columnName, _) =>
+        structDef.columnByName(columnName).columnType
+      case TTL(_, _) | WriteTime(_, _) | RowCountRef =>
+        BigIntType
+    }
+  }
+
+  /** Reads the value of the given column from the input data, without applying any conversions. */
+  private def columnValue(columnRef: ColumnRef, struct: GettableData): AnyRef = {
+    struct.getRaw(columnRef.cqlValueName)
```
jacek-lewandowski: Same as above - Isn't it too trivial to make a method?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/GettableDataToMappedTypeConverter.scala
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  /** Tries to convert the value with the given converter and handles the error, if any */
+  private def tryConvert(value: AnyRef, converter: TypeConverter[_], name: String): AnyRef = {
+    try
```
jacek-lewandowski: I don't know whether we have any rules for this - but it would be safer to avoid try-catch and other blocks without braces unless it is one-line statement, wdyt?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/GettableDataToMappedTypeConverter.scala
```diff
@@ -0,0 +1,254 @@
...
+    columnMap.setters.toArray.collect {
+      case (setterName, columnRef) if !constructorColumnRefs.contains(columnRef) =>
+        (methods(setterName), columnRef)
+    }
+  }
+
+  /** Uses setters of `T` to set values found in GettableData */
+  private def invokeSetters(data: GettableData, obj: T): Unit = {
+    for ((setter, columnRef) <- setters) {
+      val param = setterParamValue(setter.getName, data)
```
jacek-lewandowski: very very minor nit - imho it should be named `value`

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/GettableDataToMappedTypeConverter.scala
pkolaczk: Good catch. It was more complex at some time before (when I was working on this ticket), because it required pattern-matching. Now it doesn't, so I'll inline it.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/GettableDataToMappedTypeConverter.scala
pkolaczk: Ok, I'll inline.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/GettableDataToMappedTypeConverter.scala
pkolaczk: Safer? Why would it be any safer for expressions? If you add another line in the middle and forget braces, the expression wouldn't compile. I think it is only desired for imperative style, when you use if without else, so this could lead to a following bug:

```scala 
if (condition)
  doSth;
  doSthElse;   // this is not a part of if
someOtherCode
```
Here, it is an expression, not a statement, and adding anything in the middle with no braces would trigger a compile error.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/GettableDataToMappedTypeConverter.scala
```diff
@@ -0,0 +1,254 @@
...
+        TypeConverter.forType[U](Seq(keyConverter, valueConverter))
+      case (_, _) =>
+        TypeConverter.forType[U]
+    }
+  }
+
+  /** Returns a converter that can convert
+    * a Cassandra value of given `ColumnType` to Scala type given by `Type` */
+  private def converter(columnType: ColumnType[_], tpe: Type): TypeConverter[_] = {
+    type U = u forSome { type u }
```
jacek-lewandowski: IntelliJ complains - you should explicitly import this language feature.

pkolaczk: ok

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/GettableDataToMappedTypeConverter.scala
pkolaczk: indeed

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/GettableDataToMappedTypeConverter.scala
```diff
@@ -0,0 +1,254 @@
...
+    This method does not immediately call `TypeConverter.forType`, because `forType` doesn't know
+    the Cassandra `ColumnType` and therefore it can't instantiate appropriate type converters for
+    Cassandra user-defined-types, which could be nested. Therefore, for types that can nest other types
+    i.e. collections, options or UDTs, we make sure to use this method for any needed nested converters.
+    E.g. for a list type, we recursively call this method to get the converter for the
+    list items and then we call `TypeConverter.forType` to get a proper converter for lists.
+    */
+    val tpe = typeTag[U].tpe
+    (columnType, tpe) match {
+      case (argColumnType, TypeRef(_, Symbols.OptionSymbol, List(argScalaType))) =>
```
jacek-lewandowski: nit style comment: wdyt about inserting one empty line after each case?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/GettableDataToMappedTypeConverter.scala
```diff
@@ -0,0 +1,254 @@
...
+    implicitly[ColumnMapper[T]].columnMapForReading(structDef, columnSelection)
+
+  /** Returns the column mapper associated with the given type.
+    * Used to find column mappers for UDT columns.
+    * For tuples, returns [[com.datastax.spark.connector.mapper.TupleColumnMapper TupleColumnMapper]],
+    * for Java beans uses
+    * [[com.datastax.spark.connector.mapper.JavaBeanColumnMapper JavaBeanColumnMapper]],
+    * and for everything else uses
+    * [[com.datastax.spark.connector.mapper.DefaultColumnMapper DefaultColumnMapper]] */
+  private def columnMapper[U : TypeTag]: ColumnMapper[U] = {
```
jacek-lewandowski: So, the user cannot choose a custom column mapper for UDT fields?

pkolaczk: Not yet. Not saying it is impossible, but we'd need to use macros.

jacek-lewandowski:
ok, this looks very good to me
+1

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/GettableDataToMappedTypeConverter.scala
pkolaczk: ok

pkolaczk:
Reviewer suggestions applied, fixup commits squashed. Also squashed the first two commits, as they were too related and not very big.

PR #661, opened 2015-05-11
DOC: Added "Using Implicits for Configuration" example

RussellSpitzer:


pkolaczk:
+1

doc/2_loading.md
```diff
@@ -306,4 +306,29 @@ The following options can be specified in the SparkConf object or as a jvm
...
+code rather than repeatedly passing the same `ReadConf` object.
+
+```scala
+object ReadConfigurationOne {
+  implicit val readConf = ReadConf(100,100)
+}
+import ReadConfigurationOne._
+val rdd = sc.cassandraTable("write_test","collections")
+rdd.readConf
+//com.datastax.spark.connector.rdd.ReadConf = ReadConf(100,100,LOCAL_ONE,true)
```
helena: IMO you should have added the link vs //etc
https://github.com/datastax/spark-cassandra-connector/blob/master/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/ReadConf.scala#L7-L18

helena: Otherwise when the api changes you have to also modify this

PR #663, opened 2015-05-13
SPARKC-155: Get RPartitioner Hosts from Driver

RussellSpitzer:
Previously the code used CassandraConnector().hosts which only
used those hosts passed in the SparkConf object. Now host
retrevial is done via the Java Driver cluster metadata
object.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnector.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnector.scala
@@ -49,8 +49,11 @@ class CassandraConnector(conf: CassandraConnectorConf)
 
   private[this] var _config = conf
 
-  /** Known cluster hosts. This is going to return all cluster hosts after at least one successful connection has been made */
-  def hosts = _config.hosts
+  /** Known cluster hosts in the connected datacenter.*/
+  lazy val hosts: Set[InetAddress] =
+    // wrapped in a session, so we get full lists of hosts,
+    // not only those explicitly passed in the conf
+    withSessionDo { _ => _config.hosts }
 
   /** Configured native port */
   def nativePort = _config.nativePort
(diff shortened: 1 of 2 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
```diff
@@ -17,7 +20,7 @@ class ReplicaPartitioner(partitionsPerReplicaSet: Int, connector: CassandraConne
   /* TODO We Need JAVA-312 to get sets of replicas instead of single endpoints. Once we have that we'll be able to
   build a map of Set[ip,ip,...] => Index before looking at our data and give the all options for the preferred location
    for a partition*/
-  private val hosts = connector.hosts.toVector
+  private val hosts = connector.withClusterDo(_.getMetadata.getAllHosts).map(_.getAddress).toVector
```
pkolaczk: `CassandraConnector` fetches all hosts on the first connection, but it is lazy:
https://github.com/datastax/spark-cassandra-connector/blob/b1.2/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnector.scala#L75-77

Maybe we should move that code to the "CassandraConnector#hosts" method to avoid surprises?

pkolaczk:
Generally the change looks good, but I think it would be even better if it improved the `CassandraConnector#hosts` method to behave deterministically :)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
RussellSpitzer: Sounds good to me, I also realized that this implementation may have some issues on multi-dc clusters. So I think I need to make a slight change around that as well

pkolaczk:
+1

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnector.scala
```diff
@@ -72,10 +77,6 @@ class CassandraConnector(conf: CassandraConnectorConf)
   def openSession() = {
     val session = sessionCache.acquire(_config)
     try {
-      val allNodes = session.getCluster.getMetadata.getAllHosts.toSet
-      val myNodes = LocalNodeFirstLoadBalancingPolicy.nodesInTheSameDC(_config.hosts, allNodes).map(_.getAddress)
-      _config = _config.copy(hosts = myNodes)
```
pkolaczk: Having looked at this once again I noticed a problem. Now when you call `openSession()`, the hosts do not get copied to _config, because `hosts` would be never called.

RussellSpitzer:
@pkolaczk Rebased and squashed

pkolaczk:
+1

PR #664, opened 2015-05-14
SPARKC-157: Move type conversions back to RowWriters.

pkolaczk:
RowWriter is always created at the application driver site and
serialized to Spark executors. RowWriter also has full
type information about the columns it writes. Therefore
it is a good place to instantiate proper TypeConverters and
completely handle the process of converting the RDD item object of
some type T to values that can be passed to the Java Driver API for
saving as a Cassandra row. When TypeConverters are instantiated
together with RowWriter, custom TypeConverters registered in the
driver application can be found.

Unfortunately this approach requires changes to how
TypeConverters are obtained. Some type converters, like the converter
used for converting UDTValues to objects passable to the Java Driver
API, depend strongly on Java Driver objects like DataType or
ProtocolVersion. Some of those objects are not Serializable and would
prevent sending such a TypeConverter to Spark Executors. Therefore in
this commit the conversion process has been split into two phases.

First phase of conversion is done by the TypeConverters associated
with the RowWriter, as described in the first paragraph. Those
TypeConverters are based on ColumnTypes and
don't use any Java Driver APIs, hence they are Serializable. However,
this way they would not be able to create UDTValue objects required
by the Java Driver.

So in the second phase, another set of
TypeConverters is obtained on the Executor side. They do not need to
be Serializable, they are based on ColumnDefinitions / DataTypes
obtained from the Java Driver and they know the right ProtocolVersion.
They are able to convert connector's UDTValue to Java Driver UDTValue.

Note that in the second phase there is no access to user-registered
TypeConverters. If you want to register a custom converter creating
UDT values, you must register a converter converting to connector's
UDTValue class, not Java Driver's UDTValue. This way it would be
properly selected and used by the RowWriter.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,8 @@
+1.2.2 (unreleased)
+ * Fixed a bug preventing a custom type converter from being used
+   when saving data. RowWriter implementations must
+   perform type conversions now. (SPARKC-157)
+
 1.2.1
  * Fixed problems with mixed case keyspaces in ReplicaMapper (SPARKC-159)
 
--- a/doc/6_advanced_mapper.md
+++ b/doc/6_advanced_mapper.md
@@ -95,6 +95,38 @@ use the new converter for.
 Additionally, defining the `StringToEMailConverter` as an implicit object 
 allows to use generic `CassandraRow#get` with your custom `EMail` field type.
  
+The following table specifies the relationship between a Cassandra column type and 
+the type of needed `TypeConverter`.
+
+Cassandra column type | Object type to convert from / to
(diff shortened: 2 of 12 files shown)


pkolaczk:
agreed, I'll add it immediately tomorrow morning, today I'm too tired

pkolaczk:
retest this please

sdelmas:
retest this please

pkolaczk:
retest this please

pkolaczk:
@RussellSpitzer can you review my docs changes?

jacek-lewandowski:
+1 on docs

PR #665, opened 2015-05-14
Fixed problems with case sensitive keyspaces

jacek-lewandowski:


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/ReplicaMapper.scala
```diff
@@ -67,7 +67,7 @@ class ReplicaMapper[T] private(
         val clusterMetadata = session.getCluster.getMetadata
         data.map { row =>
           val hosts = clusterMetadata
-            .getReplicas(keyspaceName, routingKeyGenerator.apply(boundStmtBuilder.bind(row)))
+            .getReplicas(Metadata.quote(keyspaceName), routingKeyGenerator.apply(boundStmtBuilder.bind(row)))
```
RussellSpitzer: +1

PR #668, opened 2015-05-15
SPARKC-112 Implement basic data source API functions 

alexliu68:


spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -0,0 +1,181 @@
...
+
+  /** Quote name */
+  private[this] def quoted(str: String): String = {
+    "\"" + str + "\""
+  }
+
+  /** Add column selection and where clauses on top of baseRdd */
+  private[this] class RddTransformer(requiredColumns: Array[String], filters: Seq[Filter]) {
+
+    /** CassandraRDD[CassandraSQLRow] is the only type supported for transferring */
```
RussellSpitzer: This comment is confusing to me, i'm not sure why this statement is true or what the implications are?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -0,0 +1,181 @@
...
+    if(filterPushdown) {
+      logInfo(s"pushdown filters: ${pushdownFilters.toString()}")
+    }
+
+    val transformer = new RddTransformer(requiredColumns, pushdownFilters)
+    transformer.transform(baseRdd).asInstanceOf[RDD[Row]]
+  }
+
+  /** Quote name */
+  private[this] def quoted(str: String): String = {
```
RussellSpitzer: I think we have a few other implementations of a quote function elsewhere in the connector, so maybe this would be a good time to consolidate?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -0,0 +1,181 @@
...
+    "\"" + str + "\""
+  }
+
+  /** Add column selection and where clauses on top of baseRdd */
+  private[this] class RddTransformer(requiredColumns: Array[String], filters: Seq[Filter]) {
```
RussellSpitzer: Why does this needs to be a Class and not an Object? Also the naming may be clearer if this was something like RddPushdown, or PushdownToCql?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -0,0 +1,181 @@
...
+
+    /** Construct Cql clause and retrieve the values from filter */
+    private def filterToCqlAndValue(filter: Any): (String, Seq[Any]) = {
+      filter match {
+        case sources.EqualTo(attribute, value)            => (s"${quoted(attribute)} = ?", Seq(value))
+        case sources.LessThan(attribute, value)           => (s"${quoted(attribute)} < ?", Seq(value))
+        case sources.LessThanOrEqual(attribute, value)    => (s"${quoted(attribute)} <= ?", Seq(value))
+        case sources.GreaterThan(attribute, value)        => (s"${quoted(attribute)} > ?", Seq(value))
+        case sources.GreaterThanOrEqual(attribute, value) => (s"${quoted(attribute)} >= ?", Seq(value))
+        case sources.In(attribute, values)                 =>
```
RussellSpitzer: I think we may want to change this (IN Clause) to a parallelized operations later with joinWithCassandraTable or something like that, but we can save that for a later pr.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
alexliu68: right

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
alexliu68: Class is easy to group a few methods together. PushdownToCql is better name.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
alexliu68: Limit the type only to CassandraRDD[CassandraSQLRow]. Change it to "Define a type for CassandraRDD[CassandraSQLRow]. It's used by following methods."

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
alexliu68: new pr will be opened for this

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
alexliu68: I will remove this class and move all the methods to parent class

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/Quote.scala
```diff
@@ -0,0 +1,7 @@
+package com.datastax.spark.connector.util
+
+object Quote {
+
+   def quote(name: String): String = "\"" + name + "\""
```
pkolaczk: Don't we need to escape " here?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/TableWriter.scala
```diff
@@ -39,9 +40,6 @@ class TableWriter[T] private (
     case _ => None
   }
 
-  private def quote(name: String): String =
-    "\"" + name + "\""
-
```
pkolaczk: That's a nice change to put them in one place :)

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/DataTypeConverter.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import com.datastax.spark.connector
+import com.datastax.spark.connector.cql.ColumnDef
+import com.datastax.spark.connector.types.FieldDef
+import org.apache.spark.Logging
+import org.apache.spark.sql.types.StructField
+import org.apache.spark.sql.{types => catalystTypes}
+
+/** Convert Cassandra data type to Catalyst data type */
+object DataTypeConverter extends Logging {
```
pkolaczk: Maybe put it into ColumnType object so it is together with all the code for type conversions?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/DefaultSource.scala
```diff
@@ -0,0 +1,180 @@
...
+ *       c_table "table",
+ *       keyspace "keyspace",
+ *       cluster "test_cluster",
+ *       push_down "true",
+ *       spark_cassandra_input_page_row_size "10",
+ *       spark_cassandra_output_consistency_level "ONE",
+ *       spark_cassandra_connection_timeout_ms "1000"
+ *      )
+ */
+class DefaultSource extends RelationProvider with SchemaRelationProvider with CreatableRelationProvider with Logging {
```
pkolaczk: We have a `CassandraSourceRelation`, and this is a `Provider` for it.
Why not call it `CassandraSourceRelationProvider`?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/DefaultSource.scala
```diff
@@ -0,0 +1,180 @@
...
+  // Dot is not allowed in Options key for Spark SQL parsers, so convert . to _
+  // Map converted property to origin property name
+  // TODO check SPARK 1.4 it may be fixed
+  private val convertedPropertiesMap : Map[String, String] = {
+    val properties = ReadConf.Properties ++ WriteConf.Properties ++ CassandraConnectorConf.Properties
+    properties.map(prop => (prop.replace(".", "_"), prop)).toMap
+  }
+
+  /** Construct ReadConf, WriteConf, CassandraConnectorConf, tableSizeInBytes from options */
+  def buildConfs(
```
pkolaczk: Can't it be split into separate methods each returning just one thing?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/FilterPushdown.scala
```diff
@@ -0,0 +1,158 @@
...
+ *     only one predicate is allowed.
+ *  5. For cluster column predicates, only last predicate can be non-EQ predicate
+ *     including IN predicate, and preceding column predicates must be EQ predicates.
+ *     If there is only one cluster column predicate, the predicates could be any non-IN predicate.
+ *  6. There is no pushdown predicates if there is any OR condition or NOT IN condition.
+ *  7. We're not allowed to push down multiple predicates for the same column if any of them
+ *     is equality or IN predicate.
+ *
+ */
+object FilterPushdown {
```
pkolaczk: A déjà vu :) Now I get what you're trying to say back then that you had to redo the Catalog, too. 

So, we have a problem that previously our predicate pushdown code worked on `Expression`s. But new Spark introduced `Filter`, which is a special kind of an `Expression` for filtering, yet it doesn't implement the internal `Expression` class. So `Filter` and `Expression` are two classes serving almost the same purpose for us, but they are separate and have no common interface. Let's discuss how to get out of this trap.

Proposed solutions:
  1. Forget Expressions. Rewrite everything to use Filters. This way we get elegant code, possibly less dependent on internal Spark APIs (which might break any time), but we're not finished until we fix all the other code and the size of the changeset gets through the roof.
  1. Use old-school engineering. Copy-paste the whole thing, then find-and-replace `Expression` with `Filter`, slightly fix a few places where it doesn't compile and we're done!  BTW: This is the current solution here.
  1. Write *common* code that works both for `Expressions` and `Filters`. DRY magic.

@jacek-lewandowski @RussellSpitzer @helena @pedjak @bcantoni which one do you like the most?
The voting starts here; you can of course add more solutions.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/Quote.scala
alexliu68: yes

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/DefaultSource.scala
alexliu68: DefaultSource is the name conversion used by Spark Resources, so you don't need add class name to provider name when create a source. e.g. provider "org.apache.spark.sql.cassandra". If change it to CassandraSourceRelationProvider, provider "org.apache.spark.sql.casssandra.CassandraSourceRelationProvider". Spark SQL Internally Json, JDBC sources all use DefaultSource, so it's better to keep the  "DefaultSource"

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/DefaultSource.scala
alexliu68: sure

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/FilterPushdown.scala
RussellSpitzer: If I ranked by preference right now

2
3
1

My rational being we don't know how stable these api's will be anyway and I feel like take the least effort for us to be compatible now.  I think 3 is a little cleaner for the code base. 1 is a lot of work which makes me feel like we are just chasing Spark Apis. I don't really have confidence that the Spark project will not to make changes big enough to totally disrupt the api to the point where we have to rewrite everything again.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/DefaultSource.scala
pkolaczk: Ok, fair enough.

alexliu68:
@RussellSpitzer @pkolaczk I merged SPARKC-166 into it and replace pushdown by PredicatePushDown. It's ready for final reviewing.

RussellSpitzer:
+1

alexliu68:
travis-ci is down/blocked. Run tests at local.

PR #669, opened 2015-05-18
SPARKC-155: Update set of hosts in _config in openSession

pkolaczk:
A full list of hosts should be saved in _config when opening
a new session, for the future use, in case some of the initial
hosts go down.


pkolaczk:
@RussellSpitzer what about this?

RussellSpitzer:
Lgtm +1

pkolaczk:
Will you merge it into your branch?

PR #671, opened 2015-05-19
SPARKC-166: Refactoring: Decouple PredicatePushDown from SparkSQL

pkolaczk:
PredicatePushDown class was tied to Spark code by depending on Spark's
internal Expression class. This caused a problem when we wanted to
unit-test it separately or when we wanted to support other
representation of predicates like the new Filter class introduced by
Spark Data Sources API.

Therefore in this ticket, PredicatePushDown class has been moved out
from CassandraStrategies into its own file. Then all Spark-specific
code has been factored out to a separate PredicateOps typeclass.
PredicatePushDown has been generified to take predicate class type
parameter. This way, it can work with any type of predicate that is
in PredicateOps. PredicateOps is available for Expression and Filter.
It is very easy to add support for other types of predicates in the
future without changing PredicatePushDown code. Additionally, there
is less risk Spark API changes ever break our unit-tests.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,9 @@
 1.3.x (unreleased)
  * Removed use of Thrift describe_ring and replaced it with native Java Driver
    support for fetching TokenRanges (SPARKC-93)
+ * Decoupled PredicatePushDown logic from Spark (SPARKC-166)
+   - added support for Filter and Expression predicates
+   - improved code testability and added unit-tests
 
 ********************************************************************************
 
--- a/spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraStrategies.scala
+++ b/spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraStrategies.scala
@@ -6,7 +6,6 @@ import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.planning.PhysicalOperation
 import org.apache.spark.sql.catalyst.plans.logical
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
-import org.apache.spark.sql.columnar.InMemoryRelation
 import org.apache.spark.sql.execution.SparkPlan
 
 private[cassandra] trait CassandraStrategies {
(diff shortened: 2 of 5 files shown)


alexliu68:
+1

CHANGES.txt
```diff
@@ -1,6 +1,9 @@
 1.3.x (unreleased)
  * Removed use of Thrift describe_ring and replaced it with native Java Driver
    support for fetching TokenRanges (SPARKC-93)
+ * Decoupled PredicatePushDown logic from Spark (SPARKC-166)
```
pedjak: the content of `changes.txt` could autogenerated from Jira, with a bit of configuring Jira itself, and some scripting. I did that in the past in several project, works charming... Interest?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/PredicateOps.scala
```diff
@@ -0,0 +1,93 @@
...
+
+/** A unified API for predicates, used by [[PredicatePushDown]].
+  *
+  * Keeps all the Spark-specific stuff out of `PredicatePushDown`
+  * It is also easy to plug-in custom predicate implementations for unit-testing.
+  */
+trait PredicateOps[Predicate] {
+
+  /** Returns the column name of the only column in the predicate.
+    * Will throw `IllegalArgumentException` if predicate is not a single-column predicate. */
```
pedjak: question for all methods: is null as param value allowed? If not, what is going to happen if the null is submitted?

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
```diff
@@ -0,0 +1,241 @@
...
+
+  it should "not push down predicates on regular non-indexed columns if indexed ones are not included" in {
+    val f1 = EqFilter("r1")
+    val f2 = EqFilter("r2")
+    val ppd = new PredicatePushDown(Seq[Filter](f1, f2), table)
+    ppd.predicatesToPushDown shouldBe empty
+    ppd.predicatesToPreserve should contain allOf(f1, f2)
+  }
+
+  it should "not rely on given predicates order (1)" in {
```
pedjak: perhaps you can give more details about the test scenario - instead of adding 1, 2, 3 suffix

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraStrategies.scala
```diff
@@ -24,150 +23,15 @@ private[cassandra] trait CassandraStrategies {
...
-      val predicatesToPreserve = predicates.filterNot(predicatesToPushDown.toSet.contains)
-    }
-
-
     def apply(plan: LogicalPlan): Seq[SparkPlan] = plan match {
       case PhysicalOperation(projectList, predicates, relation: CassandraRelation) =>
         logInfo(s"projectList: ${projectList.toString()}")
         logInfo(s"predicates: ${predicates.toString()}")
-        val pushDown = new PredicatePushDown(predicates, relation)
+        val pushDown = new PredicatePushDown(predicates, relation.tableDef)
```
pedjak: Have you though about decoupling things even further by a `PredicatePushDownBuilder`, that is going to be injected into this component and used for constructing instance of `PredicatePushDown`?  This would allow us to extend testability of `CassandraStrategies`.

CHANGES.txt
pkolaczk: Yes, sounds charming, but isn't the quality/readability of the CHANGES.txt going to decrease? In CHANGES.txt we're often putting information what has *actually* been done, while JIRA often contains *what needs to be done*. Sometimes those things are different. For similar reason generating CHANGES.txt directly from the git-log is a bad idea.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraStrategies.scala
pkolaczk: So you mean, injecting a factory into CassandraStrategies instead a fixed implementation? Sound good, but I'd wait until Alex is done with his DataSource API related changes. Things are going to change a lot there, so let's do that afterwards.

CHANGES.txt
pedjak: Yes, but you can introduce additional field in Jira for CHANGES purpose. Hence, you do not have have each Jira issue mentioned in CHANGES.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraStrategies.scala
pedjak: Yes, this is what I thought..

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/PredicateOps.scala
pkolaczk: *Nulls are never permitted* in Scala, except two cases:
1. interfacing with Java code; nulls should be immediately de-nullified by wrapping them into Options.
2. critical performance path, where overhead of an Option might be too large and we know the compiler won't elide it; but then they should be very limited in scope - possibly only method local, or class-private. And it should be explicitly documented why nulls were needed.

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
pkolaczk: hmm, ok, let's call it (permutation 1, 2, 3); will it be ok?

CHANGES.txt
pkolaczk: Ah, that sounds cool. Yes, please :)

pkolaczk: How and when CHANGES.txt would be updated?

pedjak: Before releasing... you have script that generates it out Jira and then you commit it into Git.

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+
+  val table = TableDef(
+    keyspaceName = "test",
+    tableName = "test",
+    partitionKey = Seq(pk1, pk2),
+    clusteringColumns = Seq(c1, c2, c3),
+    regularColumns = Seq(i1, i2, r1, r2)
+  )
+
+  "PredicatePushDown" should "push down an equality partition key predicate" in {
```
pedjak: the test scenario speaks about *ONE* equality key predicate, but the fixture defines two? Needed?

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+    val f1 = EqFilter("pk1")
+    val f2 = EqFilter("pk2")
+    val ppd = new PredicatePushDown(Set[Filter](f1, f2), table)
+    ppd.predicatesToPushDown should contain allOf(f1, f2)
+    ppd.predicatesToPreserve shouldBe empty
+  }
+
+  it should "not push down a partition key predicate for a part of the partition key (1)" in {
```
pedjak: this test and the next one are identical, right?

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  it should "not push down a partition key predicate for a part of the partition key (2)" in {
+    val f1 = EqFilter("pk2")
+    val ppd = new PredicatePushDown(Set[Filter](f1), table)
+    ppd.predicatesToPushDown shouldBe empty
+    ppd.predicatesToPreserve should contain only f1
+  }
+
+  it should "not push down a range partition key predicate" in {
```
pedjak: the previous two tests are not needed, because this tests test the same paths, just with different input. If the tested path is the same, the number of test doing that does not bring more confidence.

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+    val f1 = EqFilter("pk1")
+    val f2 = RangeFilter("pk2")
+    val ppd = new PredicatePushDown(Set[Filter](f1, f2), table)
+    ppd.predicatesToPushDown shouldBe empty
+    ppd.predicatesToPreserve should contain allOf(f1, f2)
+  }
+
+  it should "push down an IN partition key predicate on the last partition key column" in {
```
pedjak: if it is about in filter, why not using just InFilter as input? EqFilter is already checked in other tests?

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+
+  it should "not push down an IN partition key predicate on the non-last partition key column" in {
+    val f1 = InFilter("pk1")
+    val f2 = EqFilter("pk2")
+    val ppd = new PredicatePushDown(Set[Filter](f1, f2), table)
+    ppd.predicatesToPushDown shouldBe empty
+    ppd.predicatesToPreserve should contain allOf(f1, f2)
+  }
+
+  it should "push down the first clustering column predicate" in {
```
pedjak: This test check the same path as "push down an equality partition key predicate", so I would drop it.

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+    val f1 = EqFilter("c1")
+    val ppd = new PredicatePushDown(Set[Filter](f1), table)
+    ppd.predicatesToPushDown should contain only f1
+    ppd.predicatesToPreserve shouldBe empty
+  }
+
+  it should "push down the first and the second clustering column predicate" in {
```
pedjak: The test covers the same path as the previous test, right?

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
pkolaczk: No, they are not. Look again.

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
pkolaczk: No, because the second one restricts pk2, which is a different column.

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
pedjak: ok

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
pkolaczk: Ok, so probably the wording of the test should be different... It should push down all predicates related to the partition key (and key is compound here, because it consists of 2 columns) if they include all the partition key columns. I'll fix the description.

spark-cassandra-connector/src/test/scala/org/apache/spark/sql/cassandra/PredicatePushDownSpec.scala
pkolaczk: Functionally they are different and cover different usecases. It checks if a range predicate gets pushed down correctly. It is possible for an incorrect code to pass the previous test but fail this one. I can also imagine code failing the previous one and passing this one; if for example someone hardcoded a restriction that the last predicate must be a range.

pedjak:
+1

PR #675, opened 2015-05-21
SPARKC-162 Add keyspce level/cluster level Conf settings

alexliu68:
Store keyspace level settings as cluster.keyspace/xx.xx.xx.xx
and cluseter level settings as cluster/xx.xx.xx.xx in
SQLConf

Retrieve settings in the order of
table level -> keyspace -> cluster -> default. Use the one
first available.


pkolaczk:
+1

PR #677, opened 2015-05-21
SPARKC-129 Add custom UUIDType and InetAddressType

alexliu68:
Add custom UUIDType and InetAddressType

UUIDType uses JVM data type UUID as base
InetAddressType uses JVM data type InetAddress as base

### DIFF
--- a/spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/types/InetAddressType.scala
+++ b/spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/types/InetAddressType.scala
@@ -39,12 +39,12 @@ class InetAddressType private() extends NativeType with PrimitiveType with Loggi
   //
   // Convert to host address to compare InetAddress because it doesn't support comparison.
   // It's not a good solution though.
-  private[sql] val ordering = new Ordering[JvmType] {
-    def compare(x: InetAddress, y: InetAddress) = {
-      logWarning("InetAddress doesn't support comparison. Convert it to host address to compare.")
-      x.getHostAddress.compareTo(y.getHostAddress)
-    }
-  }
+  import Ordering.Implicits._
+  def unsignedByte(x: Byte) = (x + 256) % 256
+  def address(inet: InetAddress): Iterable[Int] =
+    inet.getAddress.map(unsignedByte)
+  val ord = Ordering by address
+  private[sql] val ordering = ord
   /**
    * The default size of a value of the InetAddressType is 16 bytes.
    */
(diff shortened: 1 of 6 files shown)


spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/types/InetAddressType.scala
```diff
@@ -0,0 +1,45 @@
...
+  // this type. Otherwise, the companion object would be of type "InetAddressType$" in byte code.
+  // Defined with a private constructor so the companion object is the only possible instantiation.
+  private[sql] type JvmType = InetAddress
+  @transient private[sql] lazy val tag = ScalaReflectionLock.synchronized { typeTag[JvmType] }
+
+  // Convert to host address to compare InetAddress because it doesn't support comparison.
+  // It's not a good solution though.
+  private[sql] val ordering = new Ordering[JvmType] {
+    def compare(x: InetAddress, y: InetAddress) = {
+      logWarning("InetAddress doesn't support comparison. Convert it to host address to compare.")
```
pkolaczk: Wouldn't that cause too many warnings if e.g. Spark uses this code for... sorting?
And also this warning contradicts the code, because actually this InetAddressType offers a way to compare two InetAddresses. So if it can compare two InetAddresses, why it says it can't? What the user would do with this?

pkolaczk:
Please add tests.
Please update CHANGES.txt.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/types/InetAddressType.scala
alexliu68: IP address string comparison may not be what user wants, so warn user not to use InetAddress column in where clauses. But if they do, the message tells them what comparison is under the hood.  It will cause many warnings. Another option is to throw exception to block inetAddress ordering.

pkolaczk: So why do a String comparison if you could do a proper comparison?
```scala
def unsignedByte(x: Byte) = (x + 256) % 256
def addressTuple(inet: InetAddress) = 
  inet.getAddress.map(unsignedByte) match { case Array(a, b, c, d) => (a, b, c, d) } 
val ord = Ordering by addressTuple
ord.compare(x, y) 
```

alexliu68:
retest this please

RussellSpitzer:
This needs a documentation update for

>  // Because this new type is not a Spark internal supported data type, there is no
>  // CAST method to convert a InetAddress to a String. Spark internally converts it
>  // to Double, so it errors out the query such as
>  //    select * from table where inet_address_column = '/74.125.239.135'
>  // The correct query should be
>  //    select * from table where CAST(inet_address_column as string) = '/74.125.239.135

but +1 once a doc update is in

pkolaczk:
Please rebase onto b1.3. Also if there are conflicts, we agreed on not doing merges from the main branch into the topic branch, but we use rebase instead. This is because rebasing can be done multiple times, while merging changes from the main branch makes rebasing / squashing commits much harder.

PR #694, opened 2015-05-29
SPARKC-176: Add Guava back to Assembly Jar

RussellSpitzer:
Since this is shaded in Spark 1.2+ we need to include it in the
Connector Jar to make sure that the Java Driver has Guava.


pkolaczk:
+1

PR #695, opened 2015-05-29
SPARKC-104 Add partitionKeyMapper to repartitionByCassandraReplica()

RussellSpitzer:
Previously we relied on the implicit mapping of rdd to CassandraRows to
determine a the Primary Key present in the RDD. This patch allows the
user to specify using a column mapper the mapping between the partition
key and the RDD. The interface is the same as if they were saving to
Cassandra.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
@@ -25,12 +25,14 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
    * Saves the data from [[org.apache.spark.rdd.RDD RDD]] to a Cassandra table. Uses the specified column names.
    * @see [[com.datastax.spark.connector.writer.WritableToCassandra]]
    */
-  def saveToCassandra(keyspaceName: String,
-                      tableName: String,
-                      columns: ColumnSelector = AllColumns,
-                      writeConf: WriteConf = WriteConf.fromSparkConf(sparkContext.getConf))
-                     (implicit connector: CassandraConnector = CassandraConnector(sparkContext.getConf),
-                      rwf: RowWriterFactory[T]): Unit = {
+  def saveToCassandra(
+    keyspaceName: String,
+    tableName: String,
+    columns: ColumnSelector = AllColumns,
+    writeConf: WriteConf = WriteConf.fromSparkConf(sparkContext.getConf))(
+  implicit
+    connector: CassandraConnector = CassandraConnector(sparkContext.getConf),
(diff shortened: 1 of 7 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
```diff
@@ -152,10 +155,12 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
    * of the data specified by that row.
    * The calling RDD must have rows that can be converted into the partition key of the given Cassandra Table.
    */
-  def keyByCassandraReplica(keyspaceName: String, tableName: String)
+  def keyByCassandraReplica(keyspaceName: String,
```
jacek-lewandowski: codestyle

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
```diff
@@ -138,11 +138,14 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
    * event.
    * The calling RDD must have rows that can be converted into the partition key of the given Cassandra Table.
    **/
-  def repartitionByCassandraReplica(keyspaceName: String, tableName: String, partitionsPerHost: Int = 10)
+  def repartitionByCassandraReplica(keyspaceName: String,
```
jacek-lewandowski: codestyle

jacek-lewandowski:
Looks good, please reformat according to style guide

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
RussellSpitzer: This matches the style of the other functions in the file I think. Would you like me to switch the whole file to
```scala
def foo(
  arg1
  arg2
):type = {

}
```

jacek-lewandowski: Now i'm confused :(
```
def foo(
  arg1,
  arg2): type = {

code
}
```

or the way you mentioned. I don't really know...

RussellSpitzer:
Although we've had some disagreement on style this is what is listed in http://docs.scala-lang.org/style/declarations.html
```scala
class Person(
    name: String,
    age: Int,
    birthdate: Date,
    astrologicalSign: String,
    shoeSize: Int,
    favoriteColor: java.awt.Color) 
  extends Entity
  with Logging
  with Identifiable
  with Serializable {
}
```
So I'll move to match that?

jacek-lewandowski:
+1

PR #697, opened 2015-06-01
SPARKC-177: Refactored ColumnSelector

jacek-lewandowski:
This patch removes `select` method from `TableDef` because this method turned out to be redundant
and making a circular dependency to `ColumnSelector`. Instead of pattern matching in this method,
the proper column sequence is obtained directly in `ColumnSelector.selectFrom` method.

The exceptions have been aligned and the test cases have been moved to `ColumnSelectorSpec`.

 The case when `TableDef.select` method gets `SomeColumns` where never used but in the tests.


pkolaczk:
+1

PR #699, opened 2015-06-01
SPARKC-163 Replace CassandraRelation by CassandraSourceRelation

alexliu68:


RussellSpitzer:
Needs Changes.txt update and a few rename considerations, I'll take a look again after that gets cleaned up

PR #704, opened 2015-06-08
SPARKC-183: ClosestLiveHost should ignore localhost if down

RussellSpitzer:
Previously we could run into a situation where the node the driver was
running on could be down and the application would not run even if a
different node was specified as the intial contact point for the
cluster. This was caused because closestLiveHost was not actually
checking liveness since it relied on sortNodesByProximityAndStatus which
would always place localhost first. This means a dead localhost would be
used before a live neighbor on the driver. This code filters out dead
nodes from the nodes returned by SortNodesByProximityAndStatus
effectively removing the chance of selecting a node which is not
actually up.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnector.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnector.scala
@@ -122,8 +122,9 @@ class CassandraConnector(conf: CassandraConnectorConf)
     withClusterDo { cluster =>
       LocalNodeFirstLoadBalancingPolicy
         .sortNodesByProximityAndStatus(_config.hosts, cluster.getMetadata.getAllHosts.toSet)
+        .filter(_.isUp) //Remove localhost if it is down (SPARKC-183)
         .headOption
-        .getOrElse(throw new IOException("Cannot connect to Cassandra: No hosts found"))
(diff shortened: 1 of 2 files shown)


jacek-lewandowski:
I am wondering - maybe we should change the method `LocalNodeFirstLoadBalancingPolicy.sortNodesByProximityAndStatus` to something like `sortByStatusAndProximity` so that the node status is the major property we sort over the nodes? This would be more generic solution, WDYT?

RussellSpitzer:
That was actually my first solution but I worried about changing the behavior of the strategy. This change will actually make closestLiveHost match it's documentation so I chose that as the path of least change. Otherwise we change the documented contract of the sort function so we'll have to double check for other strange status effects. If you want me to apply the change to the LoadBalancing policy instead still let me know or we can talk about it tomorrow :)

jacek-lewandowski:
Fair enough, let's consider the change in LBP in a separate ticket
+1

pkolaczk:
LGTM. I agree with Jacek on that, I think that LBP should also not connect to localhost if it is dead. So sorting by status first and then by proximity might be a more general solution.

RussellSpitzer:
Well if both of you agree (and since that is what I did first) I can always change the approach :)

RussellSpitzer:
@jacek-lewandowski @pkolaczk Made the modification to LNFBP as well.

PR #705, opened 2015-06-09
SPARKC-161: RDD coalesce() pushdown to cassandra. That simplify RDD p…

artem-aliev:
…artitions calculations and reduce number of steps.


bcantoni:
Superseded by #985

PR #707, opened 2015-06-09
SPARKC-186: Table and Keyspace Name suggestions in DataFrames API

RussellSpitzer:
Previously keyspace/table not found messages were displayed as "key x
not found". With this patch we add in the same suggestion functionality
for the DataFrames API that exists with the normal RDD code. We attempt
to retreive keyspace and table metadata from C* and if it fails we
fall back on our suggestion utility.


alexliu68:
+1

RussellSpitzer:
retest this please

PR #708, opened 2015-06-10
Sparkc 183 b1.0

RussellSpitzer:


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnector.scala
```diff
@@ -118,9 +118,10 @@ class CassandraConnector(conf: CassandraConnectorConf)
   def closestLiveHost: Host = {
     withClusterDo { cluster =>
       LocalNodeFirstLoadBalancingPolicy
-        .sortNodesByProximityAndStatus(_config.hosts, cluster.getMetadata.getAllHosts.toSet)
+        .sortNodesByStatusAndProximity(_config.hosts, cluster.getMetadata.getAllHosts.toSet)
+        .filter(_.isUp)
```
jacek-lewandowski: Isn't this filter redundant now?

RussellSpitzer: The down nodes are still in the list and this claims that it will find the closest Live host, so if all the nodes are down it will still return a value when I believe it should not.

jacek-lewandowski:
+1

PR #712, opened 2015-06-12
SPARKC-194: Remove underlines from Cassandra DataSource options.

pkolaczk:
Renamed c_table to table.
Renamed push_down to pushdown.
The reason for this is that, probably due to a bug, Spark 1.4.0 does not accept underlines in data source options.

### DIFF
--- a/doc/14_data_frames.md
+++ b/doc/14_data_frames.md
@@ -18,10 +18,10 @@ Those followed with a default of N/A are required, all others are optional.
 
 | Option Key  | Controls                                              | Values        | Default  |
 |-------------|-------------------------------------------------------|---------------|----------|
-| c_table     | The Cassandra Table to connect to                     | String        | N/A      |
-| keyspace    | The Keyspace where c_table is looked for              | String        | N/A      |
+| table       | The Cassandra Table to connect to                     | String        | N/A      |
+| keyspace    | The Keyspace where table is looked for                | String        | N/A      |
 | cluster     | The group of the Cluster Level Settings to inherit    | String        | "default"|
-| push_down   | Enables pushing down predicates to C* when applicable | (true,false)  | true     |
+| pushdown    | Enables pushing down predicates to C* when applicable | (true,false)  | true     |
 
 ####Read, Writing and CassandraConnector Options
 Any normal Spark Connector configuration options for Connecting, Reading or Writing
@@ -47,18 +47,18 @@ val conf = new SparkConf()
 
 val df = sqlContext.load(
   "org.apache.spark.sql.cassandra", 
-   options = Map( "c_table" -> "words", "keyspace" -> "test" 
+   options = Map( "table" -> "words", "keyspace" -> "test" 
 )// This DataFrame will use a spark.cassandra.input.size of 5000
 
 val otherdf =  sqlContext.load(
   "org.apache.spark.sql.cassandra", 
-   options = Map( "c_table" -> "words", "keyspace" -> "test" , "cluster" -> "ClusterOne" )
+   options = Map( "table" -> "words", "keyspace" -> "test" , "cluster" -> "ClusterOne" )
 )// This DataFrame will use a spark.cassandra.input.size of 1000
 
 val lastdf = sqlContext.load(
                "org.apache.spark.sql.cassandra", 
                 options = Map( 
-                  "c_table" -> "words", 
+                  "table" -> "words", 
                   "keyspace" -> "test" ,
                   "cluster" -> "ClusterOne",
                   "spark.cassandra.input.split.size" -> 500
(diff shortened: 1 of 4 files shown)


spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataSourceSpec.scala
```diff
@@ -52,14 +52,9 @@ class CassandraDataSourceSpec extends SparkCassandraITFlatSpecBase {
         |CREATE TEMPORARY TABLE $tmpTable
         |USING org.apache.spark.sql.cassandra
         |OPTIONS (
-        | c_table "$table",
+        | table "$table",
         | keyspace "$keyspace",
-        | push_down "$pushDown",
-        | spark_cassandra_input_page_row_size "10",
```
alexliu68: We need keep those settings here, change it to spark.cassandra.input.page.row.size to see whether it works or not

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/DefaultSource.scala
```diff
@@ -23,10 +23,10 @@ import DefaultSource._
...
  *      USING org.apache.spark.sql.cassandra
  *      OPTIONS (
- *       c_table "table",
+ *       table "table",
  *       keyspace "keyspace",
  *       cluster "test_cluster",
- *       push_down "true",
+ *       pushdown "true",
  *       spark_cassandra_input_page_row_size "10",
  *       spark_cassandra_output_consistency_level "ONE",
```
alexliu68: We need test those settings

doc/14_data_frames.md
```diff
@@ -101,10 +101,10 @@ scala> sqlContext.sql(
...
      |USING org.apache.spark.sql.cassandra 
      |OPTIONS ( 
-     |  c_table "words",
+     |  table "words",
      |  keyspace "test", 
      |  cluster "Test Cluster", 
-     |  push_down "true", 
+     |  pushdown "true", 
      |  spark_cassandra_input_page_row_size "10", 
      |  spark_cassandra_output_consistency_level "ONE",
```
alexliu68: Check those settings work or not

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataSourceSpec.scala
pkolaczk: Huh, so the brutal truth is that they *don't work* now and we can't do anything about it. At least not with Spark 1.4.0. Instead of adding support for "." they removed support for "_" and now we have no word separator at all.

pkolaczk: I opened a ticket in DataBricks support. I hope they soon release 1.4.1.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/DefaultSource.scala
pkolaczk: I agree. Keyspace/table/cluster/pushdown should work now, others will not. But there were no tests for these settings before, so maybe let's address it with a separate ticket we'd do after we hear back from DataBricks.

doc/14_data_frames.md
```diff
@@ -101,10 +101,10 @@ scala> sqlContext.sql(
    """CREATE TEMPORARY TABLE words 
      |USING org.apache.spark.sql.cassandra 
      |OPTIONS ( 
-     |  c_table "words",
+     |  table "words",
      |  keyspace "test", 
      |  cluster "Test Cluster", 
-     |  push_down "true", 
+     |  pushdown "true", 
      |  spark_cassandra_input_page_row_size "10",
```
RussellSpitzer: This Is invalid now as well

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/DefaultSource.scala
alexliu68: It is tested before, but now you removed them. We should at least add big WARNING to document if we don't fix it and suggest users to use 1.3.1 if they need set table level settings

pkolaczk: It wasn't really tested, because I removed the settings, yet the tests pass. So the tests definitely were *not* verifying these settings being in effect.

PR #713, opened 2015-06-12
SPARKC-195: Fix CassandraSQLRow component data types

pkolaczk:
Spark 1.4.0 expects UTF8String instead of String.
It also does not expect Sets, so Sets are converted to Seqs.


alexliu68:
+1, but the jenkins tests failed

pkolaczk:
jenkins test fail, because this ticket does not fix all the problems. Other problems in tests are addressed by separate subtickets of SPARKC-192.

PR #714, opened 2015-06-12
SPARKC-193: Upgrade to Spark 1.4.0: fix SparkEnv mock

pkolaczk:
Replace SparkEnvMock with a proper mock created by mockito.
SparkEnvMock caused NPE because it passed null RpcEnv.


jacek-lewandowski:
nice, +1

PR #719, opened 2015-06-16
SPARCK 190 - Save complex objects with UDTs to Cassandra

pkolaczk:


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/TupleValue.scala
```diff
@@ -20,10 +20,11 @@ object TupleValue {
     new TupleValue(values: _*)
   }
 
-  val TupleValueTypeTag = typeTag[TupleValue]
+  val TypeTag = typeTag[TupleValue]
+  val Symbol = typeOf[TupleValue].asInstanceOf[TypeRef].sym
```
helena: Do we need this code replicated in japi as well?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
```diff
@@ -69,6 +69,13 @@ trait StructDef extends Serializable {
   /** Returns the columns that are not present in the structure. */
   def missingColumns(columnsToCheck: Seq[ColumnRef]): Seq[ColumnRef] =
     for (c <- columnsToCheck if !columnByName.contains(c.columnName)) yield c
+  
+  /** Type of the data described by this struct */
+  type ValueRepr <: AnyRef
+  
+  /** Creates new instance of this struct.
+    * Column values must be given in the same order as columnNames */
+  def newInstance(columnValues: Any*): ValueRepr
```
helena: these sort of creates always feel like they should be apply(..): SomeType

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
```diff
@@ -148,6 +155,12 @@ case class TableDef(
        |  PRIMARY KEY ($primaryKeyClause)
        |)""".stripMargin
   }
+
+  type ValueRepr = CassandraRow
+  
+  def newInstance(columnValues: Any*): CassandraRow = {
```
helena: ```def apply(columnValues: Any*): CassandraRow = { ..}```

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TupleType.scala
```diff
@@ -34,38 +36,53 @@ case class TupleType(componentTypes: TupleFieldDef*)
...
+      (componentConverters: IndexedSeq[TypeConverter[_ <: AnyRef]])
+      (componentValues: Any*): TupleValue = {
     require(
       componentValues.length == columns.length,
       s"Expected ${columns.length} components, instead of ${componentValues.length}")
     val values =
       for (i <- columns.indices) yield
-        columnTypes(i).converterToCassandra.convert(componentValues(i))
+        componentConverters(i).convert(componentValues(i))
     new TupleValue(values: _*)
```
helena: we're creating a new TupleValue, it seems this belongs in object TupleValue with an apply function.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ReflectionUtil.scala
```diff
@@ -85,13 +85,31 @@ object ReflectionUtil {
     getters(implicitly[TypeTag[T]].tpe)
   }
 
-  /** Returns the type of the parameters of the given method.
+  /** Looks up a method by name in a type.
     * The method must exist, otherwise `IllegalArgumentException` is thrown. */
-  def methodParamTypes(tpe: Type, methodName: String): Seq[Type] = TypeTag.synchronized {
+  def method(tpe: Type, methodName: String): Type = TypeTag.synchronized {
     val member = tpe.member(newTermName(methodName))
```
helena: I don't think ReflectionUtil.scala is private[connector] so consider things like this, where the function is public to the user API but there are no checks on the args passed in, should have a require, for example require(methodName != null || methodName.nonEmpty, "'methodName' must have a valid value.")

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/MappedToGettableDataConverter.scala
```diff
@@ -0,0 +1,187 @@
...
+          new JavaBeanColumnMapper[U]()(ReflectionUtil.classTag[U])
+        else
+          new DefaultColumnMapper[U]
+      }
+
+      /** For these types we don't want to use column mappers */
+      private val dynamicStructRepresentations = Set(
+        TupleValue.Symbol.fullName,
+        UDTValue.Symbol.fullName,
+        "com.datastax.spark.connector.japi.TupleValue",
```
helena: Could be brittle if changes occur in package structure for japi. Will compiler catch that if that happens?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/MappedToGettableDataConverter.scala
```diff
@@ -0,0 +1,187 @@
...
+        dynamicStructRepresentations.contains(sym.fullName)
+
+      /** Returns true for tuple types that provide full type information of their components */
+      private def isTypedTuple(sym: Symbol): Boolean =
+        ReflectionUtil.isScalaTuple(sym) ||
+          sym == Symbols.PairSymbol ||
+          sym == Symbols.TripleSymbol
+
+      /** Returns a converter for converting given column to appropriate type savable to Cassandra */
+      private def converter[U : TypeTag](columnType: ColumnType[_]): TypeConverter[_ <: AnyRef] = {
```
helena: this is a huge function

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
pkolaczk: it can't be apply, because it is not a companion object and it doesn't create an instance of StructDef.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
pkolaczk: Can't be on CassandraRow, because columnNames information is missing. And there is already an `apply` on CassandraRow taking the columnNames and columnValues.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TupleType.scala
pkolaczk: You can't create a TupleValue *with* conversion, without knowing the types of the columns in the tuple. That's why TupleType serves as a factory here.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ReflectionUtil.scala
pkolaczk: ok

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/MappedToGettableDataConverter.scala
pkolaczk: Yeah, I don't like this. The problem is, japi is an optional dependency. It is not on the classpath when compiling. I could try to annotate somehow those classes in some way that would be visible to reflection (a marker trait?) and discriminate basing on the annotation, instead a static list of classnames. Wdyt?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/MappedToGettableDataConverter.scala
pkolaczk: Yeah, big, but fortunately quite flat. This really belongs to the TypeConverter module, but this would need a bigger refactoring. Can we unify this later?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/TupleValue.scala
pkolaczk: Hmm, no. Good catch.  I added these in hope I can access them from the core, but then it turned out japi is not on the classpath. I forgot to remove them.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
helena: That is what I thought but I had to check :)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/MappedToGettableDataConverter.scala
helena: I like marker traits.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/MappedToGettableDataConverter.scala
helena: +1 I just had to point out the obvious :)

pkolaczk:
Error: Invalid or corrupt jarfile sbt/sbt-launch-0.13.8.jar 
huh? jnkns plz

pkolaczk:
retest this please

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
pkolaczk: Turned out it was not needed, because they already have a good common trait: `GettableByIndexData`.

pkolaczk:
Applied comments, rebased and squashed.

pkolaczk:
retest this please

PR #722, opened 2015-06-19
SPAKC 150 - Remove semantic API ambiguity of selecting RDDs of KV pairs

pkolaczk:


helena:
+1 great new addition :)

PR #723, opened 2015-06-19
SPARKC 150 - Remove semantic API ambiguity of selecting RDDs of KV pairs

pkolaczk:


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/KeyValueRowReader.scala
```diff
@@ -1,24 +1,18 @@
 package com.datastax.spark.connector.rdd.reader
 
 import com.datastax.driver.core.{ProtocolVersion, Row}
-import com.datastax.spark.connector.ColumnRef
+import com.datastax.spark.connector.{ColumnSelector, ColumnRef}
 import com.datastax.spark.connector.cql.TableDef
-import com.datastax.spark.connector.mapper.TupleColumnMapper
 
-class KeyValueRowReaderFactory[K, V](keyRRF: RowReaderFactory[K], valueRRF: RowReaderFactory[V])
+class KeyValueRowReaderFactory[K, V](
```
helena: I wonder about things like there, where it is not a private[connector] and thus a public API and we modify the API over time. I like the mods but wonder if it needs to be public?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/RowReaderFactory.scala
```diff
@@ -29,25 +29,24 @@ trait ThisRowReaderAsFactory[T] extends RowReaderFactory[T] {
...
   trait IsSingleColumnType[T]
-  implicit def isSingleColumnType[T](implicit ev1: TypeConverter[T], ev2: T IsNotSubclassOf (_, _)): IsSingleColumnType[T] = null
-
-  implicit def classBasedRowReaderFactory[R <: Serializable]
-      (implicit tt: TypeTag[R], cm: ColumnMapper[R], ev: R IsNotSubclassOf (_, _), ev2: R DoesntHaveImplicit IsSingleColumnType[R]): RowReaderFactory[R]  =
+  implicit def isSingleColumnType[T](
+    implicit
+      ev1: TypeConverter[T],
+      ev2: T IsNotSubclassOf (_, _),
+      ev3: T IsNotSubclassOf (_, _, _)): IsSingleColumnType[T] = null
```
helena: Null as a necessary evil? Is it required or is Option at all possible?

helena:
@pkolaczk except for a few minor questions, this looks good

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/KeyValueRowReader.scala
pkolaczk: Yeah, we can make it private now.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/RowReaderFactory.scala
pkolaczk: This is funny because, this result is never actually used - but the implicit might be called, so throwng an exception is not a solution. I'd like to know if we can use something else here instead of null, but nothing comes to my mind.

helena:
After chatting with @pkolaczk - LGTM

PR #731, opened 2015-06-23
SPARKC-198: Meaningful exception if some partition key column is null

jacek-lewandowski:


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RoutingKeyGenerator.scala
```diff
@@ -44,8 +44,12 @@ class RoutingKeyGenerator(table: TableDef, columnNames: Seq[String])
 
   private def fillRoutingKey(stmt: BoundStatement): Array[ByteBuffer] = {
     val rk = routingKey.get
-    for (i <- 0 until partitionKeyIdxs.size)
+    for (i ← partitionKeyIdxs.indices) {
+      if (stmt.isNull(partitionKeyIdxs(i)))
+        throw new NullPointerException(
```
helena: IMO never throw an NPE, in fact, why do you do _any_ work prior to checking the argument to the function?

        private def fillRoutingKey(stmt: BoundStatement): Array[ByteBuffer] = {
           require(stmt != null, "BoundStatement 'stmt' must have a valid value.") 
        }

then in your specs check for IllegalArgumentException, because really you are trying to alert the engineer in runtime that you can't pass nulls to this api function. If this is called by our api then we'd better be checking the arg state!

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/RDDSpec.scala
```diff
@@ -201,6 +201,38 @@ class RDDSpec extends SparkCassandraITFlatSpecBase {
     checkArrayCassandraRow(result)
   }
 
+  it should "throw a meaningful exception if partition column is null when joining with Cassandra table" in {
+    val source = sc.parallelize(keys).map(x ⇒ new KVWithOptionRow(None))
+    try {
+      source.joinWithCassandraTable[(Int, Long, String)](ks, tableName).collect()
+      fail()
```
helena: I'm not clear why you are manually adding fail() shouldn't you be testing args that are null/None and the changed code you wrote be throwing the failure to note the args were not valid? In which case you should be doing

intercept[IllegalArgumentException] {
  source.doThingsThatShouldFail()
}

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RoutingKeyGenerator.scala
```diff
@@ -7,7 +7,7 @@ import com.datastax.spark.connector.cql.TableDef
 
 /** This class computes the routing key of a bound statement. */
 class RoutingKeyGenerator(table: TableDef, columnNames: Seq[String])
-  extends ((BoundStatement) => ByteBuffer) {
+  extends ((BoundStatement) ⇒ ByteBuffer) {
```
helena: Did you add scala formatting? Unclear why you added ⇒ vs =>, ⇒ is what formatting adds

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/RDDSpec.scala
jacek-lewandowski: As you can see, I need to verify the exception message - that is why I did it in this way

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RoutingKeyGenerator.scala
jacek-lewandowski: Actually this is NPE because the argument is null. I'm not sure about IAE because I think this should be used for method arguments.
Also we cannot identify them sooner because they are in RDD, so in the distributed data. They are not processed in the driver.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RoutingKeyGenerator.scala
jacek-lewandowski: IntelliJ does this automatically if you type => + space

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RoutingKeyGenerator.scala
helena: I don't understand why you have to throw a NPE, why not require like my sample, I get that it's in runtime while parsing partitionKeyIdxs

helena: And if you need to validate the message in tests, why not create a custom exception and test the exception type vs strings

jacek-lewandowski: I can change to IAE. However, how introducing a custom exception would allow me to verify the exception message which should say which column is null?

helena: What about:

         throw new NullOrEmptyColumnException(s"Invalid null value for partition key part $theValue")

         @SerialVersionUID(1L)
          class NullOrEmptyColumnException(columnName: String, message: String, cause: Throwable) extends             RuntimeException(message, cause) {

               def this(columnName: String) = this(columnName,"", null) 
           }

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RoutingKeyGenerator.scala
helena: Ug, I've never seen that, not good that it does that.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RoutingKeyGenerator.scala
helena: Then we can do this and check type vs string

       intercept[InvalidColumnException] {
           source.doThingsThatShouldFailOrThisTestShouldFail()
       }

WDYT?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RoutingKeyGenerator.scala
jacek-lewandowski: Why? don't you like that nice arrows? :)

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/RDDSpec.scala
```diff
@@ -201,6 +201,38 @@ class RDDSpec extends SparkCassandraITFlatSpecBase {
...
+    } catch {
+      case ex: Exception ⇒
+        ex.getMessage should include("Invalid null value for partition key part key")
```
pkolaczk: This way you're not testing if the exception has been thrown at all. Better use the pattern recommended by scalatest: 
val e = the [Exception] thrown by { some code }
e.getMessage should ...

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RoutingKeyGenerator.scala
pkolaczk: I think this is something we must agree upon. So far, my IntelliJ does not do that, and we have nice arrows in half of the places and ugly arrows in another half.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/RoutingKeyGenerator.scala
pkolaczk: This is a good idea with a custom exception class.

jacek-lewandowski: I don't like, but I don't want to fight for not creating it. Basically because it gives nothing but another class. The user is unable to catch it because it is thrown in executor. The driver will only get `SparkException` with the message but without the cause. So such class is useless.

helena:
LGTM thanks

jacek-lewandowski:
test this please

PR #735, opened 2015-06-24
SPARKC-206: Tuples support in Java API

jacek-lewandowski:
Added methods to CassandraJavaUtil which allow to create RowReaderFactory and RowWriterFactory for tuples.


pkolaczk:
+1

PR #738, opened 2015-06-24
SPARKC-43: reimplement token ring group algorithm. It will try to inc…

artem-aliev:
…rease number of replicas per RDD partition
This algorithm should probably be merged in master only


bcantoni:
Superseded by #746

PR #742, opened 2015-06-25
SPARKC-208: Fix setting split size in MB

pkolaczk:
The parameter spark.cassandra.input.split_size_in_mb now sets the
size in megabytes, not bytes.


jacek-lewandowski:
+1

PR #746, opened 2015-06-26
SPARKC-43: add partition sorting,to move partitions with more replica…

artem-aliev:
…s to the end of the list


pkolaczk:
+1

artem-aliev:
retest this please

PR #748, opened 2015-06-26
SPARKC-129 Add custom UUIDType and InetAddressType

alexliu68:
Add custom UUIDType and InetAddressType

UUIDType uses JVM data type UUID as base
InetAddressType uses JVM data type InetAddress as base

Since there is no way to convert a String to
UUID/InetAddress from a SQL query. The correct query
should be like

  select * from table
  where CAST(inet_address_column as string)
        = '/74.125.239.135'

It errors out for the query

  select * from table
  where inet_address_column = '/74.125.239.135'

Because that Spark internally CAST string
'/74.125.239.135' to a Double.

Same should apply to UUID as well


pkolaczk:
+1, but build is still failing

alexliu68:
retest it please

PR #754, opened 2015-06-29
SPARKC-163 Replace CassandraRelation by CassandraSourceRelation

alexliu68:
Since datasource API is the major integration API,
CassandraRelation is built on legacy API, it can
be replaced by CassandraSourceRelation.

Once replacing is done, there is only one unified
API code to support.

Once replace CassandraRelation by CassandraSourceRelation,
CassandraTableScan, InsertIntoTable, CassandraStrategies
classes are no longer used, so remove them.

Add a local cache to cache the previously
retrieved table logical plan.

Implements register/ungister/refresh/tables/
tableExists methods.

Removed OverrideCatalog from CassandraCatalog

Merge tables from Cassandra and local cache when list tables/exist


pkolaczk:
+1

PR #755, opened 2015-06-30
SPARKC-164 Retrieve TableSize from C* system table for datasource rel…

alexliu68:
…ation


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/DataSizeEstimates.scala
```diff
@@ -78,11 +78,15 @@ class DataSizeEstimates[V, T <: Token[V]](
 
   /** Estimates the total amount of data in a table assuming no replication. */
   lazy val dataSizeInBytes: Long = {
-    val byteCount = tokenRanges.map(_.totalSizeInBytes).sum
-    val normalizedCount = (byteCount / ringFraction).toLong
+    val normalizedCount = (totalDataSizeInBytes / ringFraction).toLong
```
RussellSpitzer: I feel like we should rename this function tokenRangeSizeInBytes

RussellSpitzer:
I want to rename one function slightly outside of this patch, after that i'll be +1 :)

alexliu68:
close this pull by https://github.com/datastax/spark-cassandra-connector/pull/865

PR #757, opened 2015-07-02
Update Java Driver to 2.2.0-rc2

JeremiahDJordan:


pkolaczk:
Some tests related to TTL are failing. I also tried to compile against 2.2.0-rc2, but it fails to compile.

JeremiahDJordan:
@pkolaczk updated to 2.2.0rc2 which looks like it fixed the TTL stuff.  Everything seems to be looking good now besides nested tuples/udt which are broken until https://datastax-oss.atlassian.net/browse/JAVA-847 is fixed.

RussellSpitzer:
Currently on 3.0 Driver

PR #758, opened 2015-07-03
Remove white spaces in c* connection host string

noorul:
Given comma separated list and if there is space after comma then name resolution fails.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectorConf.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectorConf.scala
@@ -105,11 +105,10 @@ object CassandraConnectorConf extends Logging {
 
   def apply(conf: SparkConf): CassandraConnectorConf = {
     ConfigCheck.checkConfig(conf)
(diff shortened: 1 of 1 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectorConf.scala
```diff
@@ -105,7 +105,8 @@ object CassandraConnectorConf extends Logging {
 
   def apply(conf: SparkConf): CassandraConnectorConf = {
     ConfigCheck.checkConfig(conf)
-    val hostsStr = conf.get(CassandraConnectionHostProperty, InetAddress.getLocalHost.getHostAddress)
+    val hostsStr = "\\s+".r.replaceAllIn(conf.get(CassandraConnectionHostProperty,
+                                                  InetAddress.getLocalHost.getHostAddress), "")
```
ffarozan: Better way would be to use a hostName.trim in the for loop below.

noorul:
How do I re-trigger this check?

pkolaczk:
Squashed and merged into b1.3, b1.4 and master. Thanks.

PR #759, opened 2015-07-20
SPARKC-215: Cassandra native count is performed by `cassandraCount` method

jacek-lewandowski:
Previously native Cassandra count overrode `count` method from RDD. Because of
some request for more control when to execute native count and when not to, we
renamed Cassandra native count to `cassandraCount`. Now `count` is pure Spark
count.


pkolaczk:
+1

PR #762, opened 2015-07-20
Change example property used in DataFrame docs to split size in MB

chbatey:
It was using split size which as far as I can tell doesn't exist any more which may confuse users.

If it does exist and I missed it then ignore this and call me a fool :)


RussellSpitzer:
Good catch! I think this has been fixed. Sorry it took so long to get to this pr!

PR #780, opened 2015-07-31
SPARKC-191 Get columns from dataframe when save to table

alexliu68:


spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataSourceSpec.scala
```diff
@@ -149,4 +149,29 @@ class CassandraDataSourceSpec extends SparkCassandraITFlatSpecBase with Logging
       " FROM tmpTable) AS tmpTable1 WHERE  a1=1 and b1=2 and c1=1 and e1=1 ").collect() should have length 2
   }
 
+  it should "allow to save a DF to a Cassandra table" in {
```
RussellSpitzer: I think we should just change this title to should be able to save DF with reversed order columns, and add one more test for partial columns (which i believe you have fixed with this patch as well)

alexliu68: Sounds good.

pkolaczk:
+1 Looks good.

RussellSpitzer:
+1

PR #781, opened 2015-07-31
SPARKC-191 Get columns from dataframe when save to table

alexliu68:


spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -67,7 +68,8 @@ private[cassandra] class CassandraSourceRelation(
     }
 
     implicit val rwf = SqlRowWriter.Factory
-    data.rdd.saveToCassandra(tableRef.keyspace, tableRef.table, AllColumns, writeConf)
+    val columns = SomeColumns(data.columns.map(x => x: ColumnRef): _*)
```
RussellSpitzer: I think this is a really clean solution. +1

PR #782, opened 2015-08-04
SPARKC-226: Change output.TP_mb_per_sec -> double

RussellSpitzer:
We found that having only a whole number granularity made it difficult
to actually tune the output to get maximum throuhgput. This ticket
allows the conf to be set to a double rather than an Int providing the
end user with a much greater ability to control thorughput while
maintaining backwards compatibility.


pkolaczk:
+1, looks good

bcantoni:
Can you also update the doc pages that mention throughput_mb_per_sec to explain it's a float now? I think it's FAQ.md and 5_saving.md.

PR #783, opened 2015-08-07
SPARKC-227: Fix Connection Caching, Change SSL EnabledAlgorithms to Set

RussellSpitzer:
The bug here is that an Array(enabledAlgorithms) will not match another
Array even if their contents match exactly. This will happen even if the
array is declared statically and finally. To fix this we need to change
the enabledAlgorithms variable to a Set which is more appropriate for
this field as well (Since order and duplicates do not matter). The
original decision to make this an Array was most likely done to match
the Java Driver's implementation, unfortunately this has some bad
implications for the underlying connection pool which we can avoid with
a Set.

Since the Hash of the CassandraConnectionConf is used by
the Cache to determine whether or not a new Cluster object needs to be
made, using an Array forces the creation of a new Cluster object on
every write. These cannot be reused. Changing this to a set allows the
cache to recognize that two CassandraConnectionConfs which contain SSL
EnabledAlgorithms which have the same contents are the same.


pkolaczk:
+1

PR #785, opened 2015-08-08
updated documentation to save others time doing what might be a common task

sheldonkhall:
I had to create a cassandra table with a more complicated primary key structure than shown in the documentation examples. I finally figured out how to do this from the source code and thought I could save some others time by sharing what I have done.

### DIFF
--- a/doc/5_saving.md
+++ b/doc/5_saving.md
@@ -208,6 +208,28 @@ val collection = sc.parallelize(Seq(WordCount("dog", 50), WordCount("cow", 60)))
 collection.saveAsCassandraTableEx(table2, SomeColumns("word", "count"))
 ```
 
+To create a table with a custom definition, and define which columns are to be partition and clustering column keys:
+
+```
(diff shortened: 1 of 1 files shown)


jacek-lewandowski:
Thanks for the patch.

Could you create a ticket in Jira, squash your commits, and prepend the result commit title with "SPARKC-xxx: "

sheldonkhall:
OK I think I have done this now. I am new to the system so let me know if I need to make any further changes.

https://datastax-oss.atlassian.net/projects/SPARKC/issues/SPARKC-232?filter=allopenissues

RussellSpitzer:
I think Jacek was asking if you could present a single commit rather than 5, you should be able to do this using the get rebase command. I prefer using git rebase -i target_branch 
Then you'll see you can change the commit labels from keep to squash

See 
http://jaimeiniesta.github.io/learn.github.com/p/rebasing.html
For more info

When you have done this correctly you'll have to force push back up to your branch and this PR will end up only having a single commit in it 

Thanks so much for helping out!

sheldonkhall:
I think that I have done this correctly now, thanks for the help!

RussellSpitzer:
Yep that's it! Thanks for the update!

PR #794, opened 2015-08-14
SPARKC-234 Added C* schema cache refresh to CassandraSQLContext

alexliu68:


bcantoni:
Let's also add something in the docs explaining how and when to use this.

alexliu68:
Added docs to explain when to use it

RussellSpitzer:
I think this is a good idea and am +1 but is there a way we can make this possible from within the SQL Dialect as well?

alexliu68:
please test it

PR #798, opened 2015-09-01
SPARKC-239: Fix sbt wrapper script download of sbt-launch jar

bcantoni:
- reorder URLs so new working one is first
- update curl command to fail on 404 error (-f) and follow redirects (-L)


bcantoni:
Updated based on @RussellSpitzer suggestion to just have the one download URL since we know the other one is no longer valid.

RussellSpitzer:
+1

PR #807, opened 2015-09-04
Sparkc 202: Fix Repartition and JWC and Streaming  Checkpointing

RussellSpitzer:
In addition I removed the old actor tests since they were a bit spotty and replaced them with some RDDQueue Streams.


RussellSpitzer:
Ignore the first few Test Commits because they are overridden by the later test commits

jacek-lewandowski:
It's good that you found the way to fix that with so little changes. I have some general concerns (they are very minor though) - naming - it appears that the user should provide RowWriter for reading Cassandra table. This is a little bit misleading. Also we have two more parameters in the methods which already take so many.

RussellSpitzer:
@jacek-lewandowski I also feel a bit nervous about the amount of variables on the JoinWithCassandra Constructor but since it is package private it should only be annoying to us for now. In the long run I think we need to further decouple the construction of rowWriters and readers from the creation of any of our RDDs.

spark-cassandra-connector-java/src/main/java/com/datastax/spark/connector/japi/RDDJavaFunctions.java
```diff
@@ -114,6 +116,8 @@ public void saveToCassandra(
                 limit,
                 clusteringOrder,
                 readConf,
+                Option.<RowReader<R>>empty(),
+                Option.<RowWriter<T>>empty(),
```
pkolaczk: in the future we need to fix these extremely long parameter lists ;)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -47,6 +49,11 @@ class CassandraJoinRDD[L, R] private[connector](
 
   override protected val classTag = rightClassTag
 
+  override protected[connector] lazy val rowReader: RowReader[R] = manualRowReader match {
+    case Some(rowReader) => rowReader
```
pkolaczk: suspicious shadowing

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -34,7 +34,9 @@ class CassandraJoinRDD[L, R] private[connector](
     val where: CqlWhereClause = CqlWhereClause.empty,
     val limit: Option[Long] = None,
     val clusteringOrder: Option[ClusteringOrder] = None,
-    val readConf: ReadConf = ReadConf())(
+    val readConf: ReadConf = ReadConf(),
+    manualRowReader: Option[RowReader[R]] = None,
+    manualRowWriter: Option[RowWriter[L]] = None)(
```
pkolaczk: This constructor is private, so it is not exposed directly, so have you considered:
 - removing implicit RowReaderFactory / RowWriterFactory from here at all,
 - making rowReader / rowWriter obligatory
 - evaluating rowReader / rowWriter at the call site, right before newing this class?

pkolaczk: I can see two usages:
- https://github.com/datastax/spark-cassandra-connector/blob/SPARKC-202/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala#L142-L142
- https://github.com/datastax/spark-cassandra-connector/blob/SPARKC-202/spark-cassandra-connector-java/src/main/java/com/datastax/spark/connector/japi/RDDJavaFunctions.java#L108-L108

Only one of these usages passes a rowWriter/rowReader and another one passes a factory. Any reason why the Java one needed this refactoring while the Scala one is ok without it? Or is it an omission? 

Anyway, I'd pass a rowWriter/rowReader to any RDD implementations, as it is much safer for its serializability (no need to do @transient). Wdyt?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/streaming/DStreamFunctions.scala
```diff
@@ -48,8 +49,9 @@ class DStreamFunctions[T](dstream: DStream[T]) extends WritableToCassandra[T] wi
     currentType: ClassTag[T],
     rwf: RowWriterFactory[T]): DStream[T] = {
 
+    val converter = ReplicaMapper[T](connector, keyspaceName, tableName, partitionKeyMapper)
```
pkolaczk: why is this called a converter?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
```diff
@@ -159,14 +158,39 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
   implicit
     connector: CassandraConnector = CassandraConnector(sparkContext.getConf),
     currentType: ClassTag[T],
-    rwf: RowWriterFactory[T]) = {
+    rwf: RowWriterFactory[T]): CassandraPartitionedRDD[T] = {
+
+    val converter = ReplicaMapper[T](connector, keyspaceName, tableName, partitionKeyMapper)
```
pkolaczk: converter? What does it convert? The class name suggests it is a replica mapper (even then, still "mapper" is overloaded, for we also use mappers for mapping classes to cassandra schema).

pkolaczk:
Overall it looks good and I'm also happy it wasn't as big as I thought it would be. I was wondering if in the future we could make RRFs and RWFs Serialiable as well, so we never run into these problems again. Currently we must take extreme care that RRFs and RWFs are not accidentally passed to the inside of lambdas passed to Spark. 

The primary reason they are not Serializable now is that they rely on Scala reflection to get object-cassandra mapping, and Scala reflection primitives are (were? did Scala 2.11 fix this?) not Serializable. 

I can see two ways we fix that in the long run:
- Scala will fix it and in some new version reflection primitives are truly serializable
- we go the hard way and use macros (experimental feature) to generate RRs and RWs; macros would work even in lambdas because they would get expanded at compile time, so before serializing any lambda

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
RussellSpitzer: Yeah i'll try to rename this, this is the factory that uses the Java Driver to transform the Partition key into a set of Ip addresses. So maybe ReplicaLocator ?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
RussellSpitzer: On the usages: Only the Java constructor requires us to use the param because it specifies the later parameters and doesn't understand named parameters. The Scala code can remain the same since all the later parameters are passed by name. 

(On removing the RWF from this class completely)
I tried implementing the above approaches originally and all of them ran into some unfortunate issues. I think the real way to go in the future (and this will require a much larger refactoring) would be to remove all of the code from the RDD itself except for the compute method and partitioning methods. This would leave the join RDD taking parameters of a CassandraConnector, a CQL Statement, and a RowReader and RowWriter. All of the api would be exposed through a secondary class which could correctly build the row readers and writers based on the options passed to the  factory. We have to do this because the row writer / reader can change dependent on the various options we allow in the api.

So my proposal for a future api would be

CassandraJoinRDDFactory ( leftSide: RDD , where, on, .... clauses, Implicit RWF, RRF ) WITH CassandraRDDApi 
CassandraJoinRDD( leftSide: Rdd, selectStatement: String, rowWriter: RowWriter, rowReader: RowReader, Implicit CassandraConnector,)

Then we would have an implicit  CassandraJoinRDDFactory -> CassandraJoinRDD 
 which would instantiate a JoinRDD from the factories

For this we would have to be careful about naming as not to disturb an user code.

I hesitate from fully doing this because that would be a much larger change, but if you would prefer that I could do that as well.(I do think it is the right long term situation)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
RussellSpitzer: Agreed, but the other solution will be much more disruptive (do we have time for 1.4?)

RussellSpitzer:
So Scala 2.11 works with the code prior to the patch so that's really nice :)

I think the grand refactor of API - RDD is the real way to go forward and fix this in the future. We won't have to hit macros and we can decouple the relationship between the compute method of an RDD and the various api functions we allow.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
jacek-lewandowski: I think that @pkolaczk meant that you used the same name for lazy val and for unapply - you should just change 
``` case Some(rowReader) => rowReader ```
to something like
``` case Some(someRowReader) => someRowReader ```

**Comment by jacek-lewandowski** [comment]
I'm +1 for he grand refactor. Although I'm 90% sure we could implement type tag serialisation quite easily for Scala 2.10, the grand refactoring could be helpful when we want to implement some other features. I'm not sure for the target version of such change though.

File: spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala (hunk 2 above)
Comment (pkolaczk) [inline spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala]: yeah, just rename the pattern matching param - actually the param name can be shorter in this case, as its scope of visibility is very narrow

File: spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala (hunk 3 above)
Comment (pkolaczk) [inline spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala]: sure, let's leave it for the future - but I agree that the API needs a major cleanup now

File: spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala (hunk 5 above)
Comment (pkolaczk) [inline spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala]: +1 for ReplicaLocator

**Comment by pkolaczk** [comment]
+1 for grand refactoring in the future (1.6?). Our RDDs collected too many features over time.

**Comment by jacek-lewandowski** [comment]
Did you really mean 1.6? I though 1.5

**Comment by pkolaczk** [comment]
1.5 is going to be released too soon (isn't it?); and we want to release spark connector at the same time. So all major refactoring should be done for 1.6.

**Comment by jacek-lewandowski** [comment]
ah, ok, i forgot that

**Comment by RussellSpitzer** [comment]
@pkolaczk @jacek-lewandowski Changes made based on the review! Please let me know if you have any other comments.

File: spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala [hunk 6]
```diff
@@ -159,14 +158,39 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
   implicit
     connector: CassandraConnector = CassandraConnector(sparkContext.getConf),
     currentType: ClassTag[T],
-    rwf: RowWriterFactory[T]) = {
+    rwf: RowWriterFactory[T]): CassandraPartitionedRDD[T] = {
+
+    val converter = ReplicaLocator[T](connector, keyspaceName, tableName, partitionKeyMapper)
```
pkolaczk: You renamed ReplicaMapper to ReplicaLocator, but I still wonder, why this variable (and other parameters) are named `converter`. Why not just `replicaLocator`?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
```diff
@@ -159,14 +158,39 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
...
+      partitionKeyMapper)
+  }
+
+
+  /**
+   * A Serializable version of repartitionByCassandraReplica which removes
+   * the implicit RowWriterFactory Dependency
+   */
+  private[connector] def repartitionByCassandraReplica(
+    converter: ReplicaLocator[T],
```
pkolaczk: same as above

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
```diff
@@ -181,11 +205,22 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
...
+    val replicaLocator = ReplicaLocator[T](connector, keyspaceName, tableName, partitionKeyMapper)
+    rdd.keyByCassandraReplica(replicaLocator)
+  }
+
+  /**
+   * A Serializable version of keyByCassandraReplica which removes the implicit
+   * RowWriterFactory Dependency
+   */
+  private[connector] def keyByCassandraReplica(
+    converter: ReplicaLocator[T])(
```
pkolaczk: same as above

RussellSpitzer:
Fixed last converter renames!

pkolaczk:
+1

PR #831, opened 2015-09-11
SPARKC-149: sbt assembly task reports 4 test cases failed.

LukeTillman:
TableDefSpec was failing when building on Windows because TableDef.cql was using '\n' when generating the column list and this was being compared in the spec to a triple-quoted string with new line characters in it (which will be '\r\n' on Windows). This will make it use the platform-specific line separator instead when generating the column list.


RussellSpitzer:
Closed for PR against 1.5

PR #832, opened 2015-09-13
SPARKC-181:Implements leftJoinWithCassandraTable

ponkin:
Created CassandraLeftJoinRDD  (copy/paste from CasssandraJoinRDD) to implement RDDFunctions.leftJoinWithCassandraTable

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
@@ -7,7 +7,7 @@ import com.datastax.spark.connector.cql._
 import com.datastax.spark.connector.mapper.ColumnMapper
 import com.datastax.spark.connector.rdd.partitioner.{CassandraPartitionedRDD, ReplicaPartitioner}
 import com.datastax.spark.connector.rdd.reader._
-import com.datastax.spark.connector.rdd.{CassandraJoinRDD, SpannedRDD, ValidRDDType}
+import com.datastax.spark.connector.rdd.{CassandraLeftJoinRDD, CassandraJoinRDD, SpannedRDD, ValidRDDType}
 import com.datastax.spark.connector.writer.{ReplicaMapper, _}
 import org.apache.spark.SparkContext
 import org.apache.spark.SparkContext._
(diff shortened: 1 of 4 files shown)


RussellSpitzer:
I really appreciate the effort here. I know we've let this ticket sit for a while. I am a little worried about the implementation because of the big copy paste. I'm hoping that we can do this with a bit less code. Maybe if we extend the CassandraJoinRDD you will have less code duplication. This will also help with our coming refactor where we separate out the serializable from non-serializable code. I'll try to devote some time to a full code review after C* summit.

ponkin:
Unfortunately, we can not simply extends CassandraJoinRDD becase of RDD.compute method, for left join it must return Iterator[(L, Option[R])] not Iterator[(L, R)](like in CassandraJoinRDD). To completely eliminate copy/paste we need to refactor CassandraJoinRDD too.

RussellSpitzer:
Can't we extend CassandraJoinRDD[L,Option[R]]?

ponkin:
Here is the problem, CassandraJoinRDD must extend CassandraRDD[(L, __R__)] and CassandraLeftJoinRDD must extend CassandraRDD[(L, __Option[R]__)] cause they are collections of different objects. I do not think that we can extend one from another. May be we need to rethink CassandraRDD class in terms of self types and type classes not inheritance?

RussellSpitzer:
I think we can avoid that problem but extending the JoinRDD but the
RowReader ends up breaking us because we need to present a rowReader of
type Option[R] if we directly extend JoinRDD.

Something like
https://gist.github.com/RussellSpitzer/40d6aa6c1f0ff29c033d

The issue there is we'd also have the magic for creating a RowReader
Option[R] which means we add a little more logic to the RowReaderClass but
that brings us back to our core problem which is the API for RDD's is too
tightly coupled with the Spark Internals.

We discussed this on the checkpointing serilization ticket. We need to
refactor all the CassandraRDD classes so that their internal computation is
not so tightly coupled with their api. If for instance we had a
CassandraJoinRDDFactory, it could generate the proper rowReader and
rowWriter, then it creates a CassandraInnerJoinRDD and CassandraLeftJoinRDD
and we wouldn't have so many issues. If you'd like to be involved in those
discussions I can be sure to let you know when we start getting into the
muck :)

ponkin:
Yes, sure. Keep me in touch!

eallain-poctu:
I had planned to implement a Left Join and I've found this issue. I'm interested by this feature, is there anything missing ? I'd like to help if needed.

ponkin:
Hi, eallain-poctu
Here is the ticket [SPARKC-279](https://datastax-oss.atlassian.net/browse/SPARKC-279) where you can propose solution or something.

etspaceman:
+1 for this. We use a similar solution in production today and it would be great to have this in the general connector API.

ostefano:
+1

RussellSpitzer:
We have been taken a really long time on separating out our RowReader/Writer code, wondering if we should just rebase and merge this and clean it up when we do separate the other modules

ponkin:
Hi everybody, googd to know that @ostefano made this pull request. Let me suggest you a small fix: I have just moved all identical code in CassandraJoinRDD and CassandraLeftJoinRDD(copy-paste) to single trait AbstractCassandraJoin. Whatr do you think? I use my old fork. So if it is ok I will try to merge it with @ostefano pull request or something.

RussellSpitzer:
Merged from collaboration branch

RussellSpitzer:
https://github.com/datastax/spark-cassandra-connector/pull/1003

PR #834, opened 2015-09-14
SPARKC-229: Switched to Netty 4.0.23.Final for tests

jacek-lewandowski:


RussellSpitzer:
I need to merge this to make sure by SPARKC-202 bug is fixed.

PR #848, opened 2015-09-28
SPARKC-254 Added support timestamp in collection, udt and tuples

alexliu68:
Added a recursive method to update Sql row column with timestamp
data


pkolaczk:
+1

PR #850, opened 2015-10-01
SPARKC-259 Map InetAddress to string, UUID to binary in Spark SQL

alexliu68:
Because OSS SPARK doesn't add InetAddress and UUID as an internal type,
InetAddress is stored as host address string, and UUID is stored as
binary in the format of most significant bits first and least significant
bits second.


bcantoni:
Should we add any user docs to explain how these types are being mapped for SparkSQL? (Maybe under 2_loading.md.)

alexliu68:
Will do it.

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLSpec.scala
```diff
@@ -410,15 +410,17 @@ class CassandraSQLSpec extends SparkCassandraITFlatSpecBase {
...
       "from custom_type " +
-      "where CAST(a as string) > '/74.125.239.135'").collect()
+      "where a > '74.125.239.135'").collect()
     result should have length 1
     val result1 = cc.cassandraSql(
       "select k, v,  a, b " +
       "from custom_type " +
-      "where CAST(b as string) < '123e4567-e89b-12d3-a456-426655440000'").collect()
+      "where uuid_to_string(b) < '123e4567-e89b-12d3-a456-426655440000'").collect()
     result1 should have length 1
```
RussellSpitzer: Can you add some tests to check predicate pushdown on UUID? or will that still not work here?

alexliu68: it doesn't work at all. 

As our talk with Databrix, it's better off just to convert UUID to String. I will close this pull and open another one to converts InetAddress and UUID to String. Basically it's rolling back to original one with converting data to String.

RussellSpitzer: yeah we've had a lot of requests so although it will cost us in terms of size we should do string for now :(

alexliu68:
close this pull for now, will have a new pull to covert UUID to String

PR #851, opened 2015-10-05
SPARKC-262: Set write consistency level to LOCAL_QUORUM.

pkolaczk:
With the current default write consistency level of LOCAL_ONE,
it's relatively easy to overwhelm a cluster because it only waits
for one replica to ack the write before proceeding on.
Changing the default write consistency level to LOCAL_QUORUM means
that it will do a little more natural throttling to allow the cluster to keep up
with the writes for multiple replicas.
It seems like it would make for a more stable out of the box experience.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@
  * Added support for tinyint and smallint types (SPARKC-269)
  * Updated Java driver version to 3.0.0-alpha3.
  * Changed the way CassandraConnectorSource is obtained due to SPARK-7171 (SPARKC-268)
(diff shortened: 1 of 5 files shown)


pkolaczk:
@bcantoni fixed

bcantoni:
+1

pkolaczk:
jenkins has some problem today:
`02:29:27 ERROR: SEVERE ERROR occurs
02:29:27 org.jenkinsci.lib.envinject.EnvInjectException: hudson.AbortException: Failed to install JDK. Exit code=2`

PR #852, opened 2015-10-08
SPARKC-266 Fixed VarInt Column filter error In Spark SQL

alexliu68:
VarInt column is mapped to Spark Sql Decimal type, the data needs
be converted to decimal to Catalyst. It also needs to be converted
back to BigInteger when push down the filters to Cassandra.

### DIFF
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLSpec.scala
@@ -435,4 +435,34 @@ class CassandraSQLSpec extends SparkCassandraITFlatSpecBase {
       "where CAST(b as string) < '123e4567-e89b-12d3-a456-426655440000'").collect()
     result1 should have length 1
   }
+
+
+  it should "be able to push down filter on varint columns" in {
+    conn.withSessionDo { session =>
+      session.execute(
+        s"""
+        |CREATE TABLE sql_test.blob_test(
+        |  id varint,
+        |  series varint,
+        |  rollup_minutes varint,
+        |  event text,
+        |  PRIMARY KEY ((id, series, rollup_minutes), event)
+        |)
+      """.stripMargin.replaceAll("\n", " "))
+      session.execute(
+        s"""
+        |INSERT INTO sql_test.blob_test(id, series, rollup_minutes, event)
+        |VALUES(1234567891234, 1234567891235, 1234567891236, 'event')
+      """.stripMargin.replaceAll("\n", " "))
+    }
+    val cc = new CassandraSQLContext(sc)
+    cc.sql(
+      s"""
+        |SELECT * FROM sql_test.blob_test
+        |WHERE id = 1234567891234
+        | AND series = 1234567891235
+        | AND rollup_minutes = 1234567891236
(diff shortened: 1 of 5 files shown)


bcantoni:
Any other doc changes needed? (Probably CHANGES.txt at least.)

alexliu68:
It could be added to data mapping section to describe how VarInt data is stored in Spark SQL

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLSpec.scala
```diff
@@ -421,4 +419,33 @@ class CassandraSQLSpec extends SparkCassandraITFlatSpecBase {
...
+        |  PRIMARY KEY ((id, series, rollup_minutes), event)
+        |)
+      """.stripMargin.replaceAll("\n", " "))
+      session.execute(
+        s"""
+        |INSERT INTO sql_test.blob_test(id, series, rollup_minutes, event)
+        |VALUES(1234567891234, 1234567891235, 1234567891236, 'event')
+      """.stripMargin.replaceAll("\n", " "))
+    }
+    val cc = new CassandraSQLContext(sc)
```
RussellSpitzer: we still have no way of programatically checking if things are pushed down?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -125,19 +126,48 @@ private[cassandra] class CassandraSourceRelation(
...
+  /** If column is VarInt column, convert data to BigInteger */
+  private def toCqlValue(columnName: String, value: Any): Any = {
+    if (VarIntColumn(columnName)) {
+      toBigInt(value)
+    } else {
+      value
+    }
+  }
+
+  private def VarIntColumn(columnName: String): Boolean = {
```
RussellSpitzer: possibly name this `isVarInt(columnName:String)`

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -127,19 +128,48 @@ private[cassandra] class CassandraSourceRelation(
...
+    tableDef.columnByName(columnName).columnType == VarIntType
+  }
+
+  /** Convert decimal to BigInteger */
+  private def toBigInt(value: Any): Any = {
```
RussellSpitzer: why not lock the input of this function to decimal? Also think about naming decimalToBigInt? Then error out if called on non-bigint column

I think the best solution is to make this function and lock input to the type decimal then in toCqlValue do a match on type so the compiler will be happy

alexliu68: non-bigint should return the original value, decimal converts to BigInteger.

There are toCqlValues and toCqlValue both need this function, so if add matching to those function, we duplicate the matching in two places, that's the reason I refactor it into a function.

alexliu68:
@RussellSpitzer I simplify the functions in CassandraSourceRelation, you can check it out

RussellSpitzer:
+1

PR #856, opened 2015-10-14
SPARKC-271: Let UDTs be converted from GenericRows

RussellSpitzer:
When we read a UDT from a C* table it becomes nested generic
rows inside of dataframes. These nested row objects are not convertable
back into UDT types so they cannot be written back into C*. This patch
enables the conversion of nested GenericRows back into UDTS if the
naming is correct.

### DIFF
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLSpec.scala
@@ -14,6 +14,7 @@ class CassandraSQLSpec extends SparkCassandraITFlatSpecBase {
   var cc: CassandraSQLContext = null
 
   conn.withSessionDo { session =>
+    session.execute("DROP KEYSPACE IF EXISTS sql_test")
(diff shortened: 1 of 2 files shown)


kunalak:
How do I try this change @RussellSpitzer ? is there an udpated .jar?

RussellSpitzer:
@kunalak  https://s3.amazonaws.com/uploads.hipchat.com/6528/347558/E21k9ZmFTHlQc5o/spark-cassandra-connector-assembly-1.4.0-SNAPSHOT.jar You can download a jar i built there. In the future you can always build these artifacts by checkout the branch at running ./sbt/sbt assembly

alexliu68:
+1

goginoida:
@RussellSpitzer We are using version 1.4.4 which contains this fix (included in DSE 4.8.8). It seems if the value for UDT column is not present for a row (null when queried through cqlsh), writing it to another table throws exception saying "Cannot convert object null to com.datastax.spark.connector.UDTValue".

RussellSpitzer:
That was solved in https://datastax-oss.atlassian.net/browse/SPARKC-346 which is a different ticket and is only present on the 1.6.x and greater branches. Please file a Jira if you would like it backported

PR #860, opened 2015-10-15
SPARKC-270: Updated C* Driver Version to 3.0.0-Alpha3

RussellSpitzer:
Updated to 3.0 version of the Java Driver. The getIndex call is no
longer valid for ColumnMetdata objects because in 3.0 columns can have
multiple indexes. To accomadate this we now use the list of all indexes
in the table to determine if any have targeted the column in question.


pkolaczk:
+1

PR #867, opened 2015-10-23
SPARKC-259 Map InetAddress to string, UUID to binary in Spark SQL

alexliu68:
Because OSS SPARK doesn't add InetAddress and UUID as an internal type,
InetAddress is stored as host address string, and UUID is stored as
binary in the format of most significant bits first and least significant
bits second.


spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -150,6 +152,13 @@ private[cassandra] class CassandraSourceRelation(
       case decimal: Decimal =>
         val isVarIntColumn = tableDef.columnByName(columnName).columnType == VarIntType
         if (isVarIntColumn) decimal.toJavaBigDecimal.toBigInteger else value
+      case uTF8String: UTF8String =>
```
RussellSpitzer: uTF just looks weird to me , maybe utf8?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -150,6 +152,13 @@ private[cassandra] class CassandraSourceRelation(
...
+        val columnType = tableDef.columnByName(columnName).columnType
+        if (columnType == InetType) {
+          InetAddress.getByName(uTF8String.toString)
+        } else if(columnType == UUIDType) {
+          UUID.fromString(uTF8String.toString)
+        }
```
RussellSpitzer: also don't we need an else cluase here in case the column isn't inet or UUID?

alexliu68: right

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
alexliu68: yes

RussellSpitzer:
+1 if Decimal => decimal instead of value, I know i'm being annoying about consistency there

PR #868, opened 2015-11-02
Sparkc 277

RussellSpitzer:


jacek-lewandowski:
These great tests added a few minutes to the testing time. I noticed that most of the time is spend on creating keyspace and 11 tables - is it possible that there a single table with multiple all the required columns instead?

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/typeTests/AbstractTypeTest.scala
```diff
@@ -0,0 +1,544 @@
...
+
+import com.datastax.spark.connector.types.TypeConverter
+import com.datastax.spark.connector.writer.RowWriterFactory
+import org.apache.spark.sql.SQLContext
+
+import scala.collection.JavaConverters._
+import scala.reflect._
+
+/**
+ * A templated class for testing that various CQL Types work with the spark Cassandra Connector
```
jacek-lewandowski: few typos: templated, implmentation, overidden

jacek-lewandowski:
I also noticed that Spark Context is restarted for every suite class. It looks like the configuration is modified internally by Spark and therefore it is different to what was remembered in our testing framework.

RussellSpitzer:
@jacek-lewandowski Found the issue with the SparkConf and removed the added keys. Consolidated type keyspaces to a single keyspace. We could further reduce the number of table formats checked since it currently tries out several "normal" tables

RussellSpitzer:
How about we set the tables to test at

Compound Table (pkey, ckey, data)
Collection Table ( Map, List, Set )
Null Table ( all fields -- to be nulled)?

I do like them separate because it is much easier to determine which particular component is failing but I do not want to take up too much time :(

jacek-lewandowski:
I like this idea

RussellSpitzer:
@jacek-lewandowski Fixed this up as well, testing should be MUCH faster now :)

jacek-lewandowski:
Testing locally:
```
[error] Test com.datastax.spark.connector.types.TypeConverterTest.testLocalDate failed: java.lang.AssertionError: expected:<1985-08-03> but was:<1985-08-02>
[error]     at com.datastax.spark.connector.types.TypeConverterTest.testLocalDate(TypeConverterTest.scala:208)
[error]     ...
```

RussellSpitzer:
I hate timezones :(

doc/2_loading.md
```diff
@@ -172,7 +172,8 @@ The following table shows recommended Scala types corresponding to Cassandra col
 | `bigint`          | `Long`                                       
 | `blob`            | `ByteBuffer`, `Array[Byte]` 
 | `boolean`         | `Boolean`, `Int`              
-| `counter`         | `Long`                       
+| `counter`         | `Long` 
+| `date`            | `Int`, `String` (YYYY-MM-DD), `java.util.Date`, `java.sql.date`, `org.joda.time.DateTime`
```
jacek-lewandowski: typo: java.sql.date => java.sql.Date

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
```diff
@@ -37,12 +37,25 @@ object SparkTemplate {
 
   private var _sc: SparkContext = _
 
+
+
+  val SparkAddedKeys = Seq(
+  "spark.driver.host",
```
jacek-lewandowski: formatting

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/typeTests/AbstractTypeTest.scala
```diff
@@ -0,0 +1,429 @@
...
+         |(pkey, data1, nulldata, nullset, nulllist, nullmap)
+         |VALUES (?,?,?,?,?,?)""".stripMargin)
+    typeData.foreach { value =>
+      val driverValue = convertToDriverInsertable(value)
+      val bs = nullStatement.bind(driverValue, driverValue, null, null, null, null)
+      session.execute(bs)
+    }
+  }
+
+  def generateKsTable() = conn.withSessionDo { session =>
```
jacek-lewandowski: Ok, so if we can make it a single table - that would increase testing speed even more

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
```diff
@@ -409,6 +420,35 @@ object TypeConverter {
...
+
+  implicit object LocalDateConverter extends NullableTypeConverter[LocalDate] {
+    def targetTypeTag = LocalDateTypeTag
+    val dateRegx = """(\d\d\d\d)-(\d\d)-(\d\d)""".r
+
+    def addTimeZoneOffset(millis: Long) =  millis + defaultTimezone.getOffset(millis)
+
+    def convertPF = {
+      case x: LocalDate => x
+      case x: String => x match { case dateRegx(year,month,day) =>
```
jacek-lewandowski: formatting

jacek-lewandowski: wouldn't it be better to match dateRegx immediately?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
```diff
@@ -409,6 +420,35 @@ object TypeConverter {
...
+      case x: java.sql.Date => LocalDate.fromMillisSinceEpoch(addTimeZoneOffset(x.getTime))
+      case x: Date => LocalDate.fromMillisSinceEpoch(x.getTime)
+    }
+  }
+
+  object TimeTypeConverter extends NullableTypeConverter[java.lang.Long] {
+    val NANOS_TO_MILLIS = 1000000L
+    def targetTypeTag = JavaLongTypeTag
+    def convertPF = {
+      case x: Date => x.getTime * NANOS_TO_MILLIS
```
jacek-lewandowski: maybe `TimeUnit.MILLISECONDS.toNanos(x.getTime)` ?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
```diff
@@ -409,6 +420,35 @@ object TypeConverter {
...
+    val dateRegx = """(\d\d\d\d)-(\d\d)-(\d\d)""".r
+
+    def addTimeZoneOffset(millis: Long) =  millis + defaultTimezone.getOffset(millis)
+
+    def convertPF = {
+      case x: LocalDate => x
+      case x: String => x match { case dateRegx(year,month,day) =>
+        LocalDate.fromYearMonthDay(year.toInt, month.toInt, day.toInt)}
+      case x: Int => LocalDate.fromDaysSinceEpoch(x)
+      case x: java.sql.Date => LocalDate.fromMillisSinceEpoch(addTimeZoneOffset(x.getTime))
```
jacek-lewandowski: Perhaps this needs a comment

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
```diff
@@ -303,8 +308,14 @@ object TypeConverter {
   }
 
   implicit object SqlDateConverter extends NullableTypeConverter[java.sql.Date] {
+
+    def subtractTimeZoneOffset(millis: Long) = millis - defaultTimezone.getOffset(millis)
+
     def targetTypeTag = SqlDateTypeTag
-    def convertPF = DateConverter.convertPF.andThen(d => new java.sql.Date(d.getTime))
+
+    def convertPF = PartialFunction[Any, java.sql.Date]{ case x: LocalDate =>
```
jacek-lewandowski: Either I get it wrong or it will fail unless the object is of `LocalDate` type - is it intended here? Previously `DateConverter.convertPF` was used which is able to handle many different types...

RussellSpitzer: I thought the "andThen" would compose these to, did i mean to do an orElse?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
RussellSpitzer: Not sure what you mean here? ( i understand the formatting mistake but not the regex immediately match?, Do you mean case dateRegex instead of case String?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
RussellSpitzer: That's a good idea, i'll add that everywhere we do the * NANOS_TO_MILLIS

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
jacek-lewandowski: exactly

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
jacek-lewandowski: My understanding is that the `case x: LocalDate => ...` will be inner and `DateConverter.convertPF` will be outer. So if the inner function fails, the it will fail, because it has no alternative. Also, look that the type of the later function - the argument type is the same as return type of the first function.

Consider this:
```
scala> val pf1 = PartialFunction[Any, String] {
  case x: String => s"String $x"
  case x: Int => s"Int $x"
}
     |      |      | pf1: PartialFunction[Any,String] = <function1>

scala> val pf0 = PartialFunction[Any, String] {
  case x: Int => s"*** $x"
}
     |      | pf0: PartialFunction[Any,String] = <function1>

scala> val pf = pf0 andThen pf1
pf: PartialFunction[Any,String] = <function1>

scala> pf.apply(1)
res3: String = String *** 1

scala> pf.apply("a")
scala.MatchError: a (of class java.lang.String)
```

I think you should use `orElse` as you said.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
RussellSpitzer: check

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
RussellSpitzer: check

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
RussellSpitzer: comment added

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
RussellSpitzer: patched

RussellSpitzer:
retest this please

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/typeTests/AbstractTypeTest.scala
```diff
@@ -0,0 +1,427 @@
...
+    val start = System.currentTimeMillis()
+    session.execute(s"""DROP KEYSPACE IF EXISTS $keyspaceName""")
+
+    session.execute(
+      s"""CREATE KEYSPACE IF NOT EXISTS $keyspaceName
+         |with replication = {'class':'SimpleStrategy','replication_factor':'1'}""".stripMargin)
+    session.execute(s"""USE $keyspaceName""")
```
jacek-lewandowski: Remove drop/create keyspace, use `createKeyspace(session)` instead

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/typeTests/AbstractTypeTest.scala
```diff
@@ -0,0 +1,427 @@
...
+
+    session.execute(
```
jacek-lewandowski: use `awaitAll(Future{...}, Future{...}, ...) to create independent schema elements

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/typeTests/AbstractTypeTest.scala
```diff
@@ -0,0 +1,427 @@
...
+         |nullset set<$typeName>,
+         |nulllist list<$typeName>,
+         |nullmap map<$typeName,$typeName>,
+         |PRIMARY KEY (pkey))""".stripMargin)
+    val end = System.currentTimeMillis()
+    println(s"Took ${end-start} ms to setup tables")
+  }
+
+  def truncateTables() = conn.withSessionDo { session =>
+    session.execute(s"TRUNCATE $typeNormalTable")
```
jacek-lewandowski: use `awaitAll`

jacek-lewandowski:
@RussellSpitzer could you follow new conventions in integration tests, take a look at, say `CassandraRDDSpec` or some other integration test class. This will reduce the testing time which is needlessly added by creation of tables.

RussellSpitzer:
@jacek-lewandowski Asynced :)

jacek-lewandowski:
+1

Aarushi2016:
Could you please advise on converting cassandra "Time" column thru Java API.

we have defined a cassandra table column with "Time" as data type and trying to read the data from java API with "long" as return type. Could not find any mechanism to convert and display back as time from the long value. Please see below for an example.

Cassandra Table time column value: 10:12:24.067800000
long value returned after select from api: 36744067800000

we would like to convert and see the original table value(10:12:24.067800000) from the above long, in order to display in the front end.

Really appreciate your suggestions here.

Regards,
Ravi.

RussellSpitzer:
What you are getting *is* the original table value just in the underlying
form which is nanoseconds from midnight.
http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/TimeUnit.html
<https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/TimeUnit.html>
Use the TimeUnit Converters,

TimeUnit.SECONDS.convert(elapsedTime, TimeUnit.NANOSECONDS)


From
http://stackoverflow.com/questions/924208/how-to-convert-nanoseconds-to-seconds-using-the-timeunit-enum

scala> TimeUnit.HOURS.convert(t, TimeUnit.NANOSECONDS)
res1: Long = 10

scala> TimeUnit.MINUTES.convert(t, TimeUnit.NANOSECONDS)%60
res4: Long = 12

scala> TimeUnit.SECONDS.convert(t, TimeUnit.NANOSECONDS)%60%60
res6: Long = 24

Aarushi2016:
Thanks much Russell for your immediate response.

My API’s output is – 36744067800000, for the actual time value in the table: 10:12:24.067800000.

Was trying to get time from the above API output, it returned: 08:30:00.

Am trying to find out what is the best and optimum way to convert long value to “time”, which is in database.

Please advise if need any additional details.


Regards,
Ravi.

RussellSpitzer:
Just convert it? As I showed you can pull out the hours, minutes and
seconds components.  If you are looking for a built in solution you can try
using the converter in the TimeCodec in the java driver which is public (at
least for now)

scala> import com.datastax.driver.core.TypeCodec
import com.datastax.driver.core.TypeCodec

scala> TypeCodec.time().format(36744067800000L)
res0: String = '10:12:24.067800000'

RussellSpitzer:
My driver buddies also let me know there is "ParseUtils"
Which is what that codec uses under the hood
https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java#L398

Aarushi2016:
I was using LocalTime to convert, but as I mentioned, it did not turn up correctly.

I tried with built in function that you mentioned below, it worked well. I would like to thank you for your prompt attention and response.

Thanks again!

Aarushi2016:
Good Morning.

Could any one please assist or advise me in implementing pagination logic for cassandra.

We are trying to find the best way to fetch data from cassandra table and send to UI, using java API.

Thank you and advise me if need any additional details on the same.


Regards,
Ravi.

Aarushi2016:
Good Morning!

Can some one please advise me on inserting a "date" and "time" columns into cassandra from Java API.

I have columns memo_date as "date" datatype and memo_time as "time" datatype.
I am trying insert values into these columns, but could not able to do it successfully, ended up below errors.

My input values are - "yyyy-mm-dd" and "10:02:05:2345"

java.lang.String cannot be cast to java.lang.Integer - where i have used TypeCodec.date().foramt()
java.time.LocalDate cannot be cast to java.lang.Integer  - where i used localdate.

Any suggestion is really appreciated.



Thank you!
Ravi.

PR #870, opened 2015-11-03
DSP-7062: Open Visibility on RowReader

RussellSpitzer:
Implementing some DSE only functionality requires this to be more
publically accessible. See DSP-7062


alexliu68:
+1

PR #873, opened 2015-11-09
Sparkc 283

RussellSpitzer:


RussellSpitzer:
@jacek-lewandowski ^ :)

pkolaczk:
How do I erase some value to null then?

pkolaczk:
We probably need some special Option-like type with 3 states:
- Some(value)
- None (working as previously -> inserts null, represents lack of value)
- Unset (passing unset to the Java driver, represents lack of change)

I don't like giving Scala null any special meaning, because null is only for Java-interoperability and should not be used in the Scala code.

PR #886, opened 2015-11-20
SPARKC-291: Single Pass Filtering for Pushdown C*Source DF

RussellSpitzer:
Spark 1.6 adds a overridable method unhandledFilters which causes the
catalyst engine to ignore a set filters that the underlying data source
is exhaustively filtering. All of the filters we push down are absolute
so it is safe to remove all pushed down filters and return remaining
filters back to Catalyst.


pkolaczk:
+1

PR #887, opened 2015-11-20
SPARKC-228: Warn in Streaming Applications with Small Keep_Alive

RussellSpitzer:
We've run into several end users who are running Streaming applications
with a keep_alive short enough that a new connection is made every
batch. This ends up causing a lot of stress on the C* cluster and an
unneccessary (and time consuming) slowdown every batch.

To avoid these issues, This patch increases the default keep_alive to
5 seconds. Also warn the user if they use a C* DStream option with a
lower frequency than the batch interval.


RussellSpitzer:
retest this please

pkolaczk:
+1

PR #889, opened 2015-11-24
Improve it tests

jacek-lewandowski:
Several changes have been made in this patch. To name a few:
- Bumped Scala 2.11 compiler to the newest version 2.11.7
- Reenabled Spark REPL tests
- Integration test classes always get the configuration from SparkTemplate.defaultConf
- SparkTemplate defaultConf is now a method which returns a clone of the config
- SparkTemplate alows to disable SparkContext and doesn't start it implicitly
- Fixed SparkTemplate problem that SparkContext was unnecesarily restarted sometimes
- Cassandra runner looks for a free range of ports
- Cassandra runner stops uncessary instances immediately
- Tuned some memory requirements
- Integration tests generates their schema in parallel in order to minimise table creation fixed delays
- Data is generated using executeAsync rather than execute
- Ignored CheckpointStreamSpec as it showed flaky, especially in parallel runs (to be addressed in a separate ticket)
- Each test uses a dedicated keyspace, whose name is consistently generated from test class name
- Disabled durable writes for testing
- Tests were groupped so that at least repl, streaming, auth, ssl are run in separate processes
- Tests run in parallel, the parallelisation level is automatically resolved according to the amount of OS memory (roughly we need 3 GB per single test task)


project/Settings.scala
```diff
@@ -214,28 +246,71 @@ object Settings extends Build {
     publish in (IntegrationTest,packageBin) := ()
   )
 
+  def makeTestGroups(tests: Seq[TestDefinition]) = {
```
pkolaczk: for a complex method like this an explicit return type would improve readability

spark-cassandra-connector-embedded/scala-2.11/src/main/scala/com/datastax/spark/connector/embedded/SparkRepl.scala
```diff
@@ -28,12 +27,21 @@ trait SparkRepl {
     }
 
     val interp = new SparkILoop(Some(in), new PrintWriter(out))
-    org.apache.spark.repl.Main.interp = interp
+    Main.interp = interp
+    // due to some mysterious compilation error classServer needs to be accessed via reflection
+    //      Main.classServer.start()
```
pkolaczk: would be nice to get more info about that mysterious compilation error

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/DynamicCassandraPorts.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  private lazy val basePath = Paths.get(sys.props("baseDir"), "target", "ports")
+
+  private lazy val basePorts = 10000 until 20000 by 50
+
+  private def portLockFile(port: Int) = basePath.resolve(s"port_$port.lock")
+
+  def apply(): DynamicCassandraPorts = DynamicCassandraPorts(chooseBasePort())
+
+  def verifyRange(host: String, ports: Seq[Int]): Boolean = {
```
pkolaczk: Minor nitpick: The method name does not make it obvious what the result means. It has no side effect, so using a verb is also not needed. `areAllPortsAvailable`?

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/UserDefinedProperty.scala
```diff
@@ -0,0 +1,66 @@
+package com.datastax.spark.connector.embedded
+
+import java.net.InetAddress
+
+import org.apache.commons.configuration.ConfigurationException
+
+object UserDefinedProperty {
```
pkolaczk: De ja vu. Was it moved from somewhere?

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/SparkCassandraITFlatSpecBase.scala
```diff
@@ -13,4 +20,41 @@ trait SparkCassandraITWordSpecBase extends WordSpec with SparkCassandraITSpecBas
...
+    val className = this.getClass.getSimpleName
+    val suffix =  StringUtils.splitByCharacterTypeCamelCase(className.filter(_.isLetterOrDigit)).mkString("_")
+    s"test_$suffix".toLowerCase()
+  }
+
+  val ks = getKsName
+
+  implicit val ec = SparkCassandraITSpecBase.ec
+
+  def doAsync(units: Future[Unit]*): Unit = {
```
pkolaczk: A blocking method called "Async". I guess the purpose of this method is to just wait for all the futures to complete, right? So why not name it `awaitAll` ?

pkolaczk:
Overall looks like a great improvement. +1

project/Settings.scala
jacek-lewandowski: ok

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/DynamicCassandraPorts.scala
jacek-lewandowski: ok

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/UserDefinedProperty.scala
jacek-lewandowski: I just extracted few classes from a single file into separate files.

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/SparkCassandraITFlatSpecBase.scala
jacek-lewandowski: ok

pkolaczk:
+1

PR #892, opened 2015-12-05
SPARKC-283: Adds CassandraOption for Skipping Empty Columns

RussellSpitzer:
Previously unset columns had to be set to null since the Cassandra
Protocol <= V3 did not support unbound parameters in bound statements.
The newere version of the protocol allows for this, and this patch
enables the connector to take advtange of that behavior.

This is done through use of a new option container, CassandraOption.
This class holds an Option and the behavior that a None should have upon
being written. With this class a row can have varying behavior from
column to column and row to row.

If an Unset is used on a C* version prior to Protcol Version 3 a warning
will be logged that the Unset functionality cannot be used and the value
will be treated as a Delete instead.

### DIFF
--- a/doc/5_saving.md
+++ b/doc/5_saving.md
@@ -189,6 +189,142 @@ val company = Company("DataStax", address)
 sc.parallelize(Seq(company)).saveToCassandra("test", "companies")
 ```
 
+## Skipping Columns and Avoiding Tombstones on Writes (Connector Version 1.6+ and Cassandra 2.2+)
+Prior to Cassandra 2.2 there was no way to execute a prepared statement with unbound elements. 
+This meant every executed statement via the Spark Cassandra Connector was required 
+to bind `nulls` into for any unspecified columns. As of Cassandra 2.2, the native protocol now allows 
+for leaving parameters unbound.
+
+To take advantage of unset parameters, the Spark Cassandra Connector now provides a method for 
+taking advantage of this unbound behavior. This is done by with the 
+`com.datastax.spark.connector.types.CassandraOption` trait. 
+
+The trait has three implementations
+```scala
+sealed trait CassandraOption[+A] extends Product with Serializable
+  
+  object CassandraOption {
+    case class Value[+A](value: A) extends CassandraOption[A]
+    case object Unset extends CassandraOption[Nothing]
+    case object Null extends CassandraOption[Nothing]
+```
+
+This can be used when reading and writing from C*. When a column is loaded as a `CassandraOption` 
+any missing columns will be represented as `Unset`. On writing, these parameters will remain unbound.
+This means a table loaded via `CassandraOption` can be written to a second table without any missing
+column values being treated as deletes.
+
+####Example: Copying a table without deletes
+```sql
+//cqlsh
+CREATE TABLE doc_example.tab1 (key INT, col_1 INT, col_2 INT, PRIMARY KEY (key))
+INSERT INTO doc_example.tab1 (key, col_1, col_2) VALUES (1, null, 1)
+CREATE TABLE doc_example.tab2 (key INT, col_1 INT, col_2 INT, PRIMARY KEY (key))
+INSERT INTO doc_example.tab2 (key, col_1, col_2) VALUES (1, 5, null)
+```
+
+```scala
+//spark-shell
+val ks = "doc_example"
+//Copy the data from tab1 to tab2 but don't delete when we see a null in tab1
+sc.cassandraTable[(Int, CassandraOption[Int], CassandraOption[Int])](ks, "tab1")
+  .saveToCassandra(ks, "tab2")
+  
+sc.cassandraTable[(Int,Int,Int)](ks, "tab2").collect
+//(1, 5, 1)
+```
+ 
+For more complicated use cases the `CassandraOption` can be set to delete on a per row 
+(and per column) basis. This is done by using either the `Unset` or `Null` case objects.
+
+####Example of using different None behaviors
+```scala
+//Fill tab1 with (1, 1, 1) , (2, 2, 2) ... (6, 6, 6)
+sc.parallelize(1 to 6).map(x => (x, x, x)).saveToCassandra(ks, "tab1")
+//Delete the second column when x >= 5
+//Delete the third column when x <= 2
+//For other rows put in the value -1
+sc.parallelize(1 to 6).map(x => x match {
+  case x if (x >= 5) => (x, CassandraOption.Null, CassandraOption.Unset)
+  case x if (x <= 2) => (x, CassandraOption.Unset, CassandraOption.Null)
+  case x => (x, CassandraOption(-1), CassandraOption(-1))
+}).saveToCassandra(ks, "tab1")
+
+val results = sc.cassandraTable[(Int, Option[Int], Option[Int])](ks, "tab1").collect
+results 
+/*
+  (1, Some(1), None),
+  (2, Some(2), None),
+  (3, Some(-1), Some(-1)),
+  (4, Some(-1), Some(-1)),
+  (5, None, Some(5)),
+  (6, None, Some(6)))
+*/
+```
+  
+CassandraOptions can be converted to Scala Options via an implemented implicit. This means that 
+CassandraOptions can be dealt with as if they were normal Scala Options. For the reverse 
+transformation, from a Scala Option into a CassandraOption, you need to define the None behavior. 
+This is done via `CassandraOption.deleteIfNone` and `CassandraOption.unsetIfNone`
+ 
+####Example of converting Scala Options to Cassandra Options
+```scala
+import com.datastax.spark.connector.types.CassandraOption
+//Setup original data (1, 1, 1) ... (6, 6, 6)
+sc.parallelize(1 to 6).map(x => (x,x,x)).saveToCassandra(ks, "tab1")
+
+//Setup options Rdd (1, None, None) (2, None, None) ... (6, None, None)
+val optRdd = sc.parallelize(1 to 6)
+  .map(x => (x, None, None))
+   
+//Delete the second column, but ignore the third column
+optRdd
+  .map{ case (x: Int, y: Option[Int], z: Option[Int]) =>
+    (x, CassandraOption.deleteIfNone(y), CassandraOption.unsetIfNone(z))
+  }.saveToCassandra(ks, "tab1")
+
+val results = sc.cassandraTable[(Int, Option[Int], Option[Int])](ks, "tab1").collect
+results
+/*
+    (1, None, Some(1)),
+    (2, None, Some(2)),
+    (3, None, Some(3)),
+    (4, None, Some(4)),
+    (5, None, Some(5)),
+    (6, None, Some(6))
+*/
+```
+     
+### Leaving all nulls as Unset
+WriteConf also now contains a parameter `ignoreNulls` which can be set via using a `SparkConf` key
+`spark.cassandra.output.ignoreNulls`. The default is `false` which will cause `null`s to be treated 
+as in previous versions (being inserted into C* as is). When set to `true` all `null`s will be 
+treated as `unset`. This can be used with DataFrames to skip null records and avoid tombstones.
+
+####Example of using ignoreNulls
+```scala
+//Setup original data (1, 1, 1) --> (6, 6, 6)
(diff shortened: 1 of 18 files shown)


RussellSpitzer:
@jacek-lewandowski Rebased

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/CassandraRDDSpec.scala
```diff
@@ -332,6 +331,11 @@ class CassandraRDDSpec extends SparkCassandraITFlatSpecBase {
     rowById(2).getBytesOption("b") shouldEqual None
   }
 
+  it should "allow for reading Cassandra Options from nulls" in {
```
jacek-lewandowski: What about testing whether {{CassandraOption}} can be read from non-nulls

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/writer/BoundStatementBuilderSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+  val rowWriter = RowWriterFactory.defaultRowWriterFactory[(Int, CassandraOption[String])]
+    .rowWriter(schema.tables.head, IndexedSeq("id", "value"))
+  val rkg = new RoutingKeyGenerator(schema.tables.head, Seq("id", "value"))
+  val ps = conn.withSessionDo( session =>
+    session.prepare(s"""INSERT INTO "$ks".tab (id, value) VALUES (?, ?) """))
+
+  val protocolVersion = conn.withClusterDo(cluster => cluster.getConfiguration.getProtocolOptions.getProtocolVersion)
+
+  "BoundStatementBuilder" should "ignore Unset values if ProtocolVersion > 3" in {
+    val bsb = new BoundStatementBuilder(rowWriter, ps, protocolVersion = ProtocolVersion.V4)
```
jacek-lewandowski: perhaps should be iterated through protocol versions which `protocol.asInt >= ProtocolVersion.V4.asInt`

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/writer/BoundStatementBuilderSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+    val x = bsb.bind((1, CassandraOption.Unset))
+    x.isSet("value") should be (false)
+  }
+
+  it should "set Unset values to null if ProtocolVersion <= 3" in {
+    val bsb = new BoundStatementBuilder(rowWriter, ps, protocolVersion = ProtocolVersion.V3)
```
jacek-lewandowski: same here

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/ReadConf.scala
```diff
@@ -6,7 +6,7 @@ import org.apache.spark.SparkConf
 
 /** Read settings for RDD
   *
-  * @param splitCount number of partitions to divide the data into; unset by default
+  * @param splitCount number of partitions to divide the data into; unsetIfNone by default
```
jacek-lewandowski: ?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/Unset.scala
```diff
@@ -0,0 +1,21 @@
...
+  */
+protected[connector] object Unset extends Serializable
+
+/**
+  * An option with an extra bit of information to let us know whether a value should be treated
+  * as a delete or as an unsetIfNone value. Reading a table using this as an column type will
+  * cause all empty values to not be treated as deletes on write.
+  */
+case class CassandraOption[A](option: Option[A], unsetIfNone: Boolean = true) {
+  def UnsetIfNone = copy(option, true)
```
jacek-lewandowski: Perhaps such methods should be in a companion object

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/BoundStatementBuilder.scala
```diff
@@ -38,11 +39,18 @@ private[connector] class BoundStatementBuilder[T](
...
       val columnValue = converter.convert(buffer(i))
-      boundStatement.set(columnName, columnValue, CodecRegistryUtil.codecFor(columnTypes(i), columnValue))
-      val serializedValue = boundStatement.getBytesUnsafe(i)
-
-      if (serializedValue != null)
-        bytesCount += serializedValue.remaining()
+      //C* 2.2 (PV4) and Greater Allows us to leave fields Unset
+      if (protocolVersion.toInt < ProtocolVersion.V4.toInt || columnValue != Unset) {
+        if (columnValue == Unset){
+          logWarning(s"""Unset Value set to Null because Current $protocolVersion <
```
jacek-lewandowski: Please remove log message from there. This is executed for every single value which is inserted and will surely lead to kill the server (log file can grow much faster than the database!). Instead, you can make some flag and report warning once the whole batch is processed. Please do not log anything per column or even per row.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/BoundStatementBuilder.scala
```diff
@@ -38,11 +39,18 @@ private[connector] class BoundStatementBuilder[T](
       val converter = converters(i)
       val columnName = columnNames(i)
       val columnValue = converter.convert(buffer(i))
-      boundStatement.set(columnName, columnValue, CodecRegistryUtil.codecFor(columnTypes(i), columnValue))
-      val serializedValue = boundStatement.getBytesUnsafe(i)
-
-      if (serializedValue != null)
-        bytesCount += serializedValue.remaining()
+      //C* 2.2 (PV4) and Greater Allows us to leave fields Unset
+      if (protocolVersion.toInt < ProtocolVersion.V4.toInt || columnValue != Unset) {
```
jacek-lewandowski: I think we can make a function reference and assign one implementation or another once `BoundStatementBuilder` is created, and do not repeat here.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/ReadConf.scala
RussellSpitzer: refactors gone wild :(

RussellSpitzer:
@jacek-lewandowski Comments addressed

RussellSpitzer:
I also added a writeConf parameter for use with SparkSQL

doc/5_saving.md
```diff
@@ -189,6 +189,135 @@ val company = Company("DataStax", address)
 sc.parallelize(Seq(company)).saveToCassandra("test", "companies")
 ```
 
+## Skipping Columns and Avoiding Tombstones on Writes (Connector Version 1.6+ and Cassandra 2.2+)
```
jacek-lewandowski: @RussellSpitzer can you apply code highlighting for code? Just enclose those snippets with \`\`\`scala ... \`\`\` or \`\`\`sql ... \`\`\`

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
```diff
@@ -750,9 +768,19 @@ object TypeConverter {
 
     def targetTypeTag = implicitly[TypeTag[AnyRef]]
 
+    def cassandraOptionToAnyRef(cassandraOption: CassandraOption[_]) = {
```
jacek-lewandowski: Hmm... I'm still not convinced for this. This looks like messing different levels of encapsulation:
```scala
cassandraOption match {
  case CassandraOption.Value(x) => CassandraOption.Value(nestedConverter.convert(x).asInstanceOf[AnyRef])
  case x: CassandraOption => x
}
```
and make the writer take care of how it is handled.

Disregard this comment if you think it is a load of nonsense :)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/Unset.scala
```diff
@@ -0,0 +1,43 @@
+package com.datastax.spark.connector.types
```
jacek-lewandowski: I suppose the file should be named `CassandraOption.scala` rather than `Unset.scala`

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/BoundStatementBuilder.scala
```diff
@@ -38,12 +68,14 @@ private[connector] class BoundStatementBuilder[T](
       val converter = converters(i)
       val columnName = columnNames(i)
       val columnValue = converter.convert(buffer(i))
-      boundStatement.set(columnName, columnValue, CodecRegistryUtil.codecFor(columnTypes(i), columnValue))
```
jacek-lewandowski: Although I had something else on my mind when I commented this part of code for the last time, I'm ok with this solution.

My idea was more or else to create two methods with identical signature:
```scala
def bindColumnLegacy(stmt: RichBoundStatement, idx: Int, value: AnyRef) ...
```
```scala
def bindColumnV4(stmt: RichBoundStatement, idx: Int, value: AnyRef) ...
```
Both of them would include the appropriate implementation which would replace the code block below:
```scala
if (columnValue == Unset || (ignoreNulls && columnValue == null)) {
  maybeLeaveUnset(boundStatement, columnName)
} else {
  val codec = CodecRegistryUtil.codecFor(columnTypes(i), columnValue)
  boundStatement.set(columnName, columnValue, codec)
}
```
with appropriately inlined `maybeLeaveUnset`.

Then, you would have to do:
```scala
val bindColumn: (RichBoundStatement, Int, AnyRef) => Unit = protocolVersion match {
  case pv if pv.toInt <= ProtocolVersion.V3.toInt => bindColumnLegacy
  case _ => bindColumnV4
}
```
somewhere in main class block.

Ultimately, the it would look like:
```scala
    rowWriter.readColumnValues(row, buffer)
    var bytesCount = 0
    for (i <- 0 until columnNames.size) {
      val converter = converters(i)
      val columnName = columnNames(i)
      val columnValue = converter.convert(buffer(i))
      bindColumn(boundStatement, i, columnValue)
      val serializedValue = boundStatement.getBytesUnsafe(i)
      if (serializedValue != null) bytesCount += serializedValue.remaining()
    }
```
where `bindColumn` method would be preselected on whole `BoundStatementBuilder` creation. Semantically it would be identical to creation of two subclasses of `BoundStatementBuilder` - each with different definition of that `bindColumn` method.

jacek-lewandowski:
Apart from code highlighting and renaming one file I'm +1. Those other two comments are rather for discussion and I do not need them addressed there.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/BoundStatementBuilder.scala
RussellSpitzer: Hmm I like this approach too. I wanted to make it clear that what portion of the build was different. I have no strong feelings either way I can switch it to this if you like? I'm not sure it makes much of a different performance wise since Hotspot should optimize this loop regardless of how we write it (i think)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
RussellSpitzer: Hmmm, So my feeling here was that it is the converter's responsibility to take in Scala Objects and output something that the Driver can actually insert. I think it's ok to remove the Unset => Unset (and Unset Object) but the Null => java `null` seems to be important to me.

Basically I was trying to make sure we were as close to as possible the OptionToNull converter (prior to these mods)

RussellSpitzer:
@jacek-lewandowski Addressed all comments. I redid the bindValue function like you described, if you could take a quick peak I think thats it for this ticket

jacek-lewandowski:
+1, please squash before merging

RussellSpitzer:
Thanks will do

PR #895, opened 2015-12-10
Fix to IT tests to match Updated Spark

RussellSpitzer:
Also I missed a rename on defaultSparkConf
Spark output for pushdown filtering changed so we change our tests to
match the expected output.


bcantoni:
+1 in general, but looks like it compile stage still failing for src/it/scala/com/datastax/spark/connector/streaming/CheckpointStreamSpec.scala

RussellSpitzer:
retest this please

PR #896, opened 2015-12-10
SPARKC-297: Set Spark version to 1.6.0-rc1

jacek-lewandowski:
Also change the connector version to 1.6.0-M1 and ignore broken tests (to be addressed in other tickets)


pkolaczk:
+1

PR #898, opened 2015-12-14
SPARKC-231: Create C* Schema from DataFrame

RussellSpitzer:


alexliu68:
+1

etspaceman:
Do you have a timeframe as to when this will make it into one of the Maven version releases?

PR #900, opened 2015-12-18
Fix for sbt assembly

RussellSpitzer:
Ther move to Spark 1.6rc2 has caused a huge variety of
brokenness inside of the assembly. We have to explicitly drop
poms and such from various jars and use "last" merge strategies
on a variety of classes which are now getting included from multiple
dependencies.

Spark itself is guilty of adding

org/apache/spark/unused/UnusedStubClass.class

To every module


jacek-lewandowski:
+1, maybe we should add assembly goal as one of the tests we perform with each build?

PR #905, opened 2015-12-29
SPARKC-306: Remove Special Matching for Dataframe Writers

RussellSpitzer:
The type converters are fully capabible of handeling all of the
edge cases for DataFrame -> CassandraType conversions. We can use this
instead of the custom code previously in the SqlRowWriter class. In
addtiion, added test cases for the special matching previously done.

### DIFF
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/typeTests/InetTypeTest.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/typeTests/InetTypeTest.scala
@@ -1,8 +1,11 @@
 package com.datastax.spark.connector.rdd.typeTests
 
 import java.net.InetAddress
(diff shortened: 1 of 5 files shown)


alexliu68:
You can't simply remove them and only use spark connector testing here. Please test them in the latest DSE with spark sql thrift server

alexliu68:
Open a DSE ticket to run a regression test on it

RussellSpitzer:
@alexliu68 Will Do, This is actually to fix a few tickets in DSE

alexliu68:
My feeling is that it will break quite some testings in thrift server, for I remember it's fix for thrift server issues.

alexliu68:
test on the special data type for read and writing through beeline.

RussellSpitzer:
image

alexliu68:
+1, It looks like all those type converting is covered by TypeConverter, so it's ok to remove them from SqlRowWriter.

PR #909, opened 2016-01-05
Don't double resolve the file in DynamicCassandraPorts.

crakjie:
because the directory is resolve just once it's make an exception because directories are twice in the path but the second part is no created.

more detail [here](https://groups.google.com/a/lists.datastax.com/forum/#!topic/spark-connector-user/rokkNyJE3UQ)


jotsif:
I also have this issue and added a JIRA for it: 

https://datastax-oss.atlassian.net/browse/SPARKC-316

crakjie:
I have to register into this lira?

crakjie:
I never receive the confirmation mail from datastax Jira. Please can you add this PR to your lira ticket?

jotsif:
Yes, issues are handled by Datastax in their issue tracker.

Ok, I can add PR.

crakjie:
Thanks @jotsif ! :+1:

PR #914, opened 2016-01-13
SPARKC-299 Support append data to collection column for Spark SQL

alexliu68:
Support appending data to collection. to set pre-append per column base, 
<column_nam>.preappred should be set to true in table options.

### DIFF
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataSourceSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataSourceSpec.scala
@@ -47,6 +47,13 @@ class CassandraDataSourceSpec extends SparkCassandraITFlatSpecBase with Logging 
       },
 
       Future {
+        session.execute(s"CREATE TABLE $ks.test_list (a INT PRIMARY KEY, b List<INT>)")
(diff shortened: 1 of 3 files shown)


RussellSpitzer:
I kind of think the behavior should be slightly different. Rather than specifying column.preappend maybe we should have it were you can specify a columnname then the same word used in the scala api.

So
The equivalent of
```scala
listElements.saveToCassandra("ks", "collections_mod", SomeColumns("key", "lcol" append))
prependElements.saveToCassandra("ks", "collections_mod", SomeColumns("key", "lcol" prepend))
```

Would be something like 
```
listElements
  .write
  .option(Map("keyspace" -> "ks", "table" -> "collections_mod", "lcol.collectionBehavior" -> "append")
  .save()

and 

prependElements
  .write
  .option(Map("keyspace" -> "ks", "table" -> "collections_mod", "lcol.collectionBehavior" -> "prepend")
```
We could even drop collectionBehavior and just do `"lcol" -> "prepend"`

bcantoni:
Triage: @alexliu68 will rebase this off master (for 2.0).

Alex please also add user docs if applicable.

alexliu68:
Default is append, user only set column.preappend to true if it's not default behavior. 

It's slightly easier/simpler to use than column -> append. As user only need set it to true instead of prepend.

alexliu68:
replace it by https://github.com/datastax/spark-cassandra-connector/pull/1043

PR #919, opened 2016-01-26
SPARKC-321

jharnadeshmukh:
The routingKey is determined based on the column index in the first batch of rows sent during the save operation. Spark Cassandra connector shuffles the data around while saving the data resulting in column orders to be altered. This causes the job to fail when it's unable to find the routingKey column by that index. Instead of indexes, using columnNames should solve this issue.


RussellSpitzer:
I've been putting off looking at this because I'm not sure I understand the issue. Can you add a test case which would fail previously and would now succeed?

robertdkmarshall:
The change added a BOM to the file which is causing the CI failure.

bcantoni:
Triage: Closing per discussion in JIRA.

PR #921, opened 2016-01-29
use cached TypeConverter in SqlRowWriter

ilya-epifanov:
I have encountered a performance issue when 99% of the time was spent getting a type converter in SqlRowWriter, actually the problem was a bit deeper — I needed to make my batches per-replica instead of per-partition, but the point is that, still, cached converters weren't used.


RussellSpitzer:
retest this please

RussellSpitzer:
I'm +1 on this, i only wonder if we should fix this in previous versions. 

Thoughts @pkolaczk ?

RussellSpitzer:
Thanks @ilya-epifanov , I'm going to apply this as far back as b1.4 (seems to be the first instance)

RussellSpitzer:
I'm merging this here to save attribution, I'll have to do my own email for the previous branches (sorry)

PR #923, opened 2016-01-30
SPARKC-325: Provide an extension for New Driver Types

RussellSpitzer:
Allow for extending the `fromDriverRow` method of columnType. This
extension point should allow for custom cassandra types known by the
driver to be read.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/ColumnType.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/ColumnType.scala
@@ -5,7 +5,8 @@ import java.nio.ByteBuffer
 import java.util.{UUID, Date}
 
 import com.datastax.driver.core.{TupleType => DriverTupleType, UserType => DriverUserType, DataType}
-import com.datastax.spark.connector.util.{ReflectionUtil, Symbols}
+import com.datastax.spark.connector.util.{ConfigParameter, ReflectionUtil, Symbols}
+import org.apache.spark.SparkEnv
 
 import scala.collection.JavaConversions._
 import scala.reflect.runtime.universe._
@@ -45,7 +46,7 @@ trait ColumnType[T] extends Serializable {
 
 object ColumnType {
 
-  private val primitiveTypeMap = Map[DataType, ColumnType[_]](
+  private[connector] val primitiveTypeMap = Map[DataType, ColumnType[_]](
     DataType.text() -> TextType,
     DataType.ascii() -> AsciiType,
     DataType.varchar() -> VarCharType,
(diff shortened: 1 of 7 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/CustomDriverConverter.scala
```diff
@@ -0,0 +1,9 @@
+package com.datastax.spark.connector.types
+
+import com.datastax.driver.core.DataType
+
+trait CustomDriverConverter {
+
+  val fromDriverRowExtension:PartialFunction[(DataType, DataType.Name), ColumnType[_]]
```
jacek-lewandowski: formatting

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/ColumnType.scala
```diff
@@ -84,7 +92,8 @@ object ColumnType {
 
   def fromDriverType(dataType: DataType): ColumnType[_] = {
     val typeArgs = dataType.getTypeArguments.map(fromDriverType)
-    (dataType, dataType.getName) match {
+
+    val standardFromDriverRow: PartialFunction[(DataType, DataType.Name), ColumnType[_]] = {
```
jacek-lewandowski: Perhaps it would be better to predefine the partial function which takes what it takes now + type args. You wouldn't have create and chain them every time `fromDriverType` is invoked. I know it almost does not affect performance, so I don't insist.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/ColumnType.scala
```diff
@@ -68,6 +68,14 @@ object ColumnType {
     DataType.time() -> TimeType
   )
 
+  private lazy val customFromDriverRow: PartialFunction[(DataType, DataType.Name), ColumnType[_]] = {
+    Option(System.getenv("SCC_TYPE_CONVERTER"))
```
jacek-lewandowski: Shouldn't we get it from SparkConf?

RussellSpitzer: We don't have a spark conf here

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/ColumnType.scala
RussellSpitzer: I'm not sure what you mean by this

jacek-lewandowski:
Looks good. Will be nice if you add some test :)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/ColumnType.scala
jacek-lewandowski: More or less, I was thinking about simplifying this a little bit:

```scala
  private def typeArg(dataType: DataType, idx: Int) = fromDriverType(dataType.getTypeArguments.get(idx))

  private val standardFromDriverRow: PartialFunction[DataType, ColumnType[_]] = {
    case listType if listType.getName == DataType.Name.LIST => ListType(typeArg(listType, 0))
    case setType if setType.getName == DataType.Name.SET => SetType(typeArg(setType, 0))
    case mapType if mapType.getName == DataType.Name.MAP => MapType(typeArg(mapType, 0), typeArg(mapType, 1))
    case (userType: DriverUserType, _) => UserDefinedType(userType.getTypeName, fields(userType))
    case (tupleType: DriverTupleType, _) => TupleType(fields(tupleType): _*)
    case (dataType) => primitiveTypeMap(dataType)
  }


  def fromDriverType(dataType: DataType): ColumnType[_] = {
    val getColumnType: PartialFunction[DataType, ColumnType[_]] = customFromDriverRow orElse standardFromDriverRow
    getColumnType((dataType, dataType.getName))
  }
```

In particular, why this partial function needs to take both `DataType` and value of `name` property of it. As long as it was somewhere deep in connector code as a super private piece of code it was ok. But now we are going to have public interface which would be a little odd if it had redundant parameters. wdyt ?

jacek-lewandowski:
+1

PR #926, opened 2016-02-02
SPARKC-289: Fixed accessing multiple Cassandra clusters from Spark SQL

jacek-lewandowski:
SPARKC-289: Fixed accessing multiple Cassandra clusters from Spark SQL by adding the ability to define cluster/keyspace aliases

### DIFF
--- a/doc/14_data_frames.md
+++ b/doc/14_data_frames.md
@@ -39,23 +39,20 @@ To add these properties add keys to your `SparkConf` in the format
 
 Example Changing Cluster/Keyspace Level Properties
 ```scala
-val conf = new SparkConf()
-  .set("ClusterOne/spark.cassandra.input.split.size_in_mb","32")
-  .set("default:test/spark.cassandra.input.split.size_in_mb","128")
-
+sqlContext.setConf("ClusterOne/spark.cassandra.input.split.size_in_mb", "32")
+sqlContext.setConf("default:test/spark.cassandra.input.split.size_in_mb", "128")
 ...
-
 val df = sqlContext
   .read
   .format("org.apache.spark.sql.cassandra")
   .options(Map( "table" -> "words", "keyspace" -> "test"))
-  .load()// This DataFrame will use a spark.cassandra.input.size of 32
+  .load() // This DataFrame will use a spark.cassandra.input.size of 32
 
 val otherdf =  sqlContext
   .read
   .format("org.apache.spark.sql.cassandra")
   .options(Map( "table" -> "words", "keyspace" -> "test" , "cluster" -> "ClusterOne"))
-  .load()// This DataFrame will use a spark.cassandra.input.size of 128
+  .load() // This DataFrame will use a spark.cassandra.input.size of 128
 
 val lastdf = sqlContext
   .read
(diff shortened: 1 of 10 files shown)


RussellSpitzer:
I think I would rather we remove this functionality all together and require folks to just register their DataFrames that they create from different clusters. 

I think this capability is much less magical when going through
    csc.read.format.options(cluster, keyspace, table).load.registerTempTable("My Table")
Than adding a new interface to do so

jacek-lewandowski:
@RussellSpitzer can you take a look once again?

doc/14_data_frames.md
```diff
@@ -84,9 +82,48 @@ val df = sqlContext
...
+fox  40
+```
+
+There are also some helper methods which can make creating data frames easier. They can be accessed after importing 
+`org.apache.spark.sql.cassandra` package. In the following example, all the commands used to create a data frame are 
+equivalent:
+
+```scala
+import org.apache.spark.sql.cassandra._
+
```
RussellSpitzer: Maybe... too many options :)

doc/14_data_frames.md
```diff
@@ -146,6 +192,60 @@ df.write
...
+Since this setting is represented by `CassandraConnectorConf.ReadTimeoutParam` we can simply do:
+```scala
+options(CassandraConnectorConf.ReadTimeoutParam("7000"))
+```
+
+Each parameter, that is, each instance of `ConfigParameter` allows to invoke `apply` method with a single parameter. 
+That method returns a `Map[String, String]` (note that you need to use `options` instead of `option`) so setting 
+multiple parameters can be chained:
+```scala
+options(CassandraConnectorConf.ReadTimeoutParam("7000") ++ ReadConf.TaskMetricParam(true))
```
RussellSpitzer: I really like this

doc/14_data_frames.md
```diff
@@ -42,20 +42,18 @@ Example Changing Cluster/Keyspace Level Properties
 val conf = new SparkConf()
   .set("ClusterOne/spark.cassandra.input.split.size_in_mb","32")
```
RussellSpitzer: I think it would be nice to have a shortcut for this, I don't like having to remember the syntax here
Cluster:keyspace/key

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigParameter.scala
```diff
@@ -4,4 +4,7 @@ case class ConfigParameter[T](
   val name: String,
   val section: String,
   val default: T,
-  val description: String)
+  val description: String) extends DataFrameOption {
+
+  override val optionName = name.replaceAll("\\.", "\\_")
```
RussellSpitzer: I'm not sure this is required in the configuration, only in SQL options

RussellSpitzer:
```scala
map("key.that.has.dots" -> "value") 
```
is fine
```sql
"OPTIONS key_that_has_dots = vale" 
```

doc/14_data_frames.md
jacek-lewandowski: I like this suggestion :)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigParameter.scala
jacek-lewandowski: but this method is just for SQL options - for SparkConf options, there is still "name"

doc/14_data_frames.md
jacek-lewandowski: thinking more about this - I think I don't like this way of setting properties at all: `cluster:keyspace/propname`. I mean, these properties are very unusual and they cannot be set in spark-default.conf, because they do not follow `spark.*` pattern.
Since we want to fix multicluster, my idea is to put the cluster name and keyspace at the end of property name. This imho will be more natural:
- spark.cassandra.input.split.size_in_mb for everything
- spark.cassandra.input.split.size_in_mb.cluster for the whole cluster
- spark.cassandra.input.split.size_in_mb.cluster.keyspace for the whole keyspace

We could easily patch property name checker to such nuances and properties will be natural.
wdyt?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigParameter.scala
RussellSpitzer: Maybe then it should be val sqlOptionName ?

doc/14_data_frames.md
RussellSpitzer: I think it would be best if we had an implicit so you could just do 
SparkConf.setClusterConfig("ClusterName", OneOfOurConfigurationCaseClasses)

I really don't want to add more magic than we currently have in the configuration and we've already documented the previous pattern in a bunch of places.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/ConfigParameter.scala
jacek-lewandowski: ok

jacek-lewandowski:
@RussellSpitzer I realised few things:
- the description of setting cluster level properties had a bug - you cannot set them on `SparkConf` at all - you can only set them on `SQLContext`.
- setting the cluster level properties itself had a bug (probably) - once you set settings on `SQLContext` providing a default cluster, they were not applied to named clusters - my understanding is that once the setting is not defined explicitly for the cluster X, we try to use the setting for the default cluster
- since all the cluster/keyspace level settings are `SQLContext` only, I realised that there is no point in changing anything in this ticket. Having namespaces for different cluster in `SparkConf` seems to be whole another story.

jacek-lewandowski:
I've got +1 from @RussellSpitzer on HipChat

PR #927, opened 2016-02-03
SPARKC-326: Support for reading Materialized Views

RussellSpitzer:
A new member is added to TableDef isView:Boolean which allows us to
specify whether the shchema represented by the TableDef is a
MaterializedView. We could also refactor the TableDef into multiple
classes but I found this to take considerably more refactoring than just
adding the single element. This allows all 1.5 TableDef code to be
compatible and supports future View Development.

Added a small trap for those trying to write to a Materialized View to
avoid getting a Java Driver Exeception.


jacek-lewandowski:
+1

PR #937, opened 2016-02-20
Rename variable throughputMiBPS as throughputMBPS

noorul:
This will make it clear that we are dealing with Mega Bytes per second.


noorul:
Not sure why travis did not kick in

noorul:
I think this patch is not required.

PR #941, opened 2016-02-26
SPARKC-338-b1.4: Change DF ErrorIfExists Message

RussellSpitzer:
The current message makes it seem as if there is no way to insert to a
C* table if there is data within it. This patch changes the message so
it is clear that the error is because of the save mode and suggests a
workaround if Append behavior was expected.


jtgrabowski:
+1

RussellSpitzer:
Merged forward

PR #942, opened 2016-02-29
SPARKC-341 change index names to 'handled' names

jtgrabowski:


RussellSpitzer:
+1, old +1 lost on the old commit :)

PR #944, opened 2016-03-17
[SPARKC-346] - Nested Optional Case Class can be save as UDT

crakjie:
Allow to save nested optional case class in udt.

Some(caseClass) is save like a caseClass.
None is save like null.

Is a quick fix I thinks it's can be donne in a better way.

ps : Is may be a good idea to replace run time reflection by compile time macro.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,4 @@
+ * Nested Optional Case Class can be save as UDT (SPARKC-346)
  * Merged Java API into main module (SPARKC-335)
  * Upgraded to Spark 1.6.1 (SPARKC-344)
  * Removed the ability to specify cluster alias directly and added some helper methods which make it easier to configure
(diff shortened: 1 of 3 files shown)


RussellSpitzer:
@pkolaczk Ping

crakjie:
Yes?

pkolaczk:
Yes, it is a good idea to replace run time reflection with compile time macro, but one problem with it is that Java API would still need to rely on runtime reflection, because Java has no macros :(

pkolaczk:
+1

pkolaczk:
All tests green. Thank you! :)

crakjie:
I m working on macro that give the same fonctionality, but just on case classes. 
We have other project that not running on spark and I too used to having this fonctionality :p

Know I hop the code in this PR is clean enough. Like I said to rusell I m not used to work with reflection.

PR #945, opened 2016-03-17
Don't double resolve the file.

crakjie:
because the directory is resolve just once it's make an exception


jtgrabowski:
Thanks! :)

PR #946, opened 2016-03-20
[SPARKC-342] support Joda time for LocalDate type

maasg:
Adds type conversion support for `org.joda.time.DateTime` to `com.datastax.driver.core.LocalDate`. Solves  [SPARKC-342](https://datastax-oss.atlassian.net/browse/SPARKC-342)

@RussellSpitzer  @pkolaczk for review

PR #947, opened 2016-03-21
Sparkc 342 support joda time for date type

maasg:
As requested by @pkolaczk (https://datastax-oss.atlassian.net/browse/SPARKC-342), merging this patch also into b1.5  (see #946)


jtgrabowski:
Thanks!

PR #950, opened 2016-03-25
 SPARKC-348 accept predicates with indexed partition columns

jtgrabowski:
Preconditions with indexed columns that belong to partition key but do
not form the whole partition key generate failures.
This kind of queries should be executed as a query with index column
specified, not discarded as a query with incomplete partition key.

This changes CassandraRDDPartitioner#containsPartitionKey so that it
does not throw exception whenever the partition key is incomplete and
all precodition colums are indexed. Other cases for
CassandraRDDPartitioner#containsPartitionKey are not changed.

### DIFF
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLSpec.scala
@@ -275,7 +275,24 @@ class CassandraSQLSpec extends SparkCassandraITFlatSpecBase {
           Future {
             session.execute(s"CREATE TABLE $ks1.tuple_test1 (id int PRIMARY KEY, t Tuple<text, int>)")
             session.execute(s"INSERT INTO $ks1.tuple_test1 (id, t) VALUES (1, ('xyz', 2))")
-          }
+          },
+
+          Future {
+            session.execute(
+              s"""
+                 |CREATE TABLE IF NOT EXISTS $ks1.index_test (
+                 |  ipk1 int,
+                 |  pk2 int,
+                 |  id1 int,
+                 |  d2 int,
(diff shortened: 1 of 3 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitioner.scala
```diff
@@ -109,14 +107,17 @@ class CassandraRDDPartitioner[V, T <: Token[V]](
...
     }.toSet
 
-    if (whereColumns.nonEmpty && whereColumns.size < pk.size) {
+    val primaryKeyComplete = whereColumns.nonEmpty && whereColumns.size == pk.size
+    val whereColumnsAllIndexed = whereColumns.forall(tableDef.isIndexed)
+
+    if (!primaryKeyComplete && !whereColumnsAllIndexed) {
       val missing = pk -- whereColumns
       throw new UnsupportedOperationException(
         s"Partition key predicate must include all partition key columns. Missing columns: ${missing.mkString(",")}"
```
RussellSpitzer: probably want to change this message to also include "Or all where predicates must be indexed"

RussellSpitzer:
Tiny suggestion for changing the error message and then +1

jtgrabowski:
rebased on current master

RussellSpitzer:
There is going to be a slight conflict here with my patch for SPARKC-376, we should figure out how we want to solve that ...

jtgrabowski:
Rebased on current master.

Hey @RussellSpitzer , as we talked, I looked into possible conflict with SPARKC-376 and I believe there is none . The `containsPartitionKey` method's evaluated results are not affected by this PR. What's changed is that the exception `UnsupportedOperationException` is not thrown when we get not complete partition key.
wdyt, are w good to go?

RussellSpitzer:
+1

PR #951, opened 2016-03-25
SPARKC-330: Implement org.apache.spark.Partitioner

RussellSpitzer:
Currently once a CassandraTableScanRDD is created there is no mapping back to the partitioning that created it. For limited instances we can can use the partitioning that created the TableScanRDD to remove the need for shuffles during some spark joins. 

To this purpose we enhance the .keyBy method of C*TableScanRDD to automatically create a Spark Partitioner class for the tablescan RDD if it is valid. This partitioner can then be shared with other RDD's with the same key type and partitioning. When joined these new table will not require a shuffle

### DIFF
--- a/doc/16_partitioning.md
+++ b/doc/16_partitioning.md
@@ -1,0 +1,328 @@
+# Documentation
+
+## Spark Cassandra Connector and Spark Partitioners
+
+### What are Spark Partitioners
+
+Spark RDDs may have an element known as a [Partitioner](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/Partitioner.scala);
+these special classes are used during shuffles and co-grouping events (joins, groupBy) to map data 
+to the correct Spark Partition. When one isn't specified Spark will usually repartition using 
+a `HashPartitioner`. A repartition requires taking every value and moving it to the partition 
+specified by the given partitioner. This is the reason that many keyed operations are such expensive 
+operations.
+
+### What is a Cassandra Partitioner
+
+Cassandra also has a notion of a partitioner. The Cassandra partitioner takes the partition key of
+a CQL Row and uses it to determine what node in the Cassandra cluster should host that data. The 
+partitioner generates a `Token` which directly maps to the `TokenRange` of the CassandraCluster.
+Each node in the Cassandra Cluster is responsible for a specific section of the full `TokenRange`.
+
+For more information on Cassandra Data Modeling check out the resources on 
+[DataStax Academy](https://academy.datastax.com/courses/ds220-data-modeling).
+
+### How to take advantage of Cassandra Partitioning in Spark
+
+The Spark Cassandra Connector now implements a `CassandraPartitioner` for specific RDDs when they
+have been keyed using `keyBy`. This can only be done when a `CassandraTableScanRDD` has `keyBy` 
+called and is keyed by the partition keys of the underlying table. The key may contain other
+elements of the table but the partition key is required to build a `CassandraPartitioner`
+
+*All of these examples are run with Spark in Local Mode and --driver-memory 4g*
+
+#### Example Schema
+
+
+```sql
+CREATE KEYSPACE doc_example WITH replication = 
+    {'class': 'SimpleStrategy', 'replication_factor': '1'}  AND durable_writes = true;
+
+CREATE TABLE doc_example.purchases (
+    userid int,
+    purchaseid int,
+    objectid text,
+    amount int,
+    PRIMARY KEY (userid, purchaseid, objectid) //Partition Key "userid"
+);
+
+CREATE TABLE doc_example.users (
+    userid int PRIMARY KEY, // Partition key "userid"
+    name text,
+    zipcode int
+);
+```
+
+For a data generation sample, jump see [here](#data-generator).
+
+#### Keying a Table with the Partition Key Produces an RDD with a Partitioner
+
+```scala
+import com.datastax.spark.connector._
+
+val ks = "doc_example"
+val rdd = { sc.cassandraTable[(String, Int)](ks, "users")
+  .select("name" as "_1", "zipcode" as "_2", "userid")
+  .keyBy[Tuple1[Int]]("userid")
+}
+
+//16/04/08 11:19:52 DEBUG CassandraPartitioner: Building Partitioner with mapping
+//Vector((userid,userid))
+//for table TableDef(doc_example,users,ArrayBuffer(ColumnDef(userid,PartitionKeyColumn,IntType)),ArrayBuffer(),ArrayBuffer(ColumnDef(name,RegularColumn,VarCharType), ColumnDef(zipcode,RegularColumn,IntType)),List(),false)
+//16/04/08 11:19:52 DEBUG CassandraTableScanRDD: Made partitioner Some(com.datastax.spark.connector.rdd.partitioner.CassandraPartitioner@6709eccf) for CassandraTableScanRDD[5] at RDD at CassandraRDD.scala:15
+//16/04/08 11:19:52 DEBUG CassandraTableScanRDD: Assigning new Partitioner com.datastax.spark.connector.rdd.partitioner.CassandraPartitioner@6709eccf
+
+rdd.partitioner
+//res3: Option[org.apache.spark.Partitioner] = Some(com.datastax.spark.connector.rdd.partitioner.CassandraPartitioner@94515d3e)
+```
+
+#### Reading a table into a (K,V) type or Not Including the Partition Key will not Create a Partitioner
+
+```scala
+import com.datastax.spark.connector._
+
+val ks = "doc_example"
+val rdd1 = sc.cassandraTable[(Int,Int)](ks, "users")
+rdd1.partitioner
+//res8: Option[org.apache.spark.Partitioner] = None
+
+
+val rdd2 = { sc.cassandraTable[(Int, String)](ks, "users")
+  .select("name" as "_2", "zipcode" as "_1", "userid")
+  .keyBy[Tuple1[Int]]("zipcode")
+}
+rdd2.partitioner
+//res11: Option[org.apache.spark.Partitioner] = None
+```
+
+### Capabilities and limitations of CassandraPartitioner
+
+Having a partitioner allow for groupByKey operations to no longer require a shuffle. Previously
+`spanByKey` allowed Spark to take advantage of the natural ordering of C* tables. The new partitioner
+allows this to go a step further and group on any combination of the primary key which also contains
+the partition key.
+
+#### Grouping/Reducing on out of order Clustering keys
+
+```scala
+import com.datastax.spark.connector._
+val ks = "doc_example"
+val rdd = { sc.cassandraTable[(Int, Int)](ks, "purchases")
+  .select("purchaseid" as "_1", "amount" as "_2", "userid", "objectid")
+  .keyBy[(Int, String)]("userid","objectid")
+}
+
+rdd.groupByKey.toDebugString
+//res28: String =
+//(4) MapPartitionsRDD[50] at groupByKey at <console>:43 []
+// |  CassandraTableScanRDD[41] at RDD at CassandraRDD.scala:15 []
+//
+// Note that there are no shuffles present in this tree
+rdd.groupByKey.count
+// 16/04/08 16:23:42 INFO DAGScheduler: Job 0 finished: count at <console>:35, took 34.171709 s
+```
+
+Doing the same thing without a partitioner requires a shuffle:
+
+```scala
+val ks = "doc_example"
+
+//Empty map will remove the partitioner
+val rdd = { sc.cassandraTable[(Int, Int)](ks, "purchases")
+  .select("purchaseid" as "_1", "amount" as "_2", "userid", "objectid")
+  .keyBy[(Int, String)]("userid","objectid")
+}.map( x => x)
+
+rdd.partitioner
+//res2: Option[org.apache.spark.Partitioner] = None
+
+rdd.groupByKey.toDebugString
+//res3: String =
+//(4) ShuffledRDD[11] at groupByKey at <console>:35 []
+// +-(4) MapPartitionsRDD[10] at map at <console>:35 []
+//    |  CassandraTableScanRDD[9] at RDD at CassandraRDD.scala:15 []
+
+rdd.groupByKey.count
+//16/04/08 16:28:23 INFO DAGScheduler: Job 1 finished: count at <console>:35, took 46.550789 s
+```
+
+In this example the partitioner example takes 70% of the time of the partitioned one.  This is even
+with the entire load taking place on a single machine without any network traffic. When 
+shuffles have to spill to disk the difference in times can become even greater. With very
+large examples we've seen the cost of the shuffle even double the total compute time. 
+
+*Limitations*
+
+This can only be applied to RDDs that are keyed with their partition key, so it can not be used
+to group or reduce on a generic Cassandra column. As long as you are grouping/reducing within a C*
+partition this approach should save significant time.
+
+
+#### Joining to Cassandra RDDs from non Cassandra RDDs
+
+When joining two RDDs they are both required to be partitioned before the join can occur. Previously
+this meant that when joining against a Cassandra both the Cassandra RDD and the RDD to be joined
+would need to be shuffled. Now if the Cassandra RDD is keyed on the partition key you can join 
+without the Cassandra RDD being shuffled.
+
+```scala
+import com.datastax.spark.connector._
+
+val ks = "doc_example"
+val rdd = { sc.cassandraTable[(String, Int)](ks, "users")
+  .select("name" as "_1", "zipcode" as "_2", "userid")
+  .keyBy[Tuple1[Int]]("userid")
+}
+
+val joinrdd = sc.parallelize(1 to 10000).map(x => (Tuple1(x),x) ).join(rdd)
+joinrdd.toDebugString
+//(1) MapPartitionsRDD[26] at join at <console>:35 []
+// |  MapPartitionsRDD[25] at join at <console>:35 []
+// |  CoGroupedRDD[24] at join at <console>:35 []
+// +-(8) MapPartitionsRDD[23] at map at <console>:35 []
+// |  |  ParallelCollectionRDD[22] at parallelize at <console>:35 []
+// |  CassandraTableScanRDD[16] at RDD at CassandraRDD.scala:15 []
+joinrdd.count
+//16/04/08 17:25:47 INFO DAGScheduler: Job 14 finished: count at <console>:37, took 5.339490 s
+
+//Use an empty map to drop the partitioner
+val rddnopart = { sc.cassandraTable[(String, Int)](ks, "users")
+                  .select("name" as "_1", "zipcode" as "_2", "userid")
+                  .keyBy[Tuple1[Int]]("userid").map( x => x)
+}
+
+val joinnopart = sc.parallelize(1 to 10000).map(x => (Tuple1(x),x) ).join(rddnopart)
+joinnopart.toDebugString
+//res18: String =
+//(8) MapPartitionsRDD[73] at join at <console>:36 []
+// |  MapPartitionsRDD[72] at join at <console>:36 []
+// |  CoGroupedRDD[71] at join at <console>:36 []
+// +-(8) MapPartitionsRDD[70] at map at <console>:36 []
+// |  |  ParallelCollectionRDD[69] at parallelize at <console>:36 []
+// +-(1) MapPartitionsRDD[68] at map at <console>:34 []
+//    |  CassandraTableScanRDD[61] at RDD at CassandraRDD.scala:15 []
+
+joinnopart.count
+//16/04/08 17:30:04 INFO DAGScheduler: Job 19 finished: count at <console>:37, took 6.308165 s
+```
+
+So again we can save the time of a shuffle if we are joining on a partition key of a C* table. The
+advantages here scale as well, the larger the data to be joined the greater the advantage to be 
+gained.
+
+*Limitations*
+
+Again it is important to note that this is only possible when using the Partition key as the key
+of the RDD.
+
+#### Joining to Cassandra RDDs on Common Partition Keys
+
+One area that can benefit greatly from the Partitioner is joins between Cassandra tables on a 
+common partition key. This previously would requiring both RDDs but now requires no shuffles at
+all. To use this both RDDs must have the *same partitioner*. A helper method `applyPartitionerFrom`
+has been added to make it easy to share a partitioner between Cassandra RDDs.
+
+```scala
+val ks = "doc_example"
+val rdd1 = { sc.cassandraTable[(Int, Int, String)](ks, "purchases")
+  .select("purchaseid" as "_1", "amount" as "_2", "objectid" as "_3", "userid")
+  .keyBy[Tuple1[Int]]("userid")
+}
+
+val rdd2 = { sc.cassandraTable[(String, Int)](ks, "users")
+  .select("name" as "_1", "zipcode" as "_2", "userid")
+  .keyBy[Tuple1[Int]]("userid")
+}.applyPartitionerFrom(rdd1) // Assigns the partitioner from the first rdd to this one
+
+val joinRDD = rdd1.join(rdd2)
+joinRDD.toDebugString
+//res37: String =
+//(1) MapPartitionsRDD[123] at join at <console>:36 []
+// |  MapPartitionsRDD[122] at join at <console>:36 []
+// |  CoGroupedRDD[121] at join at <console>:36 []
+// |  CassandraTableScanRDD[115] at RDD at CassandraRDD.scala:15 []
+// |  CassandraTableScanRDD[120] at RDD at CassandraRDD.scala:15 []
+
+joinRDD.count
+//16/04/08 17:53:45 INFO DAGScheduler: Job 24 finished: count at <console>:39, took 27.583355 s
+
+//Unpartitioned join
+val rdd1nopart = { sc.cassandraTable[(Int, Int, String)](ks, "purchases")
+  .select("purchaseid" as "_1", "amount" as "_2", "objectid" as "_3", "userid")
+  .keyBy[Tuple1[Int]]("userid")
+}.map(x => x)
+
+val rdd2nopart = { sc.cassandraTable[(String, Int)](ks, "users")
+  .select("name" as "_1", "zipcode" as "_2", "userid")
+  .keyBy[Tuple1[Int]]("userid")
+}.map(x => x)
+
+val joinnopart = rdd1nopart.join(rdd2nopart)
+joinnopart.toDebugString
+//   res41: String =
+//   (4) MapPartitionsRDD[136] at join at <console>:36 []
+//    |  MapPartitionsRDD[135] at join at <console>:36 []
+//    |  CoGroupedRDD[134] at join at <console>:36 []
+//    +-(1) MapPartitionsRDD[128] at map at <console>:35 []
+//    |  |  CassandraTableScanRDD[127] at RDD at CassandraRDD.scala:15 []
+//    +-(4) MapPartitionsRDD[133] at map at <console>:35 []
+//       |  CassandraTableScanRDD[132] at RDD at CassandraRDD.scala:15 []
+
+joinnopart.count
+//16/04/08 17:54:58 INFO DAGScheduler: Job 25 finished: count at <console>:39, took 40.040341 s
+```
+
+Again we can save a significant portion of time by eliminating the need for a shuffle.
+
+
+### Caveats
+The partitioning mechanism is very sensitive to the underlying C* tables partitioning so this is
+not a generic solution to all joins and groupBys. In addition it is important to note 
+
+`applyPartitionerFrom` will copy the partitioning exactly from the host RDD to the target. 
+This means it will only work if the key portion of both RDDs are the same. When applied, the host
+partitioning takes over. This means it often makes sense to apply the partitioner from the RDD with
+more partitions to get proper parallelism.
+
+To key and apply a partitioner at the same time you can use `keyAndApplyPartitionerFrom`. These 
+APIs are documented in `CassandraTableScanRDDFunctions` and are provided as implicits on 
+`CassandraTableScanRDD`s.
+
+Due to the way RDDs are constructed, the CassandraPartitioner cannot be applied without 
+an explicit `KeyBy` call.
+
+The class of the Key must be a true class and not a primitive. The Partitioner uses the a `RowWriter`
+like the `saveToCassandra` method to convert key data to a C* token. This `RowWriter` cannot be 
+built for primitives which is why the above examples use `Tuple1` so frequently.
+
+This functionality does not currently function in DataFrames.
+
+
+###Data Generator
+```scala
+
+import com.datastax.spark.connector._
+import java.util.Random
+val ks = "doc_example"
+val r = new Random(100)
+val zipcodes = (1 to 100).map(_ => r.nextInt(99999)).distinct
+val objects = Seq("pepper", "lemon", "pear", "squid", "beet", "iron", "grass", "axe", "grape")
+
+case class RandomListSelector[T](list: Seq[T]) {
+    val r = { new Random() }
+    def next(): T = list(r.nextInt(list.length))
+}
+
+val randomObject = RandomListSelector(objects)
+val randomZipCode = RandomListSelector(zipcodes)
+
+//About 25 Seconds on my MBP 
+sc.parallelize(1 to 1000000).map(x => 
+    (x, s"User $x", randomZipCode.next)
+  ).saveToCassandra(ks, "users")
+ 
+//About 60 Seconds on my MBP
+sc.parallelize(1 to 1000000).flatMap(x => 
+    (1 to 10).map( p => (x, p, randomObject.next, p))
+  ).saveToCassandra(ks, "purchases")
+
+```
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/doc/DocExamples.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/doc/DocExamples.scala
@@ -1,5 +1,7 @@
 package com.datastax.spark.connector.doc
 
+import java.util.Random
+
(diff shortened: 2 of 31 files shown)


RussellSpitzer:
retest this please

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
```diff
@@ -0,0 +1,323 @@
...
+  lazy val ranges = partitionsAndRanges.map(x =>
+    (x._2, x._3.map( range =>
+      (range.start, range.end, range.isWrapAround))))
+
+
+  /**
+    * Generates a list of the CassandraPartitions to be used in the TableScanRDD and a list
+    * of the TokenRanges used to craft those partitions along with their indexes.
+    */
+  lazy val partitionsAndRanges: Seq[(CassandraPartition, Int, Seq[TokenRange])] = {
```
jacek-lewandowski: Why `Seq[(CassandraPartition, Int, Seq[TokenRange])]`? the second component is index which is also present in `CassandraPartition`

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableScanRDD.scala
```diff
@@ -119,30 +192,76 @@ class CassandraTableScanRDD[R] private[connector](
...
+      CassandraRDDPartitionGenerator(connector, tableDef, splitCount, splitSize)
+    }
+  }
+
+  @transient override val partitioner = overridePartitioner
+
   override def getPartitions: Array[Partition] = {
     verify() // let's fail fast
-    val partitioner = CassandraRDDPartitioner(connector, tableDef, splitCount, splitSize)
-    val partitions = partitioner.partitions(where)
```
jacek-lewandowski: Is `where` clause still considered?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
```diff
@@ -0,0 +1,323 @@
...
+  /**
+    * Since Driver TokenRange objects are not serializable we replicate the
+    * logic for TokenRange.contains(Token) here.
+    */
+  def indexOfPartitionContaining(token: com.datastax.driver.core.Token): Int = {
+    val tokenValue = token.getValue.asInstanceOf[V]
+    indexedRanges.find { case (index, ranges) =>
+      ranges.exists { case (start, end, wrap) =>
+        val (startValue, endValue) = (start.value, end.value)
+        if (end == minTokenValue && tO.lt(startValue, tokenValue)) {
```
jacek-lewandowski: min or max here?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
```diff
@@ -0,0 +1,323 @@
...
+      }
+      case other => throw new IllegalArgumentException(s"Couldn't determine the key from object $other")
+    }
+  }
+
+  /**
+    * Since Driver TokenRange objects are not serializable we replicate the
+    * logic for TokenRange.contains(Token) here.
+    */
+  def indexOfPartitionContaining(token: com.datastax.driver.core.Token): Int = {
```
jacek-lewandowski: is `getPartition` going to be called frequently?, like for each element?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableScanRDD.scala
```diff
@@ -119,30 +192,76 @@ class CassandraTableScanRDD[R] private[connector](
...
+    classtag: ClassTag[K],
+    rrf: RowReaderFactory[K],
+    rwf: RowWriterFactory[K]): CassandraTableScanRDD[(K, R)] =
     keyBy(SomeColumns(columns: _*))
 
   /** Extracts a key of the given class from all the available columns.
+    *
     * @see `keyBy(ColumnSelector)` */
-  def keyBy[K : RowReaderFactory]: CassandraTableScanRDD[(K, R)] =
+  def keyBy[K]()(implicit
```
jacek-lewandowski: side question - what's the point of this method?

jacek-lewandowski:
Have a general question - would it be possible to extract the ability to generate routing key to some trait so that both `RowReader` and `RowWriter` can use it? In the places where row reader is used we read Cassandra row and then we create some target object. Then we use row writer to convert it back to Cassandra row compatible values in order to generate the routing key, and then - the token.

For example - `keyBy` requires both `RowReaderFactory[K]` and `RowWriterFactory[K]`, which means - to convert Cassandra row to K and then from K to Cassandra row. Why don't we just get the values directly from Cassandra row to generate the routing key?

wdyt?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
```diff
@@ -0,0 +1,323 @@
...
+
+  /**
+    * Adjust the partition key string in only of the Token clauses to be used with this scan
+    * incase the names differ from the inputPartitions
+    */
+  val partitions: Seq[CassandraPartition] = {
+    val pk = tableDef.partitionKey.map(_.columnName).map(quote).mkString(", ")
+    for (partition <- inputPartitions) yield {
+      val localTokenRanges = for (cqlTokenRange <- partition.tokenRanges) yield {
+        CqlTokenRange(cqlTokenRange.cql.changePartitionKey(pk), cqlTokenRange.values: _*)
```
pkolaczk: `changePartitionKey` -> `withPartitionKey`
The general naming rule is:
Use verbs for methods that may have a side effect, e.g. that modify the subject. Use nouns for pure functions, e.g. that return a modified copy of the subject, but leave the original intact.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
```diff
@@ -0,0 +1,323 @@
...
+    }
+  }
+
+  /**
+    * Since Driver TokenRange objects are not serializable we replicate the
+    * logic for TokenRange.contains(Token) here.
+    */
+  def indexOfPartitionContaining(token: com.datastax.driver.core.Token): Int = {
+    val tokenValue = token.getValue.asInstanceOf[V]
+    indexedRanges.find { case (index, ranges) =>
```
pkolaczk: Do I understand this right that this may get called for every key? If so, then it should be as cheap as possible. The number of tokens in the token ring may be very high when virtual nodes are enabled. You may be scanning several thousands of ranges here for bigger clusters.

I can see at least two better ways of finding the range:
1. build an interval tree (once) from ranges and then lookup a range in O(log n) time
2. use hashing-like technique: you can divide the token ring into 2^n buckets of equal size, close or little smaller than the average size of the token range and assign ranges to appropriate buckets. Because buckets are equal size, finding the bucket index for a token is O(1) and very fast (I guess a single bitwise shift-right) and then the remaining operation is just scanning a few ranges assigned to that bucket (which may be stored in a vector for faster scanning). 
Overall you'd get O(1) lookup and just two or three random memory accesses with this technique (think: nanoseconds fast).

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
```diff
@@ -0,0 +1,323 @@
...
+    inputPartitions: Seq[CassandraPartition] = partitions,
+    indexedRanges: Seq[(Int, Seq[(Token[V], Token[V], Boolean)])] = indexedRanges,
+    tableDef: TableDef = tableDef,
+    connector: CassandraConnector = connector,
+    tokenBounds: (Token[V], Token[V]) = tokenBounds):CassandraRDDPartitioner[Key, V] = {
+
+    new CassandraRDDPartitioner[Key, V](partitions, indexedRanges, tableDef, connector, tokenBounds)
+  }
+
+  implicit val tO = tokenBounds._1.ord
```
pkolaczk: `tO`  looks too much like `t0`. Maybe think of a slightly longer and more descriptive name?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
pkolaczk: you can replace the if ladder with a logic expression

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
```diff
@@ -0,0 +1,323 @@
...
+  * Will take objects of type Key and determine given the token ragnes in `indexedRanges`
+  * which range the Key would belong in given the C* schema in `TableDef`
+  *
+  * Under the hood uses a bound statement to generate routing keys which are then
+  * used the driver's internal token factories to determine the token for the
+  * routing key.
+  */
+private[connector] class CassandraRDDPartitioner[Key : ClassTag, V](
+  inputPartitions: Seq[CassandraPartition],
+  val indexedRanges: Seq[(Int, Seq[(Token[V], Token[V], Boolean)])],
```
pkolaczk: why do we need both inputPartitions and indexedRages?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableScanRDD.scala
```diff
@@ -111,7 +115,76 @@ class CassandraTableScanRDD[R] private[connector](
...
+
+  /**
+    * Internal method for assigning a partitioner to this RDD, this lacks type safety checks for
+    * the Partitioner of type [K]. End users will use the implicit provided in
+    * [[CassandraTableScanPairRDDFunctions]]
+    */
+  private[connector] def withPartitioner[K, V](
+    partitioner: Option[Partitioner]): CassandraTableScanRDD[R] = {
+
+    val cassPart = partitioner match {
```
pkolaczk: unused

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableScanRDD.scala
```diff
@@ -111,7 +115,76 @@ class CassandraTableScanRDD[R] private[connector](
...
+      sc = sc,
+      connector = connector,
+      keyspaceName = keyspaceName,
+      tableName = tableName,
+      columnNames = columnNames,
+      where = where,
+      limit = limit,
+      clusteringOrder = clusteringOrder,
+      readConf = readConf,
+      overridePartitioner = partitioner)
```
pkolaczk: `partitioner` -> `cassPart` ?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
```diff
@@ -0,0 +1,323 @@
...
+  val tableDef: TableDef,
+  connector: CassandraConnector,
+  tokenBounds: (Token[V], Token[V]))(
+implicit
+  @transient rwf:RowWriterFactory[Key]) extends Partitioner with Logging {
+
+  /**
+    * Adjust the partition key string in only of the Token clauses to be used with this scan
+    * incase the names differ from the inputPartitions
+    */
```
pkolaczk: What can be the reason for this "incase"?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
```diff
@@ -0,0 +1,323 @@
...
+    new CassandraRDDPartitioner[Key, V](partitions, indexedRanges, tableDef, connector, tokenBounds)
+  }
+
+  implicit val tO = tokenBounds._1.ord
+
+  val extractedPartitions = partitions
+    .map(cqlPartition => (cqlPartition.index, cqlPartition.tokenRanges))
+
+  val (minToken, maxToken) = tokenBounds
+  val minTokenValue = minToken.value
```
pkolaczk: private ?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
```diff
@@ -0,0 +1,323 @@
...
+
+    new CassandraRDDPartitioner[Key, V](partitions, indexedRanges, tableDef, connector, tokenBounds)
+  }
+
+  implicit val tO = tokenBounds._1.ord
+
+  val extractedPartitions = partitions
+    .map(cqlPartition => (cqlPartition.index, cqlPartition.tokenRanges))
+
+  val (minToken, maxToken) = tokenBounds
```
pkolaczk: private ?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
pkolaczk: @jacek-lewandowski I think it is correct, minTokenValue is special

RussellSpitzer: I just copied all the driver code on this one so I think this is right.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
RussellSpitzer: This would be called once for every object on the partitioning a new rdd or joining of an unpartitioned RDD with this RDD. If they share the same partitioner it is not a big deal since we know getPartitionKey(x) == getPartitionKey(x) on both rdds so we know they are local in the first place. 

I agree this should be optimized

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
RussellSpitzer: This should most likely be refactored, The Index is not really needed (i'm just paranoid about the Ordering. The real issue here is that going from TokenRanges to CassandraPartitions also requires RowCount and Replicas. This means we can't just go from ranges to CassandraPartitions, The Cassandra partitions lose all of their type and have varargs which makes them more annoying to convert back to Ranges. 

Maybe the right thing to do is to make a super class PartitionRangeInformation[V](Seq[(V, V , Boolean)], RowCount, Replicas) ? This could be converted directly to a CassandraPartition

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableScanRDD.scala
RussellSpitzer: That explains some bugs I saw last night :(

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
RussellSpitzer: Yeah i mentioned this above with Jacek. The dependency between ranges are neccessary but not sufficent to build CassandraPartitions, CassandraPartitions lack the type information to recreate Ranges. I'll work on making a better super class  for Ranges that can go to CassandraPartitions

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
RussellSpitzer: Imagine Tables 1(x Key, a, b) and 2(y Key, a ,b) . The cql partitions for 1 cannot be used with 2 because they refer to toke(x) and not tokey(y) even though they are identical in composition.  This will probably be fixed if i switch to just passing a RangeClass that makes CassandraPartitions so you don't have to pass two things which are so related

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
RussellSpitzer: Refactored all this.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
RussellSpitzer: Fixed this as well in refactoring

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
RussellSpitzer: This low level modification is no longer needed, CassandraPartitions are generated based on the table def of the partitioner

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
RussellSpitzer: changed

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
RussellSpitzer: Encapsulated and privated

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
RussellSpitzer: privated just about everything.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
RussellSpitzer: Condensed into single logic expression

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionGenerator.scala
RussellSpitzer: Refactored into PartitioningMetaData

pkolaczk:
Weird. All tests repeatedly pass for me locally. @RussellSpitzer can you test it?

pkolaczk:
retest this please

pkolaczk:
+1

pkolaczk:
@RussellSpitzer I think the squashed commit message needs improvement. We should write a better overall description for all of the commits, explaining why these changes were needed. It is quite hard to figure that out from each individual message (and after squashing you even can't see the changes introduced by each individual commit).

RussellSpitzer:
Yep i was going to pare down that message after I finish writing docs and examples. Should have that all up today.

RussellSpitzer:
Reformed commit message and added a first draft of docs @jtgrabowski

doc/16_partitioning.md
```diff
@@ -0,0 +1,328 @@
+# Documentation
+
+## Spark Cassandra Connector and Spark Partitioners
+
+### What are Spark Partitioners
+
+Spark RDDs may have an element known as a [Partitioner](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/Partitioner.scala)
+these special classes are used during shuffles and co-grouping events (joins, groupBy) to help data 
+find it's way to the correct Spark Partition in another RDD. When one isn't specified Spark will 
+apply a `HashPartitioner`. This application requires shuffling the data so the the values in the RDD
```
jtgrabowski: the the

doc/16_partitioning.md
```diff
@@ -0,0 +1,328 @@
...
+
+CREATE KEYSPACE doc_example WITH replication = 
+    {'class': 'SimpleStrategy', 'replication_factor': '1'}  AND durable_writes = true;
+
+CREATE TABLE doc_example.purchases (
+    userid int,
+    purchaseid int,
+    objectid text,
+    amount int,
+    PRIMARY KEY (userid, purchaseid, objectid)
```
jtgrabowski: Should we mention that primary key is not the same thing as partition key?

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/doc/DocExamples.scala
```diff
@@ -112,4 +156,122 @@ class DocExamples extends SparkCassandraITFlatSpecBase {
...
+  it should "Grouping/Reducing on out of order Clustering keys" in {
+    val ks = "doc_example"
+    val rdd = {
+      sc.cassandraTable[(Int, Int)](ks, "purchases")
+        .select("purchaseid" as "_1", "amount" as "_2", "userid", "objectid")
+        .keyBy[(Int, String)]("userid", "objectid")
+    }
+
+    rdd.groupByKey.toDebugString should not include ("+-")
+    rdd.groupByKey.count should be >= numrows
```
jtgrabowski: == or <=

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/doc/DocExamples.scala
```diff
@@ -112,4 +156,122 @@ class DocExamples extends SparkCassandraITFlatSpecBase {
...
+    val rdd = {
+      sc.cassandraTable[(Int, Int)](ks, "purchases")
+        .select("purchaseid" as "_1", "amount" as "_2", "userid", "objectid")
+        .keyBy[(Int, String)]("userid", "objectid")
+    }.map(x => x)
+
+    rdd.partitioner shouldBe empty
+
+    rdd.groupByKey.toDebugString should include("+-")
+    rdd.groupByKey.count should be >= numrows
```
jtgrabowski: == or <=

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/PartitionedCassandraRDDSpec.scala
```diff
@@ -0,0 +1,277 @@
...
+  it should "NOT create a partitioned RDD that does not cover the Partition Key" in {
+    val keyedRDD = testRDD.keyBy[Tuple1[Int]](SomeColumns("ckey"))
+    keyedRDD.partitioner.isEmpty should be(true)
+  }
+
+  "CassandraTableScanRDD " should " not have a partitioner by default" in {
+    testRDD.partitioner should be(None)
+  }
+
+  it should " be able to be assigned a partititoner from RDD with the same key" in {
```
jtgrabowski: spelling in partititoner

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionerSpec.scala
```diff
@@ -20,19 +20,19 @@ class CassandraRDDPartitionerSpec
     session.execute(s"CREATE TABLE $ks.empty(key INT PRIMARY KEY)")
   }
 
-  // TODO: Currently CassandraRDDPartitioner uses a size-based algorithm that doesn't guarantee exact
+  // TODO: Currently CassandraPartitionGenerator uses a size-based algorithm that doesn't guarantee exact
```
jtgrabowski: Can we change Spec name to match partitioner new name?

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/PartitionedCassandraRDDSpec.scala
```diff
@@ -0,0 +1,277 @@
...
+  def checkPartitionerKeys[K: ClassTag, V: ClassTag](rdd: RDD[(K, V)]): Unit = {
+    rdd.partitioner shouldBe defined
+    val partitioner = rdd.partitioner.get
+    val realPartitionMap = getPartitionMap(rdd.keys)
+    for ((row, partId) <- realPartitionMap) {
+      partitioner.getPartition(row) should be(partId)
+    }
+  }
+
+  "A CassandraRDDPartitioner" should " be creatable from a generic CassandraTableRDD" in {
```
jtgrabowski: CassandraPartitionGenerator

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/partitioner/TokenGeneratorSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  "TokenGenerators" should "be able to determine simple partition keys" in {
+
+    val tokenGenerator = new TokenGenerator[SimpleKey](conn, simpleTableDef, simpleRW)
+    for ((key, token) <- simpleTokenMap) {
+      tokenGenerator.getTokenFor(key) should be (token)
+    }
+  }
+
+  it should "be able to determine simple complex keys" in {
```
jtgrabowski: please remove "simple"

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/partitioner/TokenGeneratorSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+import com.datastax.spark.connector.cql.CassandraConnector
+import com.datastax.spark.connector.cql.Schema
+import com.datastax.driver.core.Token
+import com.datastax.spark.connector.writer.RowWriterFactory
+import org.scalatest.Inspectors
+
+import scala.collection.JavaConversions._
+import scala.concurrent.Future
+
+class TokenGeneratorSpec extends SparkCassandraITFlatSpecBase with Inspectors  {
```
jtgrabowski: Inspectors are not used

doc/16_partitioning.md
RussellSpitzer: I was actually looking for a good link to explain this

RussellSpitzer: Added a link to DS academy and comment

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/partitioner/CassandraRDDPartitionerSpec.scala
RussellSpitzer: changed

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/partitioner/TokenGeneratorSpec.scala
RussellSpitzer: removed

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/partitioner/TokenGeneratorSpec.scala
RussellSpitzer: removed and  complex -> composite

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableScanRDD.scala
```diff
@@ -295,15 +420,17 @@ object CassandraTableScanRDD {
     implicit
       keyCT: ClassTag[K],
       valueCT: ClassTag[V],
-      rrf: RowReaderFactory[(K, V)]): CassandraTableScanRDD[(K, V)] = {
+      rrf: RowReaderFactory[(K, V)],
+      rwf: RowWriterFactory[K]): CassandraTableScanRDD[(K, V)] = {
 
-    new CassandraTableScanRDD[(K, V)](
+    val rdd = new CassandraTableScanRDD[(K, V)](
```
jtgrabowski: According to docs:
> Due to the way RDDs are constructed, the CassandraPartitioner cannot be applied without an explicit KeyBy call. 

Where do we use this than? Does this work?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/dht/TokenRange.scala
```diff
@@ -2,20 +2,33 @@ package com.datastax.spark.connector.rdd.partitioner.dht
...
       Seq(
         TokenRange(start, minToken, replicas, dataSize / 2),
         TokenRange(minToken, end, replicas, dataSize / 2))
     else
       Seq(this)
   }
-}
\ No newline at end of file
+
+  def contains(token: T)(implicit tf: TokenFactory[V, T]): Boolean = {
```
jtgrabowski: would be nice to test this

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraPartitioner.scala
```diff
@@ -0,0 +1,146 @@
...
+         |for table $tableDef""".stripMargin)
+    implicitly[RowWriterFactory[Key]]
+      .rowWriter(tableDef, partitionKeyMapping)
+  }
+
+  /** Builds and makes sure we can make a rowWriter with the current TableDef and keyMapper */
+  def verify(log: Boolean = true): Unit = {
+    val attempt = Try(partitionKeyWriter)
+    if (attempt.isFailure) {
+      if (log)
```
jtgrabowski: alternatively we could wrap attempt.failed.get with an exception with message "Unable to build (...)"

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/dht/Token.scala
```diff
@@ -1,17 +1,61 @@
 package com.datastax.spark.connector.rdd.partitioner.dht
 
+import com.datastax.spark.connector.rdd.partitioner.MonotonicBucketing
+import com.datastax.spark.connector.rdd.partitioner.MonotonicBucketing.LongBucketing
+
 trait Token[T] extends Ordered[Token[T]] {
+  def ord: Ordering[T]
```
jtgrabowski: Is this needed?

jtgrabowski:
+1, great job! 

left few minor comments

bcantoni:
+1 on the doc part - I did an editing pass and pushed a new commit here

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/dht/Token.scala
RussellSpitzer: This is needed because we can't compare both BigInt and Long's with ordered :( Our workaround was to pass the ordering for each type.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraPartitioner.scala
RussellSpitzer: Here  the issue is that we sometimes try to make a a partitioner and we expect it to fail because we have no other way of checking whether or not there is valid partitioner. For those times when we expect a failure we don't want to log, but if somehow the user is about run with a broken partitioner we need to actually break with a nice message.

RussellSpitzer: See the Try {} code in CassandraTableScan when it tries to make a new partitioner

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableScanRDD.scala
RussellSpitzer: It doesn't :( I also don't think we use this interface anywhere. I just wanted to try this and I left it in. I"ll remove it

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/dht/TokenRange.scala
RussellSpitzer: I'll add in a small suite to compare to driver tokens

RussellSpitzer:
Ok did everyone's comments, If the tests pass i'll merge :)

RussellSpitzer:
retest this please

RussellSpitzer:
Travis is messing up but for no apparent reason, I'll check that out more later

PR #952, opened 2016-03-28
Update DefaultSource.scala

skp33:
I added DefaultAuthConfFactory properties in confProperties variable to dynamically change username and password for dataframes.
Regarding Bug Id #352

### DIFF
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/cql/CassandraAuthenticatedConnectorSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/cql/CassandraAuthenticatedConnectorSpec.scala
@@ -6,7 +6,6 @@ import com.datastax.spark.connector.embedded.EmbeddedCassandra
 import com.datastax.spark.connector.SparkCassandraITFlatSpecBase
 import com.datastax.spark.connector.toDataFrameFunctions
 
-import org.apache.spark.sql.SQLContext
 import org.apache.spark.sql.cassandra.CassandraSQLContext
 
 class CassandraAuthenticatedConnectorSpec extends SparkCassandraITFlatSpecBase {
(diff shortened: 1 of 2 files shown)


skp33:
I am new here, Unfortunately I forked from a master branch which had reported build failures and so it affected my fork's builds process. Later, today i forked out a good master branch and i want to rebuild or submit a pull request so that i can get continuous-integration to work on my latest commit.

RussellSpitzer:
@alexliu68 this looks good to me, does it work for you?

alexliu68:
It looks good, but it's better to add some tests for it  since we are not hurry to have another release soon

RussellSpitzer:
@skp33 Can you please include some tests for verification ?

skp33:
Thanks @RussellSpitzer for this opportunity, But i am new here, i have no idea how to write test cases for this use case. Can you please guide me?

RussellSpitzer:
I think you should check out the tests in the integration test suite. Basically I think you just need to make a configuration with a new auth. Create a dataframe with that, then pull the rdd and make sure the Connection Configuration has adopted the new auth settings.

Check out this file https://github.com/datastax/spark-cassandra-connector/blob/master/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraSQLSpec.scala

Let me know if you have any more questions :)

RussellSpitzer:
This branch seems to have been rebased a bit dirtly, it has a bunch of cruft in it now so let's make sure to clean that up as well

skp33:
@RussellSpitzer , I have less knowledge about git. 

I have written some test cases for fix but Travis failing with `java.net.SocketException: Connection reset` error. In local machine all test cases executed successfully.

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/cql/CassandraAuthenticatedConnectorSpec.scala
```diff
@@ -4,34 +4,79 @@ import com.datastax.spark.connector.embedded.SparkTemplate._
...
+
+   it should "allow to read data with valid authentication" in {
+
+      val csc = new CassandraSQLContext(sc)
+
+      val options = Map("spark_cassandra_auth_username" -> "cassandra",
+         "spark_cassandra_auth_password" -> "cassandra",
+         "keyspace" -> ks, "table" -> "authtest")
+
+      val personDF = csc.read.format("org.apache.spark.sql.cassandra").options(options).load();
```
RussellSpitzer: We don't like to make tests interdependent. Ie if a test "reads" from C* the table should always exist and we can't rely on a previous test to make it. I think in this case it would be fine to just combine this with the previous test. So there is just one "read and write data with valid auth" this way the write portion is validated as well. Wdyt?

skp33:
Sure i will combine both test cases.

RussellSpitzer:
retest this please

RussellSpitzer:
I'm +1 on this, just running the integration tests again

RussellSpitzer:
Tests run fine. Merging

PR #953, opened 2016-03-30
SPARKC-356 fix itegration test classpath

jtgrabowski:
Old guava and slf4j made running integration tests within idea
impossible. This commit removes these two dependencies from test
classpath.

Now it is possible to run/debug integration tests within idea. Before
running tests specify -DbaseDir=<xyz> parameter.

### DIFF
--- a/project/Versions.scala
+++ b/project/Versions.scala
@@ -34,7 +34,7 @@ object Versions {
   val CommonsIO       = "2.4"
   val CommonsLang3    = "3.3.2"
   val Config          = "1.2.1"
-  val Guava           = "16.0.1"
+  val Guava           = "18.0"
   val JDK             = "1.7"
   val JodaC           = "1.2"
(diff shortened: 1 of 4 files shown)


jacek-lewandowski:
test this please

jtgrabowski:
retest this please

project/Versions.scala
```diff
@@ -34,7 +34,7 @@ object Versions {
   val CommonsIO       = "2.4"
   val CommonsLang3    = "3.3.2"
   val Config          = "1.2.1"
-  val Guava           = "16.0.1"
```
jtgrabowski: This seems to be delicate matter. Do you have any special corner cases on your mind that I should tests with new guava version?

RussellSpitzer: Running with Hadoop? As long as assembly works.

PR #963, opened 2016-04-12
SPARKC-360: Disable delayed retrying

pkolaczk:
Delayed retrying was implemented by Thread.sleep, which was
unfortunately executed by a Netty thread-pool. Blocking that thread pool
may lead to a general lockup of the connector.

Currently, there is no safe way of implementing delayed retries, because
such a feature would have to be implemented in the Java driver.

This change does not introduce API changes. Configuring delayed retrying is
still possible, but will have no effect.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/MultipleRetryPolicy.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/MultipleRetryPolicy.scala
@@ -5,29 +5,56 @@ import com.datastax.driver.core.policies.RetryPolicy.RetryDecision
 import com.datastax.driver.core.{ConsistencyLevel, Statement, WriteType}
 
 /** Always retries with the same CL, constant number of times, regardless of circumstances */
-class MultipleRetryPolicy(maxRetryCount: Int, retryDelay: CassandraConnectorConf.RetryDelayConf)
+class MultipleRetryPolicy(maxRetryCount: Int)
   extends RetryPolicy {
 
-  private def retryOrThrow(cl: ConsistencyLevel, nbRetry: Int): RetryDecision = {
+  private def retryManyTimesOrThrow(cl: ConsistencyLevel, nbRetry: Int): RetryDecision = {
     if (nbRetry < maxRetryCount) {
-      if (nbRetry > 0) {
-        val delay = retryDelay.forRetry(nbRetry).toMillis
-        if (delay > 0) Thread.sleep(delay)
-      }
       RetryDecision.retry(cl)
     } else {
       RetryDecision.rethrow()
     }
   }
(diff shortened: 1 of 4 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/MultipleRetryPolicy.scala
```diff
@@ -5,29 +5,56 @@ import com.datastax.driver.core.policies.RetryPolicy.RetryDecision
...
+  override def onUnavailable(
+      stmt: Statement,
+      cl: ConsistencyLevel,
+      requiredReplica: Int,
+      aliveReplica: Int,
+      nbRetry: Int): RetryDecision = {
+
+    // We retry once in hope we connect to another
+    // coordinator that can see more nodes (e.g. on another side of the network partition):
+    retryOnceOrThrow(cl, nbRetry)
```
jacek-lewandowski: Does it mean we may loose the whole partition during some administrative operations. I mean it should be probably configurable as well.

spark-cassandra-connector/src/test/scala/com/datastax/spark/connector/cql/CassandraConnectorConfSpec.scala
```diff
@@ -53,30 +53,31 @@ class CassandraConnectorConfSpec extends FlatSpec with Matchers {
     connConf.cassandraSSLConf.enabledAlgorithms should contain theSameElementsAs Seq("TLS_ECDH_RSA_WITH_AES_256_CBC_SHA384", "TLS_DHE_RSA_WITH_AES_256_CBC_SHA256")
   }
 
-  it should "resolve default retry delay settings correctly" in {
+  // retry delay temporary disabled due to SPARKC-360
+  ignore should "resolve default retry delay settings correctly" in {
```
jacek-lewandowski: Are those tests retained until Java driver implementation is changed so that we may reimplement delayed retries?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/MultipleRetryPolicy.scala
pkolaczk: But there is one retry here.
There is no point in retrying more often, because on unavailable, there is no timeout to slow down retrying.

spark-cassandra-connector/src/test/scala/com/datastax/spark/connector/cql/CassandraConnectorConfSpec.scala
pkolaczk: Right.

jacek-lewandowski:
+1

PR #968, opened 2016-04-19
SPARKC-355: Shade Cassandra driver, Guava and Netty

pkolaczk:


kovalenko-boris:
this pull request is very much needed, I run into issue described in SPARKC-355 Please fix it and merge in.

webstergd:
I would also be a huge fan of this. This would fix the major headache I encounter when using tools like Zeppelin.

project/Settings.scala
```diff
@@ -341,6 +341,14 @@ object Settings extends Build {
         case PathList("META-INF", xs @ _*) => MergeStrategy.last
         case x => old(x)
       }
+    },
+    assemblyShadeRules in assembly := {
+      val shadePackage = "shade.com.datastax.spark.connector"
+      Seq(
+        ShadeRule.rename("com.datastax.cassandra.driver.**" -> s"$shadePackage.driver.@1").inAll,
```
bchen-talend: com.datastax.driver instead of com.datastax.cassandra.driver?

pkolaczk: good catch

RussellSpitzer:
Seen some issues with this

1) I see references to the Driver broken from some call to sc.cassandraTable using the assembly jar via --jars
2) User reported errors relating to commons-configuration (apache) references being shaded but the jar not be included in assembly. (Traced this down to the jar being brought in by Hadoop2ClientLIb but excluded since that is part of Spark which is excluded)

pkolaczk:
I need to reopen this against b1.6

PR #969, opened 2016-04-19
SPARKC-233 use executeAsync when joining with C* table

jtgrabowski:
This ought to speed up joining with C* table by using executeAsync
(instead of execute) and prematerializing C* ResultSets during join
operation. Current implementation materializes all ResultSetFutures
and returns iterator that blocks on future's result.

This also adds performance module with jmh plugin. Within module
there is a simple benchmark that was used to generate results 
in one of the following comments.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
@@ -2,7 +2,7 @@ package com.datastax.spark.connector.rdd
 
 import org.apache.spark.metrics.InputMetricsUpdater
 
-import com.datastax.driver.core.Session
+import com.datastax.driver.core.{ResultSet, Session}
 import com.datastax.spark.connector._
 import com.datastax.spark.connector.cql._
 import com.datastax.spark.connector.rdd.reader._
@@ -12,8 +12,9 @@ import com.datastax.spark.connector.writer._
 import com.datastax.spark.connector.util.Quote._
 import org.apache.spark.rdd.RDD
 import org.apache.spark.{Partition, TaskContext}
-
 import scala.reflect.ClassTag
+
+import com.google.common.util.concurrent.{FutureCallback, Futures, SettableFuture}
 
 /**
  * An [[org.apache.spark.rdd.RDD RDD]] that will do a selecting join between `left` RDD and the specified
@@ -205,6 +206,12 @@ class CassandraJoinRDD[L, R] private[connector](
     query
   }
 
+  private[rdd] def boundStatementBuilder(session: Session): BoundStatementBuilder[L] = {
+    val protocolVersion = session.getCluster.getConfiguration.getProtocolOptions.getProtocolVersion
+    val stmt = session.prepare(singleKeyCqlQuery).setConsistencyLevel(consistencyLevel)
+    new BoundStatementBuilder[L](rowWriter, stmt, where.values, protocolVersion = protocolVersion)
+  }
+
   /**
    * When computing a CassandraPartitionKeyRDD the data is selected via single CQL statements
    * from the specified C* Keyspace and Table. This will be preformed on whatever data is
@@ -212,9 +219,7 @@ class CassandraJoinRDD[L, R] private[connector](
    */
   override def compute(split: Partition, context: TaskContext): Iterator[(L, R)] = {
     val session = connector.openSession()
-    val protocolVersion = session.getCluster.getConfiguration.getProtocolOptions.getProtocolVersion
-    val stmt = session.prepare(singleKeyCqlQuery).setConsistencyLevel(consistencyLevel)
-    val bsb = new BoundStatementBuilder[L](rowWriter, stmt, where.values, protocolVersion = protocolVersion)
+    val bsb = boundStatementBuilder(session)
     val metricsUpdater = InputMetricsUpdater(context, readConf)
     val rowIterator = fetchIterator(session, bsb, left.iterator(split, context))
     val countingIterator = new CountingIterator(rowIterator, limit)
@@ -230,18 +235,36 @@ class CassandraJoinRDD[L, R] private[connector](
     countingIterator
   }
 
-  private def fetchIterator(
+  private[rdd] def fetchIterator(
     session: Session,
     bsb: BoundStatementBuilder[L],
-    lastIt: Iterator[L]): Iterator[(L, R)] = {
-
+    leftIterator: Iterator[L]): Iterator[(L, R)] = {
     val columnNamesArray = selectedColumnRefs.map(_.selectedAs).toArray
-    for (leftSide <- lastIt;
-         rightSide <- {
-           val rs = session.execute(bsb.bind(leftSide))
-           val iterator = new PrefetchingResultSetIterator(rs, fetchSize)
-           iterator.map(rowReader.read(_, columnNamesArray))
-         }) yield (leftSide, rightSide)
+    val rateLimiter = new RateLimiter(
+      readConf.throughputJoinQueryPerSec, readConf.throughputJoinQueryPerSec)
+
+    def pairWithRight(left: L): SettableFuture[Iterator[(L, R)]] = {
+      val resultFuture = SettableFuture.create[Iterator[(L, R)]]
+      val leftSide = Iterator.continually(left)
+
+      val queryFuture = session.executeAsync(bsb.bind(left))
+      Futures.addCallback(queryFuture, new FutureCallback[ResultSet] {
+        def onSuccess(rs: ResultSet) {
+          val resultSet = new PrefetchingResultSetIterator(rs, fetchSize)
+          val rightSide = resultSet.map(rowReader.read(_, columnNamesArray))
+          resultFuture.set(leftSide.zip(rightSide))
+        }
+        def onFailure(throwable: Throwable) {
+          resultFuture.setException(throwable)
+        }
+      })
+      resultFuture
+    }
+    val queryFutures = leftIterator.map(left => {
+      rateLimiter.maybeSleep(1)
+      pairWithRight(left)
+    }).toList
+    queryFutures.iterator.flatMap(_.get)
   }
 
   override protected def getPartitions: Array[Partition] = {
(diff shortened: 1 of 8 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/ReadConf.scala
```diff
@@ -49,12 +53,20 @@ object ReadConf {
     description = """Sets whether to record connector specific metrics on write"""
   )
 
+  val CassandraJoinConcurrentReads = ConfigParameter[Int](
+    name = "spark.cassandra.input.join.concurrent.reads",
+    section = ReferenceSection,
+    default = 5,
```
RussellSpitzer: Gut instinct here, but I think this is low I would try benchmarking some higher defaults as well.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/MaterializingIterator.scala
```diff
@@ -0,0 +1,37 @@
...
+    }
+  }
+
+  override def hasNext = !materialized.isEmpty || iterator.hasNext
+
+  override def next() = {
+    if(!hasNext) {
+      throw new NoSuchElementException("No more elements")
+    }
+    materialize()
```
maasg: Not sure I understand the advantage here. 
This implementation will pay upfront the price of an initial n fetches but after that will behave like the underlying iterator until there're no more element in the backing iterator, when this wrapper will be faster b/c of the cached elements.

Is that what is intended? I guess that if we really want to amortize the underlying fetch for more than the initial `n` elements and to spread that profit across the iterating process, we will need some low watermark to trigger a materialization stage. 

WDYT?

RussellSpitzer: I think in the implementation it's a bit confusing. Won't this basically just equate to executing BUFFER statements async. Then waiting for at least one of them to be "gettable" before moving issuing another? The other issue is if we mix one Long query in with a bunch of slow ones. I think we may be better off just issuing async queries for everything in the iterator at once, and the force get's at the exit of the iterator. This way the Java Driver would be responsible for execution pooling. 

1. Get a ResultFuture for every join query
2. Let the driver figure out when to schedule everything
3. Use map (.get) to force syncing on the way out

I think that the driver and C* are stable enough for this now but I could be wrong

jtgrabowski: @maasg low watermark looks good, if this way of fetching rows holds I'll add it. 

@RussellSpitzer the intention of this iterator was to avoid memory overfilling by keeping number of created/fetched objects under control. With async queries for all `left` elements, aren't we at risk of OutOfMemory errors?
WDYT about modifying this by adding separate thread that would be responsible for filling up queue?

RussellSpitzer: Could be. I think the lazy loading is a probably a benefit but i'm not sure how much it really helps. If the next operation is a shuffle the whole block will have to be materialized anyway.

I'm mostly worried though about using a straight queue here, If we have Queries [a,b,c,d] In my queue, and a takes say 1 minute to pull, but b,c,d all take 1ms. I'm going to wait 1 minute without issuing any new queries.

jtgrabowski:
@RussellSpitzer , below there is a comparison of original implementation and current (based on your suggestion). As one can imagine this is most beneficial for small and oneElement partitions.

Results look ok, indeed there is a speedup, but materializing this eagerly doesn't seem right. This solution is only ok when this is true:

> If the next operation is a shuffle the whole block will have to be materialized anyway.

I have not found the confirmation yet. I'll dig deeper but maybe you already know where to look and could point me to a right class?

Benchmark | Mode | Cnt | Score | Error |Units
----------|------|-----|-------|-------|-----
measureBigPartitionsJoin_old | ss | 10 | 1519.481 |± 254.452 | ms/op
measureBigPartitionsJoin_new | ss | 10 | 1351.666 |± 318.551 | ms/op
measureModeratePartitionsJoin_old | ss | 10 | 1830.509 |± 236.377 | ms/op
measureModeratePartitionsJoin_new | ss | 10 | 778.343 |± 194.493 | ms/op
measureSmallPartitionsJoin_old | ss | 10 | 9076.712 |± 200.409 | ms/op
measureSmallPartitionsJoin_new | ss | 10 | 1790.086 |± 205.705 | ms/op
measureOneElementPerPartitionsJoin_old | ss | 10 | 74703.597 |± 708.829 | ms/op
measureOneElementPerPartitionsJoin_new | ss | 10 | 12936.777 |± 384.309 | ms/op

RussellSpitzer:
The materialization thing comes from spark need to serialize the entire partition before it can shuffle it. It doesn't move single elements from a partition during a shuffle. Since it's sending an entire block over the wire the whole partition must be in memory.  This is why you basically always need numCores * maxPartitionSize  Heap on your executors for reliable spark execution. You can sometimes spill during shuffles but it is sad times.

I wanted to up the throughput mostly because my mental model for C* is, 2ms a query but can handle roughly 20~30k queries per second. The only way to hit the throughput limit is to never block waiting for a response. Even at 5 queries in flight at a time we can only do 5 X 500 ~ 2500 (assuming perfect always 5 queries in flight) queries a second (given my imaginary latency). Anyway i'm mostly worried here about max throughput. 

If you look at the CassandraTableScanCode you'll see that it's actually materializing "fetchSize" rows at a time which is ~1000 by default. (guessing didn't check). So even when it looks like it's only doing one at a time the driver under the hood is pulling 1000 rows down.

Maybe we should just match the "fetchSize" input param. We'll obviously be underestimating for very wide partitions but it will be great for narrow partitions.

jtgrabowski:
Thanks Russ.
I added a column with throughput called 'Queries/sec' for each entry in the latest results table. Each case has 200k rows to pull so the number of queries is derived from partitions size: big, moderate, small, oneElement which is (rows per partition): 5000, 500, 10, 1. Intuitively, this fix works better for cases with smaller partitions/bigger number of queries - for oneElement per partition we are around 15k/sec on my laptop.
It makes me think that we should give users a way to throttle this, WDYT?

jtgrabowski:
So in other words, my proposal is to go with current implementation (materialize everything with one go) but give users the ability to slow down the process. `RateLimiter` class could nicely cut down the materialization process rate to desired queries/second value. The default could be not set.

RussellSpitzer:
I like that idea, I think it's a good balance.

jtgrabowski:
squashed and rebased on current master

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -230,18 +235,36 @@ class CassandraJoinRDD[L, R] private[connector](
...
-
+    leftIterator: Iterator[L]): Iterator[(L, R)] = {
     val columnNamesArray = selectedColumnRefs.map(_.selectedAs).toArray
-    for (leftSide <- lastIt;
-         rightSide <- {
-           val rs = session.execute(bsb.bind(leftSide))
-           val iterator = new PrefetchingResultSetIterator(rs, fetchSize)
-           iterator.map(rowReader.read(_, columnNamesArray))
-         }) yield (leftSide, rightSide)
+    val rateLimiter = new RateLimiter(readConf.throughputJoinQueryPerSec,
```
RussellSpitzer: style here, needs endline after open (

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -230,18 +235,36 @@ class CassandraJoinRDD[L, R] private[connector](
...
+        def onFailure(throwable: Throwable) {
+          resultFuture.setException(throwable)
+        }
+      })
+      resultFuture
+    }
+    val materializedFutures = leftIterator.map(l => {
+      rateLimiter.maybeSleep(1)
+      pairWithRight(l)
+    }).toList
```
RussellSpitzer: I'm a little confused about the toList here, won't this materialize the iterator here?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -230,18 +235,36 @@ class CassandraJoinRDD[L, R] private[connector](
...
+    materializedFutures.iterator.flatMap(_.get)
```
RussellSpitzer: I feel like we just list and that back to iterator here, don't want to flatMap on the materializedFutures?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
jtgrabowski: Yes it will. My intention was to issue all queries here (as an option, queries per second rate can be limited) and not to wait for results.
After all queries are issued, iterator is returned. The iterator blocks for each element that has no result yet.
I probably misunderstood the suggestion about issuing all queries "at once". What am I missing?

jtgrabowski:
retest this please

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraJoinRDD.scala
```diff
@@ -230,18 +235,36 @@ class CassandraJoinRDD[L, R] private[connector](
...
+          val rightSide = resultSet.map(rowReader.read(_, columnNamesArray))
+          resultFuture.set(leftSide.zip(rightSide))
+        }
+        def onFailure(throwable: Throwable) {
+          resultFuture.setException(throwable)
+        }
+      })
+      resultFuture
+    }
+    val materializedFutures = leftIterator.map(left => {
```
RussellSpitzer: Perhaps we want to rename this to something to illustrate that this is the set of query futures and that they may not have their values yet. "Materialized" in my head kept making me think that we expected the results to be here.

RussellSpitzer:
+1

PR #970, opened 2016-04-22
SPARKC-368: Cql for SinglePartition Queries needs to use >= MinToken

RussellSpitzer:
While <= MinToken wraps and works correctly by itself, when coupled with
a predicate such as "x =1" it fails to wrap. To fix this we just check
that the token(pk) >= MinToken if the partition spans (MinToken,
Mintoken)


bcantoni:
CHANGES.txt as well?

pkolaczk:
+1

pkolaczk:
@bcantoni  not needed - I always complete CHANGES.txt on making a new release.

bcantoni:
@pkolaczk thanks - that's one of my "process improvement" questions :) doing so as part of the release makes perfect sense

PR #971, opened 2016-04-22
SPARKC-367 move back to VM infrastructure

jtgrabowski:
As we have problems with container based infrastructure (these manifest
themselves as JVM exiting with 137 code). This commit moves travis
builds back to VM infrastructure, to Trusty (with 7.5GB of memory), see:
https://docs.travis-ci.com/user/ci-environment/

Also this commit bumps up parallelism for test execution

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: scala
 jdk:
   - oraclejdk8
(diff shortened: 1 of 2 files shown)


bcantoni:
+1

PR #972, opened 2016-04-27
SPARKC-372 Infer schema from DataFrame

compae:
Now the Cassandra DataSource API when implements the DefaultSource of Spark search the schema in Cassandra. It is possible that the Schema of the DataFrame is not the same with some data types. This produce errors in some scenarios.
When creating one Relation inside the DefaultSource and the data is contained in a Spark DataFrame, the CassandraSourceRelation should be created inferring the schema associated to this DataFrame.


bcantoni:
Triage: Closing per discussion in JIRA.

PR #974, opened 2016-05-05
SPARKC-376: Special Case Single Partitions and PartitionKey Restrictions

RussellSpitzer:
On older versions of Cassandra 2.1.X, it is illegal to combine Token
    restrictions with Partition Key restrictions. To be compatible with 2.1.x
    single partitions now return a single token range (rather than multiple
    with vnodes) and the CassandraTableScanRDD will no append the TokenRange
    predicate when the partition key is present. If we did not use a single
    TokenRange the query would be applied many times and you would recieve
    responses * vnodes.

dfc8058 (Russell Spitzer, 23 hours ago)
SPARKC-376: Add version gating for 2.2 and greater CQL

    So we can run tests against 2.1.14 for CassandraRDDSpec

822c98c (Russell Spitzer, 25 hours ago)
SPARKC-376: Fix Build Process for Testing against Older C* Versions

    Resources now contains template file for all currenlty supported C*
    versions. This allows the Connector to start embedded Cassandra even when
    not using the most recent C* build. There are still many tests that use 3.0
    specific features so the test suites are still failling on earlier C*
    versions.


crakjie:
I tried with my project and it's work for me. 👍

spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/EmbeddedCassandra.scala
```diff
@@ -23,10 +39,25 @@ trait EmbeddedCassandra {
     require(hosts.isEmpty || configTemplates.size <= hosts.size,
       "Configuration templates can't be more than the number of specified hosts")
 
+
+    val templateDir = {
```
crakjie: Why not alway use the full version number?

RussellSpitzer: The yamls should only be breaking on Major's from here on out (3+), but previously (2.2=-) they broke on minor prior to that.

crakjie: Ok

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/CassandraRDDSpec.scala
```diff
@@ -59,10 +59,13 @@ class CassandraRDDSpec extends SparkCassandraITFlatSpecBase {
 
     awaitAll(
       Future {
-        session.execute( s"""CREATE TABLE $ks.short_value (key INT, value SMALLINT, PRIMARY KEY (key))""")
-        session.execute( s"""INSERT INTO $ks.short_value (key, value) VALUES (1,100)""")
-        session.execute( s"""INSERT INTO $ks.short_value (key, value) VALUES (2,200)""")
-        session.execute( s"""INSERT INTO $ks.short_value (key, value) VALUES (3,300)""")
+        if (versionGreaterThanOrEquals(2,2)) {
+          println(s"Found version $cassandraMajorVersion  $cassandraMinorVersion")
```
jtgrabowski: is this intentional?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraPartitionGenerator.scala
```diff
@@ -50,10 +50,20 @@ private[connector] class CassandraPartitionGenerator[V, T <: Token[V]](
...
       val metadata = cluster.getMetadata
       for (tr <- metadata.getTokenRanges.toSeq) yield tokenRange(tr, metadata)
     }
+
+    /**
+      * When we have a single Spark Partition use a single global range. This
+      * will let us more easily deal with Partition Key equals and In clauses
+      */
+    if (splitCount == Some(1)) {
+      Seq(ranges(0).copy[V, T](tokenFactory.minToken, tokenFactory.maxToken))
```
jtgrabowski: Do I understand correctly that resulting range has invalid `replicas` and `dataSize`? If so we could skip creating `ranges` and create artificial `TokenRange` on our own.

RussellSpitzer: It has valid replicas but invalid datasize. Basically if we only have a single partition it covers the entire token range and our replica code will pick only local dc replicas (or should at least prioritize them). I considered creating a new range from scratch but you still need to correctly grab hosts for replicas to ensure you are in the right datacenter. 

This ended up seeming like a much cleaner solution for the time being than rewriting the Range creation code.

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/CassandraRDDSpec.scala
RussellSpitzer: Nah i should drop that println, that's leftover from debugging when I accidentally had my `>` switched with `<`

jtgrabowski:
+1

PR #976, opened 2016-05-09
SPARKC-377 shutdown cluster and client connection sequentially

jtgrabowski:


jtgrabowski:
retest this please

jtgrabowski:
retest this please

jtgrabowski:
retest this please

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/SerialShutdownHooks.scala
```diff
@@ -0,0 +1,29 @@
...
+import scala.collection.concurrent.TrieMap
+
+import org.apache.spark.Logging
+
+private[connector] object SerialShutdownHooks extends Logging {
+
+  private val hooks = new TrieMap[String, () => Unit]
+
+  def add(name: String)(body: () => Unit): Unit = {
+    hooks.put(name, body)
```
jacek-lewandowski: Just add some precondition so that we can be sure we won't add anything when shutdown is in progress.

.travis.yml
```diff
@@ -7,6 +7,10 @@ scala:
   - 2.10.5
   - 2.11.7
 
+cache:
+  directories:
+  - $HOME/.ivy2
```
jacek-lewandowski: Maybe it would make sense to cache `sbt` as well (`$HOME/.sbt`). I found this:
```
...
Next, we can put cache section as follows:

# These directories are cached to S3 at the end of the build
cache:
  directories:
    - $HOME/.ivy2/cache
    - $HOME/.sbt/boot/
Finally, the following a few lines of cleanup script are added:

before_cache:
  # Tricks to avoid unnecessary cache updates
  - find $HOME/.ivy2 -name "ivydata-*.properties" -delete
  - find $HOME/.sbt -name "*.lock" -delete

```
in [SBT docs](http://www.scala-sbt.org/0.13/docs/Travis-CI-with-sbt.html)

jacek-lewandowski:
+1

PR #977, opened 2016-05-10
SPARKC-379: Retry Schema Checks in case of Java Driver Debounce

RussellSpitzer:
Refactors all tableDef obtaining operations through tableFromCassandra
which attempts to retreive the given requested table def. If the call
fails, it sleeps for 1 second and retries. On the second failure it
throws an exception reporting that the table could not be found with the
table name recommendations as before. The sleep should be replaced with
a forced schema update once that becomes available in the Java Driver.


jtgrabowski:
+1, please squash

RussellSpitzer:
Sqooushed

RussellSpitzer:
retest this please

PR #979, opened 2016-05-13
SPARKC-233 use executeAsync when joining with C* table

jtgrabowski:
This is a backport of #969


jtgrabowski:
retest this please

PR #981, opened 2016-05-17
Add CLUSTERING ORDER in cql statement

skp33:
I added code for **Clustering Order** support for table creation. Please review the code.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
@@ -88,11 +88,30 @@ case class ClusteringColumn(index: Int) extends ColumnRole
 case object StaticColumn extends ColumnRole
 case object RegularColumn extends ColumnRole
 
+/**
+  * Cluster Column Order Implementation */
+object ClusteringOrder {
+  abstract sealed class Order(val ordinal: Int) extends Ordered[Order]
+       with Serializable {
+     def compare(that: Order) = that.ordinal compare this.ordinal
+
+     def toInt: Int = this.ordinal
+  }
+
+  case object Ascending extends Order(0)
+  case object Descending extends Order(1)
+
+  def fromInt(i: Int): Order = values.find(_.ordinal == i).get
+
+  val values = Set(Ascending, Descending)
+}
+
 /** A Cassandra column metadata that can be serialized.  */
 case class ColumnDef(
   columnName: String,
   columnRole: ColumnRole,
-  columnType: ColumnType[_]) extends FieldDef {
+  columnType: ColumnType[_],
+  clusteringOrder: ClusteringOrder.Order = ClusteringOrder.Ascending) extends FieldDef {
 
   def ref: ColumnRef = ColumnName(columnName)
   def isStatic = columnRole == StaticColumn
@@ -128,6 +147,15 @@ object ColumnDef {
     val columnType = ColumnType.fromDriverType(column.getType)
     ColumnDef(column.getName, columnRole, columnType)
   }
+  
+  def apply(
+    column: ColumnMetadata,
+    columnRole: ColumnRole,
+    clusteringOrder: ClusteringOrder.Order): ColumnDef = {
+
+    val columnType = ColumnType.fromDriverType(column.getType)
+    ColumnDef(column.getName, columnRole, columnType, clusteringOrder)
+  }
 }
 
 /** A Cassandra table metadata that can be serialized. */
@@ -138,7 +166,8 @@ case class TableDef(
     clusteringColumns: Seq[ColumnDef],
     regularColumns: Seq[ColumnDef],
     indexes: Seq[IndexDef] = Seq.empty,
-    isView: Boolean = false) extends StructDef {
+    isView: Boolean = false, 
+    options: String = "") extends StructDef {
 
   require(partitionKey.forall(_.isPartitionKeyColumn), "All partition key columns must have role PartitionKeyColumn")
   require(clusteringColumns.forall(_.isClusteringColumn), "All clustering columns must have role ClusteringColumn")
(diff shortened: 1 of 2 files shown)


RussellSpitzer:
I think this is a good addition but I have two major requests. 

1. We need a Jira for tracking
2. I don't like just having a String "option". I think we are slowly approaching fully redoing the Java Driver TableMetadata so I think we should have it be a copy of that object. We don't have to necessarily expose all the parameters right now but I'd feel more comfortable with TableDef just having a list of ClusteringOrders.

https://docs.datastax.com/en/drivers/java/2.0/com/datastax/driver/core/TableMetadata.html

ClusteringOrder is represented by a list of Orderings so very close to what you currently have.

skp33:
@RussellSpitzer based on your suggestion i changed the code. Do i need to add & change anything else?

RussellSpitzer:
Make a jira https://datastax-oss.atlassian.net/projects/SPARKC/ For tracking (this is how we catch everything up in release notes and track versioning)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
```diff
@@ -138,7 +149,9 @@ case class TableDef(
     clusteringColumns: Seq[ColumnDef],
     regularColumns: Seq[ColumnDef],
     indexes: Seq[IndexDef] = Seq.empty,
-    isView: Boolean = false) extends StructDef {
+    isView: Boolean = false,
+    clusteringOrder: Option[Seq[ClusteringOrder]] = None,
```
RussellSpitzer: Shouldn't this be a property of the ColunDef now? Ie do you really need to specify it separately?

skp33:
@RussellSpitzer , Jira is already there, this is the [Link](https://datastax-oss.atlassian.net/browse/SPARKC-403).

According to discussion, we need to have **list of ClusteringOrders inside TableDef**, correct me if i miss understood.

And for property part, it depends on use case, i don't have better project knowledge, so let me know what should be there?

RussellSpitzer:
It looks to me like the information about ClusteringOrders is in two places within the TableDef. 

1) Available as an indexedSequence `clusteringOrder: Option[Seq[ClusteringOrder]] = None,`
2) Available within the clusteringColumn objects themselves `+  clusteringOrder: ClusteringOrder = ClusteringOrder.ASC) extends FieldDef {`

I think only 2 is neccessary? Having both 1 and 2 gives us the possibility of having them not matching which seems like a bit of a hole in the implementation. I think it would be find to have a function on tableDef

`def clusterOrder: Seq[ClusteringOrder] = clusteringColumns.map(_.clusteringOrder`

If we want it exposed at that level.

skp33:
I changed according to you, Please verify. And do we need to add some functionality for this method also **DataFrameFunctions.createCassandraTable**?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
```diff
@@ -138,7 +150,8 @@ case class TableDef(
     clusteringColumns: Seq[ColumnDef],
     regularColumns: Seq[ColumnDef],
     indexes: Seq[IndexDef] = Seq.empty,
-    isView: Boolean = false) extends StructDef {
+    isView: Boolean = false,
+    options: String = "") extends StructDef {
```
RussellSpitzer: I still don't like being able to just pass a string here, If you really think we need this I think it should at least be Seq[String] and we should just require that they not contain "AND" or "WITH". Then we can convert the append code below into  a string join instead of having the more complicated logic.


This removes the need for the "appendOptions" function and replaces it with
```
require(options.forAll( option => !(option.toLowerCase.contains("and") && !(option.toLowerCase.contains("WITH")), "Table options must not contain "WITH OR AND"
(options +: clusteringOptions).mkString("WITH", "AND")```
```

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
```diff
@@ -185,11 +198,25 @@ case class TableDef(
     val clusteringColumnNames = clusteringColumns.map(_.columnName).map(quote)
     val primaryKeyClause = (partitionKeyClause +: clusteringColumnNames).mkString(", ")
 
-    s"""CREATE TABLE ${quote(keyspaceName)}.${quote(tableName)} (
+    val stmt = s"""CREATE TABLE ${quote(keyspaceName)}.${quote(tableName)} (
        |  $columnList,
        |  PRIMARY KEY ($primaryKeyClause)
        |)""".stripMargin
+    val ordered = if (clusteringColumns.size > 0)
+      s"$stmt${Properties.lineSeparator}WITH CLUSTERING ORDER BY (${clusteringColumnOrder(clusteringColumns)})"
```
RussellSpitzer: also when including line seps I think is fine to just use """ quotes

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
```diff
@@ -185,11 +198,25 @@ case class TableDef(
     val clusteringColumnNames = clusteringColumns.map(_.columnName).map(quote)
     val primaryKeyClause = (partitionKeyClause +: clusteringColumnNames).mkString(", ")
 
-    s"""CREATE TABLE ${quote(keyspaceName)}.${quote(tableName)} (
+    val stmt = s"""CREATE TABLE ${quote(keyspaceName)}.${quote(tableName)} (
        |  $columnList,
        |  PRIMARY KEY ($primaryKeyClause)
        |)""".stripMargin
+    val ordered = if (clusteringColumns.size > 0)
```
RussellSpitzer: this could be 
```orderOptions = clusteringColumns.map(_.clusteringOrder).mkString("WITH CLUSTERING ORDER BY", ",")```
To remove the If Statements

File: spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala [hunk 5]
```diff
@@ -151,11 +151,12 @@ case class TableDef(
     regularColumns: Seq[ColumnDef],
     indexes: Seq[IndexDef] = Seq.empty,
     isView: Boolean = false,
-    options: String = "") extends StructDef {
+    options: Seq[String] = Seq.empty) extends StructDef {
```
RussellSpitzer: Perhaps this should be tableOptions (matching the cql docs), and also now that I think about this perhaps it fits a Map better than a sequence? This would make it much clearer that we are looking for a set of key-value pairs.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
```diff
@@ -151,11 +151,12 @@ case class TableDef(
...
 
   require(partitionKey.forall(_.isPartitionKeyColumn), "All partition key columns must have role PartitionKeyColumn")
   require(clusteringColumns.forall(_.isClusteringColumn), "All clustering columns must have role ClusteringColumn")
   require(regularColumns.forall(!_.isPrimaryKeyColumn), "Regular columns cannot have role PrimaryKeyColumn")
+  require(options.forall( option => !(option.toLowerCase.contains("and") && !(option.toLowerCase.contains("with")))), "Table options must not contain WITH OR AND")
```
RussellSpitzer: What if I want to set my comment on a table to Sand Castles _with_ Judge's Rankings. Basically I think we should let the driver validate the tableOptions. If we want to test them here we should probably only test keys  (once we change the tableOptions to a map). I think we are probably best off without the requires here.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
```diff
@@ -202,21 +203,12 @@ case class TableDef(
...
-  def clusterOrder: Seq[ClusteringOrder] = clusteringColumns.map(_.clusteringOrder)
+    val ordered = clusteringColumns.map( col => s"${quote(col.columnName)} ${col.clusteringOrder}")
+      .mkString("CLUSTERING ORDER BY (", ", ",")")
 
-  private[this] def appendOptions(stmt: String, opts: String) =
-    if (stmt.contains("WITH") && opts.startsWith("WITH")) s"$stmt${Properties.lineSeparator}AND ${opts.substring(4)}"
-    else if (!stmt.contains("WITH") && opts.startsWith("AND")) s"WITH ${opts.substring(3)}"
-    else if (opts == "") s"$stmt"
-    else s"$stmt${Properties.lineSeparator}$opts"
+    val orderWithOptions:Seq[String] = if (clusteringColumns.size > 0) options.+:(ordered) else options
```
RussellSpitzer: lets just go straight to the treble quote here, I think it may be a bit clearer like
```
val tableOptionsClause = s"WITH $clusteringOrderingClause ${(tableOptions...).mkString(AND)}"
"""$stmt $tableOptionsClause"""

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
skp33: I changed accordingly. What will be the better option to validate tableOptions keys?

PR #983, opened 2016-05-19
SPARKC-383: cache all common CassandraRow data in a single CassandraR…

artem-aliev:
…owMetaData object

One CassandraRowMetaData per resultSet/Partition.
That reduce object creation time and memory footprint.
At this momement it is:
1. Java driver codec
2. Map comlumn names -> index for get by name methods
3. CqlColumnNames -> index

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
@@ -1,6 +1,6 @@
 package com.datastax.spark.connector
 
-import com.datastax.driver.core.Row
+import com.datastax.driver.core.{CodecRegistry, ResultSet, Row, TypeCodec}
 
 /** Represents a single row fetched from Cassandra.
   * Offers getters to read individual fields by column name or column index.
@@ -17,25 +17,25 @@ import com.datastax.driver.core.Row
   *
   * Recommended getters for Cassandra types:
   *
-  *   - `ascii`:     `getString`, `getStringOption`
-  *   - `bigint`:    `getLong`, `getLongOption`
-  *   - `blob`:      `getBytes`, `getBytesOption`
-  *   - `boolean`:   `getBool`, `getBoolOption`
-  *   - `counter`:   `getLong`, `getLongOption`
-  *   - `decimal`:   `getDecimal`, `getDecimalOption`
-  *   - `double`:    `getDouble`, `getDoubleOption`
-  *   - `float`:     `getFloat`, `getFloatOption`
-  *   - `inet`:      `getInet`, `getInetOption`
-  *   - `int`:       `getInt`, `getIntOption`
-  *   - `text`:      `getString`, `getStringOption`
-  *   - `timestamp`: `getDate`, `getDateOption`
-  *   - `timeuuid`:  `getUUID`, `getUUIDOption`
-  *   - `uuid`:      `getUUID`, `getUUIDOption`
-  *   - `varchar`:   `getString`, `getStringOption`
-  *   - `varint`:    `getVarInt`, `getVarIntOption`
-  *   - `list`:      `getList[T]`
-  *   - `set`:       `getSet[T]`
-  *   - `map`:       `getMap[K, V]`
+  * - `ascii`:     `getString`, `getStringOption`
+  * - `bigint`:    `getLong`, `getLongOption`
+  * - `blob`:      `getBytes`, `getBytesOption`
+  * - `boolean`:   `getBool`, `getBoolOption`
+  * - `counter`:   `getLong`, `getLongOption`
+  * - `decimal`:   `getDecimal`, `getDecimalOption`
+  * - `double`:    `getDouble`, `getDoubleOption`
+  * - `float`:     `getFloat`, `getFloatOption`
+  * - `inet`:      `getInet`, `getInetOption`
+  * - `int`:       `getInt`, `getIntOption`
+  * - `text`:      `getString`, `getStringOption`
+  * - `timestamp`: `getDate`, `getDateOption`
+  * - `timeuuid`:  `getUUID`, `getUUIDOption`
+  * - `uuid`:      `getUUID`, `getUUIDOption`
+  * - `varchar`:   `getString`, `getStringOption`
+  * - `varint`:    `getVarInt`, `getVarIntOption`
+  * - `list`:      `getList[T]`
+  * - `set`:       `getSet[T]`
+  * - `map`:       `getMap[K, V]`
   *
   * Collection getters `getList`, `getSet` and `getMap` require to explicitly pass an appropriate item type:
   * {{{
@@ -46,17 +46,17 @@ import com.datastax.driver.core.Row
   *
   * Generic `get` allows to automatically convert collections to other collection types.
   * Supported containers:
-  *   - `scala.collection.immutable.List`
-  *   - `scala.collection.immutable.Set`
-  *   - `scala.collection.immutable.TreeSet`
-  *   - `scala.collection.immutable.Vector`
-  *   - `scala.collection.immutable.Map`
-  *   - `scala.collection.immutable.TreeMap`
-  *   - `scala.collection.Iterable`
-  *   - `scala.collection.IndexedSeq`
-  *   - `java.util.ArrayList`
-  *   - `java.util.HashSet`
-  *   - `java.util.HashMap`
+  * - `scala.collection.immutable.List`
+  * - `scala.collection.immutable.Set`
+  * - `scala.collection.immutable.TreeSet`
+  * - `scala.collection.immutable.Vector`
+  * - `scala.collection.immutable.Map`
+  * - `scala.collection.immutable.TreeMap`
+  * - `scala.collection.Iterable`
+  * - `scala.collection.IndexedSeq`
+  * - `java.util.ArrayList`
+  * - `java.util.HashSet`
+  * - `java.util.HashMap`
   *
   * Example:
   * {{{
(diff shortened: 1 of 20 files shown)


artem-aliev:
retest this please

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
```diff
@@ -72,12 +72,65 @@ import com.datastax.driver.core.Row
   *   - java.sql.Date
   *   - org.joda.time.DateTime
   */
-final class CassandraRow(val columnNames: IndexedSeq[String], val columnValues: IndexedSeq[AnyRef])
+final class CassandraRow(val metaData:CassandraRowMetaData, val columnValues: IndexedSeq[AnyRef])
```
RussellSpitzer: style, missing space

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
```diff
@@ -72,12 +72,65 @@ import com.datastax.driver.core.Row
...
   extends ScalaGettableData with Serializable {
 
+  /**
+    * the consturctor is for testing and backward compatibility only.
+    * Use default constructor with shared metadata for memory saving and performance.
```
RussellSpitzer: Maybe mark this as deprecated?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
```diff
@@ -72,12 +72,65 @@ import com.datastax.driver.core.Row
...
+    * @param columnNames
+    * @param columnValues
+    */
+  def this (columnNames: IndexedSeq[String], columnValues: IndexedSeq[AnyRef]) =
+    this(CassandraRowMetaData.fromColumnNames(columnNames), columnValues)
+
   override def toString = "CassandraRow" + dataAsString
 }
 
+case class CassandraRowMetaData(columnNames: IndexedSeq[String], resultSetColumnNames:Option[IndexedSeq[String]] = None, @transient codecs: Option[IndexedSeq[TypeCodec[AnyRef]]] = None) {
```
RussellSpitzer: This seems like it's copying a lot of functionality of some existing classes. The classes in https://github.com/datastax/spark-cassandra-connector/blob/master/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala seem to already cover a lot of this `ColumnRefs` should already have their aliases, indexing and maps should exist in `TableDef`. Perhaps it would be best to store the codec's their as well?

RussellSpitzer: I think my confusing is behind we this isn't just an index of ColumnNames and Codecs

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
artem-aliev: NP, will fix

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
artem-aliev: I do not like to mix in sparkSQL here and schema, tableDefs are much more complicated then needed. The other purpose of the RowMetaData is the place for any data that can be shared among rows. I mean all data execept the cell values itself. The dedicated object will flag developers to use it

RussellSpitzer: OK, as long as Piotr is ok with it too :)

pkolaczk: ColumnRefs and ColumnDefs are Serializable, so adding codecs to them is not possible.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
```diff
@@ -72,12 +72,66 @@ import com.datastax.driver.core.Row
   *   - java.sql.Date
   *   - org.joda.time.DateTime
   */
-final class CassandraRow(val columnNames: IndexedSeq[String], val columnValues: IndexedSeq[AnyRef])
+final class CassandraRow(val metaData: CassandraRowMetaData, val columnValues: IndexedSeq[AnyRef])
```
pkolaczk: I only wonder if  `metaData` doesn't end up as multiple objects after deserializing Rows. I hope Spark doesn't serialize / deserialize rows one at a time.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
pkolaczk: We can't add codecs to Schema (ColumnDefs / ColumnRefs), because Schema must be Serializable.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
```diff
@@ -72,12 +72,66 @@ import com.datastax.driver.core.Row
...
+        s"Available columns are: ${columnNames.mkString("[", ", ", "]")}")
+  }
+
+  @transient
+  lazy val indexOfOrThrow = namesToIndex.withDefault { name =>
+    throw new ColumnNotFoundException(
+      s"Column not found: $name. " +
+        s"Available columns are: ${columnNames.mkString("[", ", ", "]")}")
+  }
+  def unaliasColumnNames = resultSetColumnNames.getOrElse(columnNames)
```
pkolaczk: Did you mean `unaliasedColumnNames`?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
pkolaczk: Why resultSetColumnNames are an `Option`? What does it mean when they are not set?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
artem-aliev: run following code to get all Row Serialised:
cala> sc.cassandraTable("ks", "tab").repartition(2).count
res0: Long = 20000000

jvisualvm memory sampler show that number of meta store object is small:
com.datastax.spark.connector.CassandraRowMetaData	0.03575835062886026	129,600 B (0%)	3,240 (0%)
com.datastax.spark.connector.CassandraRow	1.0719691326853993	3,885,168 B (1.1%)	161,882 (1.9%)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
artem-aliev: That means that row was constructed not from ResultSet. Manually for example.
actually it is always resolved as resultSetColumnNames.getOrElse(columnNames). So I can copy it at creation time.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
artem-aliev: yep, will fix

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
pkolaczk: rename `MetaData` to `Metadata` to be consistent with Java driver naming

pkolaczk: also, this line is far too long

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
```diff
@@ -72,12 +72,66 @@ import com.datastax.driver.core.Row
...
+  @deprecated
+  def this (columnNames: IndexedSeq[String], columnValues: IndexedSeq[AnyRef]) =
+    this(CassandraRowMetaData.fromColumnNames(columnNames), columnValues)
+
   override def toString = "CassandraRow" + dataAsString
 }
 
+case class CassandraRowMetaData(columnNames: IndexedSeq[String], resultSetColumnNames:Option[IndexedSeq[String]] = None, @transient codecs: Option[IndexedSeq[TypeCodec[AnyRef]]] = None) {
+  @transient
+  lazy val namesToIndex: Map[String,Int] = columnNames.zipWithIndex.toMap.withDefaultValue(-1);
```
pkolaczk: unneeded ;

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
pkolaczk: codecs are marked @transient, are you sure you want them to become null after deserializing?
```
scala> case class X(@transient f: Option[Int] = None)
defined class X

scala> import java.util._
import java.util._

scala> import java.io._
import java.io._

scala> val bos = new ByteArrayOutputStream
bos: java.io.ByteArrayOutputStream =

scala> val os = new ObjectOutputStream(bos)
os: java.io.ObjectOutputStream = java.io.ObjectOutputStream@7f9fcf7f

scala> os.writeObject(new X(Some(123)))

scala> os.close()

scala> val bis = new ByteArrayInputStream(bos.toByteArray)
bis: java.io.ByteArrayInputStream = java.io.ByteArrayInputStream@22555ebf

scala> val is = new ObjectInputStream(bis)
is: java.io.ObjectInputStream = java.io.ObjectInputStream@131ef10

scala> val deserialized = is.readObject()
deserialized: Object = X(null)
//             ^^^^^^^^^^^^^^^^^^ see? null
```

pkolaczk: I think because of nulling the codecs field, it would be much safer to make them `private`. They seem to be needed only for constructing `CassandraRows` and nothing more

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
```diff
@@ -72,12 +72,66 @@ import com.datastax.driver.core.Row
...
+        s"Available columns are: ${columnNames.mkString("[", ", ", "]")}")
+  }
+  def unaliasColumnNames = resultSetColumnNames.getOrElse(columnNames)
+}
+
+object CassandraRowMetaData {
+
+  def fromResultSet(columnNames:IndexedSeq[String], rs: ResultSet) = {
+    import scala.collection.JavaConversions._
+    val columnDefs = rs.getColumnDefinitions.asList().toList
```
pkolaczk: unneccessary `toList`

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
```diff
@@ -72,12 +72,66 @@ import com.datastax.driver.core.Row
...
+  /**
+    * create metadata object without codecs. Should be used for testing only
+    *
+    * @param columnNames
+    * @return
+    */
+  def fromColumnNames (columnNames: IndexedSeq[String]): CassandraRowMetaData =
+    CassandraRowMetaData(columnNames, None)
+
+  def fromColumnNames (columnNames: Seq[String]): CassandraRowMetaData = fromColumnNames(columnNames.toIndexedSeq)
```
pkolaczk: line slightly too long

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
```diff
@@ -86,18 +140,23 @@ object CassandraRow {
...
     * slower than fetching by index. */
-  def fromJavaDriverRow(row: Row, columnNames: Array[String]): CassandraRow = {
-    val data = new Array[Object](columnNames.length)
-    for (i <- columnNames.indices)
-      data(i) = GettableData.get(row, i)
-    new CassandraRow(columnNames, data)
+  def fromJavaDriverRow(row: Row, metaData: CassandraRowMetaData): CassandraRow = {
+    val data = new Array[Object](metaData.columnNames.length)
+    for (i <- metaData.columnNames.indices) {
+      data(i) = metaData.codecs match {
```
pkolaczk: I think that if you took out the match out of the inner loop, and had two loops under match, it would be even faster (probably not by much, but worth trying).

pkolaczk:
I finished the review. Overall looks good. There are minor code style issues.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/japi/CassandraRow.scala
```diff
@@ -1,17 +1,26 @@
...
-  private[spark] def this() = this(null: IndexedSeq[String], null) // required by Kryo for deserialization :(
+  private[spark] def this() = this(null: CassandraRowMetaData, null: IndexedSeq[AnyRef]) // required by Kryo for deserialization :(
 
+  def this(metaData: CassandraRowMetaData, columnValues: Array[AnyRef]) =
+    this(metaData, columnValues.toIndexedSeq)
+
+  /**
+    * the consturctor is for testing and backward compatibility only.
+    * Use default constructor with shared metadata for memory saving and performance.
+    * @param columnNames
```
pkolaczk: remove, if not documented

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
pkolaczk: would be good to document constructor arguments

artem-aliev: Aha, Scala is fun, It reconstruct lazy values after deserialzation, but do not use default values.

artem-aliev: fixed. will use null instead of option

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
artem-aliev: fixed

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
artem-aliev: that is java to scala int conversion. toScala does not work

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
artem-aliev: Fixed, I also replace for() with mutable while() to reduce number of created objects

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/japi/CassandraRow.scala
artem-aliev: fixed

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/CassandraRow.scala
```diff
@@ -86,18 +153,37 @@ object CassandraRow {
...
+        i += 1
+      }
+    }
+    else {
+      while (i < length) {
+        data(i) = GettableData.get(row, i, metaData.codecs(i))
+        i += 1
+      }
+    }
+    new CassandraRow(metaData, data)
```
artem-aliev: I have rewiten that block to java style to remove closure object creation for each row

pkolaczk:
+1

PR #984, opened 2016-05-24
 SPARKC-385 add driver.LocalDate conversion

jtgrabowski:
Conversion is added because driver's LocalDate is not
serializable.

GettableData now converts LocalDate to Joda's LocalDate.


RussellSpitzer:
Good catch I could have sworn LocalDate was serializable before.

RussellSpitzer:
You may want to add a few more tests to the unit test suite
https://github.com/datastax/spark-cassandra-connector/blob/master/spark-cassandra-connector/src/test/scala/com/datastax/spark/connector/types/TypeConverterTest.scala#L216
But i'm +1 on the patch :)

jtgrabowski:
@RussellSpitzer during implementing additional tests you recommended I realized that conversion from Joda's LocalDate -> Date is not behaving as conversion from driver's LocalDate -> Date. I fixed this and added some tests in latest commit.

RussellSpitzer:
Now I don't feel bad about asking for more tests :) +1

PR #987, opened 2016-06-02
SPARKC-278: Make Repartition by Cassandra Replica Deterministic

RussellSpitzer:
Previously the behavior of repartition by Cassandra Replica was random
within the bounds of assigining keys to the the correct replicas. This
meant that multiple application of the partitioner could result with
rows in different partitions even if they share the same partition key.

The new behvaior pulls the token determination and replica determination
into the partitioner. This removes the need for the
"keyByCassandraReplica" function for repartitionByCassandraReplica.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
@@ -38,6 +38,7 @@ implicit
   @transient lazy private val metadata = connector.withClusterDo(_.getMetadata)
   @transient lazy private val protocolVersion = connector
     .withClusterDo(_.getConfiguration.getProtocolOptions.getProtocolVersion)
+  @transient lazy private val clazz = implicitly[ClassTag[T]].runtimeClass
 
   private val hosts = connector.hosts.toVector
   private val hostSet = connector.hosts
@@ -62,7 +63,7 @@ implicit
    */
   override def getPartition(key: Any): Int = {
     key match {
-      case key: T =>
+      case key: T if clazz.isInstance(key) =>
         //Only use ReplicaEndpoints in the connected DC
         val token = tokenGenerator.getTokenFor(key)
         val tokenHash = Math.abs(token.hashCode())
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
@@ -2,8 +2,13 @@ package com.datastax.spark.connector.rdd.partitioner
 
 import java.net.InetAddress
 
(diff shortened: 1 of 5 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
```diff
@@ -38,14 +62,20 @@ class ReplicaPartitioner(partitionsPerReplicaSet: Int, val connector: CassandraC
    */
   override def getPartition(key: Any): Int = {
     key match {
-      case key: Set[_] if key.size > 0 && key.forall(_.isInstanceOf[InetAddress]) =>
+      case key: T =>
```
pkolaczk: This is unchecked and will always match.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
```diff
@@ -38,14 +62,20 @@ class ReplicaPartitioner(partitionsPerReplicaSet: Int, val connector: CassandraC
...
         if (replicaSetInDC.nonEmpty) {
-          val endpoint = replicaSetInDC(rand.nextInt(replicaSetInDC.size))
-          hostMap(endpoint)(rand.nextInt(partitionsPerReplicaSet))
+          val endpoint = replicaSetInDC(tokenHash % replicaSetInDC.size)
+          hostMap(endpoint)(tokenHash % partitionsPerReplicaSet)
         } else {
-          hostMap(randomHost)(rand.nextInt(partitionsPerReplicaSet))
+          hostMap(randomHost(tokenHash))(tokenHash % partitionsPerReplicaSet)
         }
       case _ => throw new IllegalArgumentException(
```
pkolaczk: Does it make any sense to keep it if we're matching all objects with the previous case?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
RussellSpitzer: I pass in the classtag? Isn't that enough. I don't see a warning in the compiler?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/ReplicaPartitioner.scala
```diff
@@ -38,14 +63,20 @@ class ReplicaPartitioner(partitionsPerReplicaSet: Int, val connector: CassandraC
    */
   override def getPartition(key: Any): Int = {
     key match {
-      case key: Set[_] if key.size > 0 && key.forall(_.isInstanceOf[InetAddress]) =>
+      case key: T if clazz.isInstance(key) =>
```
RussellSpitzer: Added in the class check, Not sure if this is the right thing to do

pkolaczk:
+1

PR #988, opened 2016-06-06
SPARKC-393 : Keep the partitioner of the left RDD

crakjie:
Because it not modify the partitioner.
And it's help in case of groupBy, the previous partitioner will be selected by default. 
For exemple after a repartitionByCassandraReplica.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,6 @@
+2.0.0 M3
+ * Includes all patches up to 1.6.2
+
 2.0.0 M2
  * Includes all patches up to 1.6.1
 
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ This project has also been published to the Maven Central Repository.
 For SBT to download the connector binaries, sources and javadoc, put this in your project 
 SBT config:
                                                                                                                            
-    libraryDependencies += "com.datastax.spark" %% "spark-cassandra-connector" % "2.0.0-M2"
+    libraryDependencies += "com.datastax.spark" %% "spark-cassandra-connector" % "2.0.0-M3"
 
 ## Building
 See [Building And Artifacts](doc/12_building_and_artifacts.md)


RussellSpitzer:
I think this is always the right thing to do. I only have one request that we clean up the Commit Message to reference the SPARKC ticket and be a little more explanatory of the change.

What was the code before?
What is the code now?
What positive and negatives will this change have if any?

RussellSpitzer:
I meant the title of the commit 

*Keep the partitioner of the left RDD *

to 

*SPARKC-393 : Keep the partitioner of the left RDD after JWCT*


You should b e able to just do a git commit --amend , the force push the new commit

crakjie:
Yes yes sorry, it just a matter of time.
I'm overflowed by work in this end of week.
I do that when I have time, don't worry.

But you answer in the jira give me a doute about this. I also think is the right thing to do, but I wonder if it's can not be tricky for users.

RussellSpitzer:
You can always change it to an optional parameter? ( , preservePartitioner = Boolean )

crakjie:
If you think it's the right way to do, I'm ok.
default true or false?

RussellSpitzer:
default to the old behavior, False, that seems safest @jtgrabowski any comments?

crakjie:
call for review

RussellSpitzer:
LGTM +1 but currently needs to be rebased

doc/2_loading.md
```diff
@@ -283,12 +282,22 @@ Cassandra Tables which have partitioned with different partition keys.
...
-val localQueryRDD = oddIds.repartitionByCassandraReplica("test","customer_info").joinWithCassandra("test","customer_info")
+val localQueryRDD = oddIds.repartitionByCassandraReplica("test","customer_info").joinWithCassandraTable("test","customer_info")
 repartitionRDD.collect.foreach(println)
 //(CustomerID(1),CassandraRow{cust_id: 1, address: East Coast, name: Helena})
 //(CustomerID(3),CassandraRow{cust_id: 3, address: Poland, name: Jacek})
 ```
 
+####Preserve partitoner after joinWithCassandraRDD
+
+If you want to keep the previous rdd use the parameter preservePartitioner
```
RussellSpitzer: rdd's partitioner

doc/2_loading.md
```diff
@@ -283,12 +282,22 @@ Cassandra Tables which have partitioned with different partition keys.
...
+
+In the following example the repartitioning by cassandra replica is preserved
```
RussellSpitzer: The "partitioner from reparitionByCassandraReplica is preserverd" Note that the partitioner is not guaranteed to be able to work with the subsequent keys.

crakjie: But after a join the subsequent keys are generated by the partitioner ?

crakjie:
Art I have forgot there is java code.
How this param should be shown in java code since there is no default parameters?

when you say "needs to be rebased" you mean not merge commit ?

RussellSpitzer:
repartitionByCassandraReplica for example loses the key it uses to repartition with. So the partitioner is no longer good

Rebasing is a git term for update the branch to the latest master

```
git rebase -i branchToRebaseTo
```

crakjie:
Yes I know, I have rebased instead of merging ( for the first time I used the "Github desktop" which has a nice UI but do weird things like merge and remove spaces ).

For the java code that need the param should I avoid the preservePartitioner parameter and set it to false? Or should I expose it to java users ?

PR #991, opened 2016-06-13
SPARKC-230 add minimal threshold for number of partitions

jtgrabowski:
This changes CassandraPartitionGenerator so that it takes only
splitCount parameter (instead of splitCount and splitSize parameters).
This change is propagated down to Clusterer and Splitters.

Also this adds minimal threshold for number of partitions generated by
CassandraPartitionGenerator. Minimal threshold is evaluated as:
1 + context.defaultParallelism * 2


spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/CassandraJavaRDDSpec.scala
```diff
@@ -413,9 +417,10 @@ class CassandraJavaRDDSpec extends SparkCassandraITFlatSpecBase {
   }
 
   it should "allow to set limit" in {
-    val rdd = javaFunctions(sc).cassandraTable(ks, "test_table").limit(1L)
+    val limit = 1
+    val rdd = javaFunctions(sc).cassandraTable(ks, "limit_test_table").limit(limit.toLong)
     val result = rdd.collect()
-    result should have size 1
+    result.size shouldBe <= (rdd.getNumPartitions * limit)
```
RussellSpitzer: Why are we multiplying by 1?

RussellSpitzer: Actually i'm very confused by this... What is this supposed to be checking?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraPartitionGenerator.scala
```diff
@@ -30,23 +29,11 @@ private[connector] class CassandraPartitionGenerator[V, T <: Token[V]](
   private val keyspaceName = tableDef.keyspaceName
   private val tableName = tableDef.tableName
 
-  private val totalDataSize: Long = {
```
RussellSpitzer: this weirdness definitely outstayed it's welcome :) Thanks for getting rid of it

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/dht/TokenRange.scala
```diff
@@ -3,7 +3,11 @@ package com.datastax.spark.connector.rdd.partitioner.dht
 import java.net.InetAddress
 
 case class TokenRange[V, T <: Token[V]] (
-    start: T, end: T, replicas: Set[InetAddress], dataSize: Long) {
+    start: T, end: T, replicas: Set[InetAddress], tokenFactory: TokenFactory[V, T]) {
+
+  private[partitioner] lazy val distance = tokenFactory.distance(start, end)
```
RussellSpitzer: rangeSize may be a more descriptive name

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/Murmur3PartitionerTokenRangeSplitter.scala
```diff
@@ -1,34 +1,21 @@
...
   extends TokenRangeSplitter[Long, LongToken] {
 
-  private val tokenFactory =
-    TokenFactory.Murmur3TokenFactory
+  private type TokenRange = com.datastax.spark.connector.rdd.partitioner.dht.TokenRange[Long, LongToken]
 
-  private type TR = TokenRange[Long, LongToken]
+  override def split(tokenRange: TokenRange, splitCount: Int): Seq[TokenRange] = {
+    val rangeTokenCount = tokenRange.distance
+    val splitPointsCount = if (rangeTokenCount < splitCount) rangeTokenCount.toInt else splitCount
```
RussellSpitzer: I'm a little confused why we are building this special case, Are we worried about a token range having a size of less than splitCount? (I mean this looks like the right thing to do but i'm not sure we need to do it?) I guess without this we have the worry of getting a SubRange that's empty...

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/TokenRangeSplitter.scala
```diff
@@ -1,16 +1,33 @@
...
+import scala.collection.parallel.ForkJoinTaskSupport
+import scala.concurrent.forkjoin.ForkJoinPool
 
 
+/** Splits a token ranges into smaller sub-ranges,
+  * each with the desired approximate number of rows. */
+private[partitioner] trait TokenRangeSplitter[V, T <: Token[V]] {
+
+  def split(tokenRanges: Iterable[TokenRange[V, T]], splitCount: Int): Iterable[TokenRange[V, T]] = {
+    val wholeRing = tokenRanges.map(_.ringFraction).sum
```
RussellSpitzer: Shoudln't this be 1?

spark-cassandra-connector/src/test/scala/com/datastax/spark/connector/rdd/partitioner/TokenRangeClustererTest.scala
```diff
@@ -52,50 +54,38 @@ class TokenRangeClustererTest {
...
-    val trc = new TokenRangeClusterer[Long, LongToken](10)
+    val trc = new TokenRangeClusterer[Long, LongToken](2)
     val groups = trc.group(Seq(tr1, tr2, tr3, tr4)).map(_.toSet).toSet
     assertEquals(2, groups.size)
     assertTrue(groups.contains(Set(tr1, tr2)))
     assertTrue(groups.contains(Set(tr3, tr4)))
   }
 
   @Test
-  def testTooLargeRanges() {
```
RussellSpitzer: Why remove this test?

RussellSpitzer:
I have a few minor questions, but I think this is all a great improvement.

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/CassandraJavaRDDSpec.scala
jtgrabowski: This is how `limit(...)` works. It limits the amount of RDD elements to `rdd.getNumPartitions * limit`.

Previously `getNumPartitions` was equal to `1`, so asserting ` result should have size 1` was fine.
Now, when there is a minimal threshold for partitions number, it gets a bit more complicated.
`val limit = 1` is as good as any I guess, should we change it some different value?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/TokenRangeSplitter.scala
jtgrabowski: That's a great question! I'm not 100% sure the answer :) 

Apparently there was a [case found](https://datastax.slack.com/archives/drivers/p1465394335000262) when `Metadata#getTokenRanges` did not return the whole ring. Don't know if this was confirmed.

[Tried to ask](https://datastax.slack.com/archives/drivers/p1465821536000002) with no luck. Going to do that again.

spark-cassandra-connector/src/test/scala/com/datastax/spark/connector/rdd/partitioner/TokenRangeClustererTest.scala
jtgrabowski: As `TokenRange` no longer carries `dataSize` I couldn't find a reasonable corresponding alternative to this test.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/TokenRangeSplitter.scala
jtgrabowski: Got the answer, this should be the whole ring. Going to change this to `1`.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/Murmur3PartitionerTokenRangeSplitter.scala
jtgrabowski: I agree either this or filtering out empty subranges. Let's leave it as it is, ok?

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/CassandraJavaRDDSpec.scala
RussellSpitzer: Huh i thought it should just be limit regardless of numPartitions. Don't we have a pushdown followed by a reduce?

jtgrabowski: We have a pushdown only. We push user-provided limit for each cql query we issue.

jtgrabowski:
@RussellSpitzer I applied your remarks (last one was to 1.0 as whole ring "fraction").

RussellSpitzer:
+1 (Can you also add a doc notes on how this is set up? )

jtgrabowski:
@RussellSpitzer  added info about miminal number of partitions in FAQ and reference.md, do you think it is sufficient?

RussellSpitzer:
+1

RussellSpitzer:
Although I haven't read the "partitioning.md" Because that isn't included :)

jtgrabowski:
squashed and rebased

PR #992, opened 2016-06-14
SPARKC-390: Replace org.apache.spark.Logging with internal Logging Trait

RussellSpitzer:
Spark 2.0 moves the logging trait to be fully internal this means the
connector can no longer use it. To avoid conflicts we extract the trait
as we were using it so that the 1.6 and future versions of the connector
will work with Spark 2.0.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/LocalNodeFirstLoadBalancingPolicy.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/LocalNodeFirstLoadBalancingPolicy.scala
@@ -11,7 +11,6 @@ import scala.collection.JavaConversions._
 import scala.util.Random
 
 import com.datastax.spark.connector.util.Logging
-import com.datastax.spark.connector.util.Logging
 
 /** Selects local node first and then nodes in local DC in random order. Never selects nodes from other DCs.
   * For writes, if a statement has a routing key set, this LBP is token aware - it prefers the nodes which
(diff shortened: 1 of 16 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/LocalNodeFirstLoadBalancingPolicy.scala
```diff
@@ -10,7 +10,8 @@ import java.net.{InetAddress, NetworkInterface}
 import scala.collection.JavaConversions._
 import scala.util.Random
 
-import org.apache.spark.Logging
+import com.datastax.spark.connector.util.Logging
+import com.datastax.spark.connector.util.Logging
```
jacek-lewandowski: duplicate

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/Logging.scala
```diff
@@ -0,0 +1,80 @@
...
+  * log level is enabled.
+  *
+  *
+  * This is a copy of what Spark Previously held in org.apache.spark.Logging. That class is
+  * now private so we will expose similar functionality here.
+  */
+trait Logging {
+  // Make the log field transient so that objects with Logging can
+  // be serialized and used on another machine
+  @transient private var log_ : Logger = null
```
jacek-lewandowski: I know Spark did it this way, but I'm strongly against underscores following the argument - can you please put the underscore before the name?

jacek-lewandowski:
Minor comments, +1

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/LocalNodeFirstLoadBalancingPolicy.scala
RussellSpitzer: I just really wanted to import it.

jacek-lewandowski: cool :)

jacek-lewandowski:
+1

PR #993, opened 2016-06-15
SPARKC-391: Fix conversion of LocalDate to Joda LocalDate

RussellSpitzer:
We transform LocalDate objects to JodaLocal date objects so that we'll
have a serializable AnyRef fo moving around in Spark. Previously this
was done by just using the default constructor which assumes the
timestamp provided is in the DefaultTimezone. This is incorrect at
certain times of the year in certain places since LocalDate will give
back the Timestamp in GMT.

To fix this we always pass GMT as the Timezone to convert LocalDates
with.


jacek-lewandowski:
Just repeat the test against at least 2 timezones.
+1

jacek-lewandowski:
+1

RussellSpitzer:
@jacek-lewandowski + @jtgrabowski I went a little crazy and refactored all the Date class stuff. Please review again :|

RussellSpitzer:
but it looks like it only works on my machine :(

RussellSpitzer:
retest this please

jtgrabowski:
Hey Russ I think I fixed the two tests that were failing. 
In one the main problem was that we switched time zone after test data was created and in the other one we were hitting a cached timezone in Catalyst. 
See https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala#L66 for details.
Theoretically we could do some magic with threads (since it is thread local) but few lines above there is also a `@transient lazy val defaultTimeZone = TimeZone.getDefault` which probably will cause other problems in the future.
For now only reliable solution I can think of is to take two different timezones and run tests against them in different JVMs. Please feel free to revert or change if you think I went to far :)
The build time increase is probably the main problem with this solution.
Gotta go, will test this more and do the `TODO:` on Monday, for now just wanted to share the progress.

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataFrameSpec.scala
```diff
@@ -200,61 +192,4 @@ class CassandraDataFrameSpec extends SparkCassandraITFlatSpecBase {
       session.execute(s"select count(1) from $ks.tuple_test2").one().getLong(0) should be (1)
     }
   }
-
-  it should "read and write C* LocalDate columns" in {
```
RussellSpitzer: This test was working for me on jenkins? Was it breaking for you?

RussellSpitzer:
Since i'm going to be afk for the next two weeks @jtgrabowski I'll leave this in your capable hands :)

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataFrameSpec.scala
jtgrabowski: The original version was working, but it was changed to this:
https://github.com/datastax/spark-cassandra-connector/pull/993/commits/e66fccc7be053309e25750f13047fdc63740ff37
Which was not passing on Jenkins - builds:1025, 1026

RussellSpitzer:
+1 Let's never talk about dates again

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/typeTests/DateTypeCETTest.scala
```diff
@@ -0,0 +1,15 @@
...
+  * Following tests are executed in separate JVM
+  */
+class DateTimeTypeCETTest extends DateTimeTypeTest(TimeZone.getTimeZone("CET")) {
+}
+
+class DateTypeCETTest extends DateTypeTest(TimeZone.getTimeZone("CET")) {
+}
+
+class SqlDateTypeCETTest extends SqlDateTypeTest(TimeZone.getTimeZone("CET")) {
+}
```
jacek-lewandowski: here and everywhere else - please add a new line at the and of file

jtgrabowski: done

PR #999, opened 2016-07-08
SPARKC-399: Removed CassandraSQLContext

jacek-lewandowski:


spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/cql/CassandraAuthenticatedConnectorSpec.scala
```diff
@@ -1,72 +1,65 @@
...
+      ("Andy", 28, "America"),
+      ("Kaushal", 25, "India"),
+      ("Desanth", 27, "India"),
+      ("Mahendra", 26, "Rajasthan"))).toDF("name", "age", "address")
+
+    createKeyspace(conn.openSession())
+    personDF1.createCassandraTable(ks, "authtest", Some(Array("address")), Some(Array("age")))(conn)
+
+    val options = Map("spark_cassandra_auth_username" -> "cassandra",
+      "spark_cassandra_auth_password" -> "cassandra",
```
alexliu68: Can you check whether we can use dot instead of _ for spark 2.0, _ is used in previous release because of the limit of spark SQL.

jacek-lewandowski:
@alexliu68 - yes, it works with dots now.

alexliu68:
+1

PR #1000, opened 2016-07-13
Requesting a pull to datastax:b1.6 from datastax:SPARKC-397

RussellSpitzer:
6d39d73 (Russell Spitzer, 80 seconds ago)
    SPARKC-397: Docs for null TTL and WRITETIME behavior

 2d77829 (Russell Spitzer, 13 minutes ago)
    SPARKC-397: Prevent Null Pointer Exceptions on TTL and WriteTime

    Both TTL and WriteTime cannot be set to null and for C* < 2.2 there is no
    way to leave the parameter unbound. To avoid null pointers we provide the
    following defaults unless modified by the end user.

    TTL is set to the table default_ttl WriteTime is set to currentTimeMillis *
    1000L on Insertion

    This has required some rather extensive refactoring which has changed
    several interfaces. The general thread is pass information via a ColumnRef
    down to the BoundStatementBuilder where it can read the metadata provided
    and take evasive manuevers when it sees a NULL in a non-nullable parameter.

    WriteOption has been extended so that perRow now also allows for an
    additional valueIfNull parameter which can be used to override the defaults
    specified above.

    Parallel extensions where made to the Java interface to allow for the same
    access from java.

 7abb242 (Russell Spitzer, 29 hours ago)
    SPARKC-397: Refactor RowWriter to Carry ColumnRef Information

    Previously only the string names of the columns were possessed by the
    underlying RowWriter. This made it impossible to change bind behavior based
    on the ColumnRef Type. This modification allows for us to act differently
    based on the underlying columnref created by the User.

### DIFF
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/writer/TableWriterSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/writer/TableWriterSpec.scala
@@ -14,7 +14,9 @@ import com.datastax.spark.connector.types._
 case class KeyValue(key: Int, group: Long, value: String)
 case class KeyValueWithTransient(key: Int, group: Long, value: String, @transient transientField: String)
 case class KeyValueWithTTL(key: Int, group: Long, value: String, ttl: Int)
+case class KeyValueWithNullableTTL(key: Int, group: Long, value: String, ttl: java.lang.Integer)
 case class KeyValueWithTimestamp(key: Int, group: Long, value: String, timestamp: Long)
+case class KeyValueWithNullableTS(key: Int, group: Long, value: String, timestamp: java.lang.Long)
 case class KeyValueWithConversion(key: String, group: Int, value: String)
 case class ClassWithWeirdProps(devil: String, cat: Int, value: String)
 case class Address(street: String, city: String, zip: Int)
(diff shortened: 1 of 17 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
```diff
@@ -348,4 +348,21 @@ object Schema extends Logging {
...
+  /**
+    * Get default TTL for Table
+    */
+  def ttlFromCassandra(
+    connector: CassandraConnector,
+    keyspaceName: String,
+    tableName: String) : Option[Int] = {
+
+    tableFromCassandra(connector, keyspaceName, tableName) // Checks for existence of table
+    Try(connector.withClusterDo(
```
RussellSpitzer: This may be the wrong thing to do, but without atomic metadata updates this can throw a null pointer on very recently created tables ... Maybe we have this sleep and retry?

vinaykumarchella:
Looks good to me.

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/writer/TableWriterSpec.scala
```diff
@@ -571,10 +573,51 @@ class TableWriterSpec extends SparkCassandraITFlatSpecBase {
...
+      KeyValueWithNullableTTL(3, 3L, "value3", 300))
+    sc.parallelize(col).saveToCassandra(ks, "key_value", writeConf = WriteConf(ttl = TTLOption.perRow("ttl")))
+
+    verifyKeyValueTable("key_value")
+
+    val resultMap = conn.withSessionDo { session =>
+      val result = session.execute(s"""SELECT key, TTL(value) FROM $ks.key_value""").all()
+      result should have size 3
+      result.map( row => (row.getInt(0), row.getInt(1))).toMap
+      }
```
jacek-lewandowski: nit: formatting

spark-cassandra-connector/src/main/java/com/datastax/spark/connector/japi/CassandraJavaUtil.java
```diff
@@ -1710,11 +1710,11 @@ public static ColumnName column(String columnName) {
     }
 
     public static TTL ttl(String columnName) {
```
jacek-lewandowski: perhaps you should also add the methods for ttl and writetime with specified default in Java

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/ColumnRef.scala
```diff
@@ -68,23 +68,42 @@ case class CollectionColumnName(
...
+case class TTL(
+  columnName: String,
+  alias: Option[String] = None,
+  writeDefault: Option[Int] = None) extends ColumnRef with ValueOnNull {
+
   override val cql = s"""TTL("$columnName")"""
   override val cqlValueName = s"ttl($columnName)"
   override def selectedAs = alias.getOrElse(cqlValueName)
   override def toString: String = cqlValueName
+  override def valueOnNull: AnyRef = writeDefault.getOrElse[Int](0): java.lang.Integer
```
jacek-lewandowski: shouldn't it be table default rather than 0?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Schema.scala
jacek-lewandowski: actually... why don't you add it as a another property in `TableDef`?

jacek-lewandowski:
I can see one old bug which can be fixed here or in another ticket - seems like when `queryTemplateUsingUpdate` is used TTL and WRITETIME are not considered at all.

jacek-lewandowski:
btw. @RussellSpitzer how do you create such a nice PR description ?

jacek-lewandowski:
In general it looks good. However I'm wondering whether we shouldn't make this a part of `RowWriter`. For me it is the most natural place to apply any defaults: `RowWriter` has a function `readColumnValues`:

```scala
  override def readColumnValues(data: CassandraRow, buffer: Array[Any]) = {
    for ((c, i) <- columnNames.zipWithIndex) {
      val value = data.getRaw(c)
      val convertedValue = converters(i).convert(value)
      buffer(i) = convertedValue
    }
  }
```

There could be a constant buffer with default values, so it could look like this:

```scala
  private val defaultValues = Array[Any]()
  
  override def readColumnValues(data: CassandraRow, buffer: Array[Any]) = {
    for ((c, i) <- columnNames.zipWithIndex) {
      val value = data.getRaw(c)
      val convertedValue = converters(i).convert(value)
      buffer(i) = if (convertedValue != null) convertedValue else defaultValues(i)
    }
  }
```

It would be more generic and would allow for replacing nulls in any column if needed in the future. I'm suggesting this because I slightly don't like having a default value in `ColumnRef` (TTL, WriteTime). Also, `RowWriter` is the place where the conversion happens and replacing nulls with some default value looks like a part of conversion chain. 
wdyt?

PR #1004, opened 2016-07-14
SPARKC-387: Have columntype dependent on C* Protocol Version

RussellSpitzer:
There is an error where the connector will attempt to use types that are not valid for the C* version which is connected to. This patch makes all Type decisions protocol version aware to fix backwards compatibility.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
@@ -92,7 +92,9 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
     rwf: RowWriterFactory[T],
     columnMapper: ColumnMapper[T]): Unit = {
 
-    val table = TableDef.fromType[T](keyspaceName, tableName)
+    val protocolVersion = connector.withClusterDo(_.getConfiguration.getProtocolOptions.getProtocolVersion)
+
+    val table = TableDef.fromType[T](keyspaceName, tableName, protocolVersion)
     saveAsCassandraTableEx(table, columns, writeConf)
   }
 
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/ColumnType.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/ColumnType.scala
@@ -2,26 +2,17 @@ package com.datastax.spark.connector.types
 
 import java.net.InetAddress
 import java.nio.ByteBuffer
-import java.util.{UUID, Date}
+import java.util.{Date, UUID}
 
 import org.apache.spark.SparkEnv
 import org.apache.spark.sql.catalyst.ReflectionLock.SparkReflectionLock
-
-import com.datastax.driver.core.{TupleType => DriverTupleType, UserType => DriverUserType, DataType}
+import com.datastax.driver.core.{DataType, ProtocolVersion, TupleType => DriverTupleType, UserType => DriverUserType}
+import com.datastax.driver.core.ProtocolVersion._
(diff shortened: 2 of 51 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/ColumnType.scala
```diff
@@ -157,7 +153,13 @@ object ColumnType {
   }
 
   /** Returns natural Cassandra type for representing data of the given Scala type */
-  def fromScalaType(dataType: Type): ColumnType[_] = {
+  def fromScalaType(
+    dataType: Type,
+    protocolVersion: ProtocolVersion = ProtocolVersion.NEWEST_SUPPORTED): ColumnType[_] = {
+
+    import ProtocolVersion._
+    def versionGTE( targetProtocolVersion: ProtocolVersion) : Boolean =
```
jacek-lewandowski: not very scala-ish... :)
maybe a value class with > and  < methods would be better?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/RefCountedCache.scala
```diff
@@ -28,7 +30,7 @@ final class RefCountedCache[K, V](create: K => V,
   private[cql] val deferredReleases = new TrieMap[V, ReleaseTask]
 
   private def createNewValueAndKeys(key: K): (V, Set[K]) = {
-    val value = create(key)
+    val value = RetryUtil.TryUntilValid(create(key))
```
jacek-lewandowski: why ?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
```diff
@@ -92,7 +92,9 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
     rwf: RowWriterFactory[T],
     columnMapper: ColumnMapper[T]): Unit = {
 
-    val table = TableDef.fromType[T](keyspaceName, tableName)
+    val protocolVersion = connector.withClusterDo(_.getConfiguration.getProtocolOptions.getProtocolVersion)
```
jacek-lewandowski: so why it is not retrieved from the configuration ? why do we have that configuration entry?

RussellSpitzer: This is the "actual" protocol version being used. The Configuration entry is for forcing an alternate PV. The Default "null" uses the "HighestSupportedVersion" by the driver.

Connecting to C* 3.X with the 3.0 Driver and the Parameter == null yields PV4
Connecting to C* 3.X with the 3.0 Driver and the Parameter == PV3 yields PV3
Connecting to C* 2.1.X with the 3.0 Driver and the Parameter == null yields PV3

I can remove this now though as I decided to just gate the whole test suite

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/ColumnType.scala
RussellSpitzer: It implements comparable? I could have that wrapped I think?

RussellSpitzer:
@jacek-lewandowski I think this is ready for review now, in the 2.1.15 build we should have 2 Failing tests, which I can fix in https://datastax-oss.atlassian.net/projects/SPARKC/issues/SPARKC-402 As these are not regressions. They are simply bugs in the code that previously existed

jtgrabowski:
+1

jtgrabowski:
can we merge this together with SPARKC-402 so that we have repo history without commits that do not build?

RussellSpitzer:
Merging with SPARKC-402

PR #1007, opened 2016-07-22
SPARKC-402: Partition Key Indexes Pushdown Incorrect for PV3

RussellSpitzer:
6f23b30 (Russell Spitzer, 34 minutes ago)
    SPARKC-402: PartitionKey Indexes Pushdown Incorrect for PV3

    In PV3 it is illegal to pushdown an equality predicate based on a secondary
    index for a partition key while also specifying inequalities on the token
    of that partition key. In PV4 it is legal. The pushdowns have now been
    changed to check the PV4 and skip equality pushdowns for partition keys if
    PV3 is in use.

 f82b544 (Russell Spitzer, 8 days ago)
    SPARKC-387: ColumnType Should be ProtocolVersion Aware

    The Connector currently always assumes it is connecting to the newest
    version of C*. This leads to some incorrect decisions when checking against
    ColumnTypes as PV4 Types are not valid in PV3. To fix this all of the
    ColumnType Selection code is now also checks the Protocol Version and
    chooses the Correct Type based Protocol Version instead.

    In addition, the Integeration tests have been version gated to allow for
    accurate testing against previous versions of Cassandra.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,4 +20,6 @@ before_cache:
 script:
   - "sbt ++$TRAVIS_SCALA_VERSION -Dtravis=true test"
   - "sbt ++$TRAVIS_SCALA_VERSION -Dtravis=true it:test"
+  - "sbt ++$TRAVIS_SCALA_VERSION -Dtest.cassandra.version=2.1.15 -Dtravis=true test"
+  - "sbt ++$TRAVIS_SCALA_VERSION -Dtest.cassandra.version=2.1.15 -Dtravis=true it:test"
   - "sbt ++$TRAVIS_SCALA_VERSION -Dtravis=true assembly"
--- a/spark-cassandra-connector/src/it/resources/2.1/cassandra-default.yaml.template
+++ b/spark-cassandra-connector/src/it/resources/2.1/cassandra-default.yaml.template
@@ -421,7 +421,7 @@ native_transport_port: ${native_transport_port}
 # native_transport_max_concurrent_connections_per_ip: -1
 
 # Whether to start the thrift rpc server.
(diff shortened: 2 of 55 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/RetryUtil.scala
```diff
@@ -0,0 +1,21 @@
+package com.datastax.spark.connector.util
+
+import scala.annotation.tailrec
+import scala.util.{Failure, Success, Try}
+
+object RetryUtil {
```
jtgrabowski: Would [Eventually](http://www.artima.com/docs-scalatest-2.0.M5/org/scalatest/concurrent/Eventually.html) be ok instead?

jtgrabowski:
+1 and minor comment, feel free to merge

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/util/RetryUtil.scala
RussellSpitzer: yep :) I thought I would need something like this in live code, but you are right we don't need this currently.

RussellSpitzer:
@jacek-lewandowski + @jtgrabowski Do you want me to just rebase this and merge it?

jtgrabowski:
I thought that 387 (among other things) introduces tests that fail and this one solves the problem. Let's not ignore 387, let's merge them together.
Are there any showstoppers for 387?

RussellSpitzer:
I don't think so, I"ll do this rebase

PR #1009, opened 2016-07-26
SPARKC-405: Disallow TimeUUID Predicate Pushdown

RussellSpitzer:
Timeuuids are compared differently in Spark and in Cassandra. In
Cassandra Timeuuids are compared bas on the Time portion of the binary
data. In Spark they are compared lexically (in byte order). In Spark <
1.6 all filters in Dataframes are applied twice, once at the DataSource
and once in Spark. This double application would be done using both of
the above comparitors leading to incorrect results. We will now throw
exceptions if we sense an end user is trying to perform one of these
impossible comparisons.


jacek-lewandowski:
+1

PR #1010, opened 2016-07-26
SPARKC-405: B1.6

RussellSpitzer:
Throws an exception when trying to do a Spark(lexical) comparison on a TimeUUID, but allows comparisons if C* handles the entire comparison(Compared based on time).


jacek-lewandowski:
+1

PR #1017, opened 2016-08-09
SPARKC-370 saving DFs with camel cased fields into underscored C* cols

jtgrabowski:


jtgrabowski:
Russ, first commit add camelcased DataFrame fields conversion to underscored C* columns. This is always on. 

Two later commits add an option to convert underscored C* column fields to camelcased DataFrame fields. 
Actually the last commit adds only one test. It verifies whether we implement column aliasing for all the filters, that Spark provides. It is rather crud (hardcoded number of known filters) but gives hope that we are not going to miss new filters in future Spark versions.

Also I though about using reflection `AliasedCassandraSourceRelation.aliasColumnName` but I could not see any benefits over proposed solution.

jtgrabowski:
Yes. 
I thought that you are on board with this :) Should we close as "won't fix"?

RussellSpitzer:
I'm a little iffy on putting in automatic casing stuff in DataFrames, but I think keeping it consistent is important. Can I let you make the call? :)

RussellSpitzer:
At least one more user requesting this, http://stackoverflow.com/questions/40187480/datastax-spark-cassadra-connector-writing-df-to-cassandra-table

etspaceman:
+1

drewrobb:
Is there any chance of this getting merged? It would be useful for me.

PR #1018, opened 2016-08-11
SPARKC-413: Avoid overflow on SplitSizeInMB param

RussellSpitzer:
Previously the split_size parameter was an Integer and was computed by
multiplying split_size_in_mb * 1024 * 1024 which will overflow for
values greater than 2047. Changing this to a Long moves this threshold
to 2097152. Hopefully in 10 years I'll look back and say what a small
amount of ram, 2TB per spark partition.


pkolaczk:
+1

PR #1019, opened 2016-08-12
SPARKC-312: Implementing FilterOptimizer

ponkin:
FilterOptimizer will to transform `where` clause to equivalent disjunction normal form. 
For example where clause `a = 5 and (b > 5 or b < 3)` can be transformed to equivalent `a = 5 and b > 5 or a =5 and b < 3`, so now we can create two different table scans with where clause `a = 5 and b > 5` and `a =5 andb < 3` and union them.

### DIFF
--- a/spark-cassandra-connector/src/it/scala/org/apache/spark/sql/CassandraPrunedFilteredScanSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/org/apache/spark/sql/CassandraPrunedFilteredScanSpec.scala
@@ -32,6 +32,12 @@ class CassandraPrunedFilteredScanSpec extends SparkCassandraITFlatSpecBase with 
             s"""CREATE TABLE IF NOT EXISTS $ks.fields
                 |(k INT, a TEXT, b TEXT, c TEXT, d TEXT, e TEXT, PRIMARY KEY (k)) """
                 .stripMargin)
+        },
+        Future {
+          session.execute(
+            s"""CREATE TABLE IF NOT EXISTS $ks.metrics
+                |(k TEXT, a INT, b INT, c INT, PRIMARY KEY (k, a, b)) """
+              .stripMargin)
         }
       )
     }
@@ -39,6 +45,7 @@ class CassandraPrunedFilteredScanSpec extends SparkCassandraITFlatSpecBase with 
 
   val colorOptions = Map("keyspace" -> ks, "table" -> "colors")
   val fieldsOptions = Map("keyspace" -> ks, "table" -> "fields")
+  val metricsOptions = Map("keyspace" -> ks, "table" -> "metrics")
   val withPushdown = Map("pushdown" -> "true")
   val withoutPushdown = Map("pushdown" -> "false")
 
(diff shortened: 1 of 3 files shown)


spark-cassandra-connector/src/it/scala/org/apache/spark/sql/CassandraPrunedFilteredScanSpec.scala
```diff
@@ -32,14 +32,22 @@ class CassandraPrunedFilteredScanSpec extends SparkCassandraITFlatSpecBase with
...
         }
       )
     }
   }
 
   val colorOptions = Map("keyspace" -> ks, "table" -> "colors")
   val fieldsOptions = Map("keyspace" -> ks, "table" -> "fields")
+  val metricsOptions = Map("keyspace" -> ks, "table" -> "metrics")
   val withPushdown = Map("pushdown" -> "true")
+  val withWhereClauseOptimizationEnabled = Map("spark.cassandra.sql.enable.where.clause.optimization" -> "true")
```
RussellSpitzer: Replace the string here with the parameter defined in the conf file 
`EnableWhereClauseOptimizationParam.name` just in case we change things later :)

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -79,7 +78,15 @@ private[cassandra] class CassandraSourceRelation(
   def buildScan(): RDD[Row] = baseRdd.asInstanceOf[RDD[Row]]
 
   override def unhandledFilters(filters: Array[Filter]): Array[Filter] = filterPushdown match {
-    case true => predicatePushDown(filters).handledBySpark.toArray
+    case true => 
+      val optimizedFilters = FiltersOptimizer(filters).build()
```
RussellSpitzer: I think it may be better if we got the FilterOptimzer into the predicatePushDown function, then I think we could skip having it written in a bunch of places.

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/FiltersOptimizer.scala
```diff
@@ -0,0 +1,106 @@
...
+  * will become equivalent
+  * 'field1 < 3 AND field2 = "val1" OR field1 < 3 AND field2 = "val2" OR
+  * field1 > 7 AND field2 = "val1" OR field1 > 7 AND field2 = "val2"'
+  *
+  * @param filters Array of logical statements [[org.apache.spark.sql.sources.Filter]]
+  * that forms `where`-clause with `AND` operator, for example:
+  * val Array(f1, f2, ... fn) = ... // such that `where f1 AND f2 AND ... AND fn`
+  *
+  */
+class FiltersOptimizer(filters: Array[Filter]) {
```
RussellSpitzer: I'm a little confused why there is a separate class here, do we ever use this without calling .build() immediately after?

spark-cassandra-connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -125,12 +132,29 @@ private[cassandra] class CassandraSourceRelation(
...
+
   override def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] = {
     val prunedRdd = maybeSelect(baseRdd, requiredColumns)
     val prunedFilteredRdd = {
       if(filterPushdown) {
-        val pushdownFilters = predicatePushDown(filters).handledByCassandra.toArray
-        val filteredRdd = maybePushdownFilters(prunedRdd, pushdownFilters)
+        val optimizedFilters = new FiltersOptimizer(filters).build()
+        val optimizationCanBeApplied = isOptimizationAvailable(optimizedFilters)
+        val filteredRdd = if(optimizationCanBeApplied) {
```
RussellSpitzer: I think this is a rather dangerous optimization sometimes, so I think we should default to off. For example

Table where x < 3 or x > 5 and x ranges from 1 to 10000. Doing two scans here is probably much more expensive than a single scan.

ponkin: I thought that both scans will be done in parallel.

PR #1020, opened 2016-08-17
SPARKC-355

RussellSpitzer:
Request cfca49e (Russell Spitzer, 26 hours ago)
    SPARKC-355: Wire publish and publishLocal to the new package task

    With this last change the default connector project is basically a
    distribution project. It contains the shaded marked as provided so pulling
    from maven will not pull the shaded guava or netty. It's publish tasks all
    go through the new packageBin which ends up running the assembly task from
    the Shaded project.

 5b8ee42 (Russell Spitzer, 27 hours ago)
    SPARKC-355: Update to Released Spark Packages

    The `connector` project now only includes the shaded libraries as test,
    this means the resultant POM file wil not cause end users to pull down our
    Guava by default.

 2dbea19 (Russell Spitzer, 5 days ago)
    SPARKC-355: Remove Default Artifact, Publish Only Shaded Jars

    Inorder to make sure there is no confusion about what Jar an end user
    should be including in their build file, the only artifact that will be
    published under the name spark-cassandra-connector will be the shaded jar.
    This means all package and packageBin tasks are essentially noops.

    publishLocal, and publish will now use the Shaded jar as the the
    spark-cassandra-connector artifact.

 177bf38 (Russell Spitzer, 8 days ago)
    SPARKC-355: Add shadedConnector Project

    Adds an additional project which mirrors the normal connector build project
    except it marks ALL libraries except those we are explictly shading as
    provided. This makes it so it's assembly task will only include the shaded
    jars we want and no other dependencies.

    This assembly task which only includes shaded deps is then set as the
    assembly target of spPackages which in coordination with spShade changes
    the spDist task to use the shaded assembly jar as it's target rather than
    the standard assembly jar with all dependencies.

 3a8947a (Russell Spitzer, 8 days ago)
    SPARKC-355: Explicitly use netty-all at a specified version

    To add a bit more explictness to our build, all netty sub components are
    excluded and instead we will rely on the netty-all package for the maximum
    version required by the Cassandra Java Driver. This jar will also be
    shaded.

### DIFF
--- a/project/SparkCassandraConnectorBuild.scala
+++ b/project/SparkCassandraConnectorBuild.scala
@@ -20,6 +20,11 @@ import java.io.File
 import pl.project13.scala.sbt.JmhPlugin
 import sbt.Keys._
 import sbt._
+import sbt.Keys._
+import sbtassembly._
+import sbtassembly.AssemblyKeys._
+import sbtsparkpackage.SparkPackagePlugin.autoImport._
+import pl.project13.scala.sbt.JmhPlugin
 
 object CassandraSparkBuild extends Build {
   import Settings._
@@ -34,7 +39,7 @@ object CassandraSparkBuild extends Build {
     name = "root",
     dir = file("."),
     settings = rootSettings ++ Seq(cassandraServerClasspath := { "" }),
-    contains = Seq(embedded, connector, demos)
+    contains = Seq(embedded, connectorDistribution, demos)
   ).disablePlugins(AssemblyPlugin, SparkPackagePlugin)
 
   lazy val cassandraServerProject = Project(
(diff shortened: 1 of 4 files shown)


project/SparkCassandraConnectorBuild.scala
```diff
@@ -270,11 +348,36 @@ object Dependencies {
...
+    *
+    * In the shaded build we will place theses in the assembly jar and remove all other libs
+    */
+  val shaded = Seq(guava, cassandraDriver)
+
+  val connector = testKit ++
+    metrics ++
+    jetty ++
+    logging ++
+    akka ++
```
jacek-lewandowski: akka ?

RussellSpitzer:
NOT USING THIS ONE

PR #1023, opened 2016-08-23
SPARKC-355: Have all public artifacts use the Shaded Jar

RussellSpitzer:
High Level Summary:
 Cassandra Java Driver and Shaded Guava are in the release artifacts
 No Shaded Netty (Did not seem to conflict)
 Package, Publish* tasks push shaded artifacts

Details

This patch publishes artifacts as a jar with the Cassandra
Java Driver included and guava shaded. This is a necessity because of
guava conflicts in Spark 2.0. Essentially any references to Guava that
the driver uses must be to a Guava version greater than 16. We
accomplish this by shading all references within the driver which we
also include. This makes it impossible to change the C* java driver
used with the connector. This is not a large issue because any Cassandra
Java Driver without shaded guava on the  classpath will throw incompatible guava errors.

To distribute this artifact the connector project (now connectorDistribution)
has several sibling-projects.

To produce the shadedJar we add shadedConnector. This project marks
all the dependencies of the project as provided except for the shaded
dependencies. This causes its assembly task to produce a jar which only
contains the Connector classes, Shaded Guava, and a Java Driver which
refers to the shaded guava. The assembly of this project is used as the
packageBin method in the connectorDistribution as well as an artifact
in the publish* tasks.

The connector distribution POM now no longer needs compile configuration
for the shaded dependencies since they will already be in the jar. These
deps are marked as provided but but this also removes them from the
assembly task within connectorDistribution.

To fix this we add another project fullConnector. This project contains all of the
dependencies required to run the connector and the shaded libraries are
not marked as provided. This produces an assembly task with everything
needed. The artifact produced by the normal build of this project is
identical to our old `connector` project where the connector.jar
requires seperate guava and the java driver. The assembly task produces
the full shaded jar with all dependencies and is used as the assembly
task in connectorDistribution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ See [Building And Artifacts](doc/12_building_and_artifacts.md)
   - [Partitioner](doc/16_partitioning.md)
   - [Frequently Asked Questions](doc/FAQ.md)
   - [Configuration Parameter Reference Table](doc/reference.md)
+  - [Tips for Developing the Spark Cassandra Connector](doc/developers.md)
 
 ## Online Training
 ### DataStax Academy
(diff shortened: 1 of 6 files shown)


pkolaczk:
+1

RussellSpitzer:
Last changes made, 
"shaded" variable refactored to "includedInShadedJar"
netty included as dependency

PR #1024, opened 2016-08-31
Set PackagedArtifact instead of PackagedArtifacts

RussellSpitzer:
Advice from some helpful folks on sbt/sbt gitter noted that we should be
setting packagedArtifact in pacakageBin to reassign the publish
artifacts.


pkolaczk:
+1

pkolaczk:
It works, I can see that after uploading to staging repo in nexus, the jar is 5.68 MB large.

PR #1027, opened 2016-09-12
SPARKC-419: Fix partitioner on Integer.MIN_VALUE

jcharlytown:
Tokens might give Integer.MIN_VALUE as hashcode. This will lead to a negative
index into the set of replicas for the corresponding key and fail the task.

The reason is that Math.abs() returns a negative value for Integer.MIN_VALUE
and % gives a negative result in case the dividend is negative.

To fix this the partitioner now applies Math.abs() to the result of the modulo
calculation instead of to the token hash (the dividend). This guarantees that
the index is never negative.

### DIFF
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/PartitionedCassandraRDDSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/PartitionedCassandraRDDSpec.scala
@@ -71,6 +71,10 @@ class PartitionedCassandraRDDSpec extends SparkCassandraITFlatSpecBase {
           session.executeAsync(ps.bind(value: JInt, value: JInt, (rowCount - value): JInt))
         }
         results.map(_.get)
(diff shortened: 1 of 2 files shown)


jcharlytown:
Added a test to demonstrate successful handling of a key which in the end hashes to Integer.MIN_VALUE. It's the opposite of minimalistic, however, I couldn't think of a more elegant solution of making sure that the test actually covers an affected key.

Three more questions:

1. Is this the correct branch, or rather b1.6? 
2. Do you want me to squash the commits?
3. The readme says to include a note in CHANGES.txt. I am unsure whether or not this happens in the bugfixes themselves or with the corresponding release. Can someone clarify?

RussellSpitzer:
Don't worry about Changes.TXT we usually update that right before release

jcharlytown:
Thanks for reviewing, I'll try to include the requested changes today or tomorrow.

jcharlytown:
Rebased and squashed.

PR #1032, opened 2016-09-14
SPARKC-424: Refactor Cassandra runner to use template transformations

jacek-lewandowski:


spark-cassandra-connector-embedded/src/main/scala/com/datastax/spark/connector/embedded/CassandraRunner.scala
```diff
@@ -67,14 +65,21 @@ private[connector] class CassandraRunner(val configTemplate: String, props: Map[
       logConfigFileProperty, "-cp", classPath, cassandraMainClass, "-f")
     .inheritIO()
     .start()
+
   val startupTime = System.currentTimeMillis()
 
-  val nativePort = props.get("native_transport_port").get.toInt
-  if (!waitForPortOpen(InetAddress.getByName(props.get("rpc_address").get), nativePort, 100000))
-    throw new IOException("Failed to start Cassandra.")
+  if (!waitForPortOpen(InetAddress.getByName(baseConfiguration.rpcAddress), baseConfiguration.nativeTransportPort, 100000, () => !process.isAlive)) {
```
jtgrabowski: why do we need to pass `!process.isAlive` here instead of just checking before `waitForPortOpen`?

jacek-lewandowski: because Cassandra may be starting but it may not start. Say we started the process, Cassandra needs some time to startup, we passed the condition "if process is alive", and we are waiting for 100 seconds to acquire the connection. But Cassandra could die in the meantime, so there is no point in waiting any longer.

jtgrabowski:
LGTM. I like this PR's stats `+491 −6,544` :)

RussellSpitzer:
+1

PR #1035, opened 2016-09-19
[SPARKC-427] Allow Java Object Mapped Beans with Annotations to have the name mappings honored

hasSoumya:
New to Scala. Feel free to comment liberally :)

### DIFF
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/CassandraRDDSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/CassandraRDDSpec.scala
@@ -1,19 +1,20 @@
 package com.datastax.spark.connector.rdd
 
+import com.datastax.driver.core.ProtocolVersion._
+import com.datastax.spark.connector._
+import com.datastax.spark.connector.cql.CassandraConnector
+import com.datastax.spark.connector.embedded.YamlTransformations
+import com.datastax.spark.connector.mapper.{DefaultColumnMapper, JavaBeanColumnMapper, JavaTestBean, JavaTestUDTBean}
+import com.datastax.spark.connector.types.{CassandraOption, TypeConverter}
+
 import java.io.IOException
 import java.util.Date
+
+import org.joda.time.{DateTime, DateTimeZone, LocalDate}
 
 import scala.collection.JavaConversions._
 import scala.concurrent.Future
 import scala.reflect.runtime.universe.typeTag
-import org.joda.time.{DateTime, DateTimeZone, LocalDate}
-import com.datastax.spark.connector._
-import com.datastax.spark.connector.cql.CassandraConnector
-import com.datastax.spark.connector.embedded.YamlTransformations
-import com.datastax.spark.connector.mapper.DefaultColumnMapper
-import com.datastax.spark.connector.types.{CassandraOption, TypeConverter}
-
-import com.datastax.driver.core.ProtocolVersion._
 
 case class KeyValue(key: Int, group: Long, value: String)
 case class KeyValueWithConversion(key: String, group: Int, value: Long)
@@ -119,6 +120,11 @@ class CassandraRDDSpec extends SparkCassandraITFlatSpecBase {
         session.execute( s"""CREATE TYPE $ks.address (street text, city text, zip int)""")
         session.execute( s"""CREATE TABLE $ks.udts(key INT PRIMARY KEY, name text, addr frozen<address>)""")
         session.execute( s"""INSERT INTO $ks.udts(key, name, addr) VALUES (1, 'name', {street: 'Some Street', city: 'Paris', zip: 11120})""")
+      },
+
+      Future {
+        session.execute( s"""CREATE TYPE $ks.nested (field int, another_field int, yet_another_field int)""")
+        session.execute( s"""CREATE TABLE $ks.udts_nested(property_1 INT PRIMARY KEY, camel_case_property text, nested frozen<nested>)""")
       },
 
       Future {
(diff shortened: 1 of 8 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/mapper/ReflectionColumnMapper.scala
```diff
@@ -30,6 +29,26 @@ abstract class ReflectionColumnMapper[T : ClassTag] extends ColumnMapper[T] {
       paramName: String,
       columns: Map[String, ColumnRef]): Option[ColumnRef]
 
+  protected def annotationForFieldName(
```
RussellSpitzer: Not sure why this function def is here and not in the JavaMBean mapper since it is only used there

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/rdd/CassandraRDDSpec.scala
```diff
@@ -511,6 +522,43 @@ class CassandraRDDSpec extends SparkCassandraITFlatSpecBase {
...
+    row.getInt(0) should be(1)
+    row.getInt(1) should be(2)
+
+    val udtValue = row.getUDTValue(2)
+    udtValue.size should be(3)
+    udtValue.getInt("field") should be(3)
+    udtValue.getInt("another_field") should be(4)
+    udtValue.getInt("yet_another_field") should be(5)
+
+  }
```
RussellSpitzer: I think we should also enable reading back into the POJO that we just wrote from. Ie on Read we should use the same annotation structure for mapping

hasSoumya: I thought of it, and then I left it for another day. I'll work on writing those tests.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/mapper/ReflectionColumnMapper.scala
hasSoumya: The reasoning I had was because this function was doing 'reflection'y stuff, it should belong to the class that claims to be the owning that functionality. I don't have opinions to offer on which way we should go. Happy to move it to the sub class.

hasSoumya:
@RussellSpitzer - made some changes and added an extra test to validate column mappings. Please take a look.

bcantoni:
Triage: Russ will review the latest changes. @hasSoumya could you also please include a user doc update to explain this?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/mapper/JavaBeanColumnMapper.scala
```diff
@@ -28,6 +34,24 @@ class JavaBeanColumnMapper[T : ClassTag](columnNameOverride: Map[String, String]
     method.getParameterTypes.size == 1 &&
     method.getReturnType == Void.TYPE
 
+  private def annotationForFieldName(fieldName: String): String = {
```
RussellSpitzer: This may just be the monad lover in me, but if this returned Option[String]

We could be a little more functional and the above code for field name could be

`annotationName getOrElse fieldName`

Also this bit can be changed to

```
val maybeField = Try(cls.getField(fieldName))
maybeField.toOption.flatMap {
  case f =>
    if (f.isAnnotationPresent(classOf[Column])) {
      Some(f.getAnnotation(classOf[Column]).name())
  } else if (f.isAnnotationPresent(classOf[Field])) {
    Some(f.getAnnotation(classOf[Field]).name()) 
  } else None
}
```

You could also probably change the internal calls to 
`(Try(f.getAnnotation(classOf[Column]).name()) orElse Try(f.getAnnotation(classOf[Field]).name())).toOption`

Which would change the whole thing to

```
val maybeField = Try(cls.getField(fieldName))
maybeField.flatMap( f => 
  (
    Try(f.getAnnotation(classOf[Column]).name())  orElse 
    Try(f.getAnnotation(classOf[Field]).name())
   ).toOption
```

RussellSpitzer: That said I'm fine if you just want to change the return of this function to Option[String] :) That's the only thin I think is important here

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
```diff
@@ -50,7 +50,7 @@ trait TypeConverter[T] extends Serializable {
 }
 
 /** Handles nullable types and converts any null to null. */
-trait NullableTypeConverter[T <: AnyRef] extends TypeConverter[T] {
+trait NullableTypeConverter[T] extends TypeConverter[T] {
```
RussellSpitzer: This has been proposed before, But i'm not quite sure why this needs to happen. Could you explain it please?

spark-cassandra-connector/src/test/java/com/datastax/spark/connector/mapper/JavaTestBean.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.io.Serializable;
+
+/**
+ * This is a Java Bean style class with Java Driver style annotations built in.
+ * This class also contains nested UDTs with its own mappings
+ */
+public class JavaTestBean implements Serializable {
+
+    @Column(name = "property_1")
```
RussellSpitzer: I know this is unnecessary but touched we make these a little more different. property1 -> "cassandraproperty1" 
Just so the mapping is much clearer when we look back :)

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/mapper/JavaBeanColumnMapper.scala
hasSoumya: I can do it the concise `Scala` way :) - I'm new to the language anyway.

spark-cassandra-connector/src/test/java/com/datastax/spark/connector/mapper/JavaTestBean.java
hasSoumya: Can do.

hasSoumya:
@RussellSpitzer I'll make the proposed changes today. However, as for moving the function around, I originally had it as a `protected function` in the `ReflectionColumnMapper` class. You suggested in the first CR to move it closer to the `JavaBeanColumnMapper`. I'm okay going either direction. Having it in the `ReflectionColumnMapper` made more sense to me anyway.

RussellSpitzer:
Yep! I just thought It would be nice if the reflection mapper also did the annotation lookup not just have the function :D This way it would work on classes being analyzed by the ReflectionColumnMapper as well!

bcantoni:
@hasSoumya, we're rolling out a Contributor License Agreement for the Spark Cassandra Connector and need your signature to be able to use your contributions.

Please [follow instructions](http://spark-cassandra-connector-cla.datastax.com/) to sign, the whole process is completely electronic and will only take minutes.

Don't hesitate to <a href="mailto:cla@datastax.com?subject=Spark%20Cassandra%20Connector%20CLA">email cla at datastax.com</a> if you have any additional questions.

hasSoumya:
@bcantoni @RussellSpitzer I finally have time today. I will try and close out all pending work on this today. Thanks for your patience.

bcantoni:
CLA has been received, thanks

hasSoumya:
@RussellSpitzer Addressed all your concerns. Please take another look.

RussellSpitzer:
Thanks @hasSoumya ! It looks like there is a test failure on jenkins though

> [info] - should allow to save UDT columns from mapped Java objects and read them as UDTValue or Java objects *** FAILED *** (6 milliseconds)
> [info]   java.lang.IllegalArgumentException: requirement failed: Columns not found in class com.datastax.spark.connector.mapper.JavaTestBean: [cassandra_property_1, cassandra_camel_case_property]

hasSoumya:
@RussellSpitzer I could have sworn I ran this test. I have fixed it. It has to do with the visibility of the fields when Reflection is invoked. Not sure how this ever succeeded. 🎱

RussellSpitzer:
+1 Merging, We are going to need a brief doc section on this but we can do that in another PR

hasSoumya:
@RussellSpitzer Let me know the specifics, I can add it.

RussellSpitzer:
Sure! I was just thinking a little section in the https://github.com/datastax/spark-cassandra-connector/blob/master/doc/6_advanced_mapper.md Page describing how the connector will use annotations provided on objects from the Java Driver :)

hasSoumya:
I'll make a PR as soon as I get time.

RussellSpitzer:
Awesome, Thanks again for your contribution!

PR #1041, opened 2016-09-29
SPARKC-428: Better error messages on CreateCassandraTable from DataFrame

RussellSpitzer:
Give a useful error message when creating a table and a column specified
to be a partition key or clustering key cannot be found. Previously the
user would get a criptic error on a None type.


jtgrabowski:
+1

PR #1042, opened 2016-10-18
SPARKC-433: Correct order of "Limit" and "orderBy" clauses in JWCT

RussellSpitzer:
JoinWithCassandraTable places limit and order queries in the wrong
order. Thanks to bryanventeicher@gmail.com for pointing this out.


alexliu68:
+1

PR #1043, opened 2016-10-19
SPARKC-299 append data to collection column

alexliu68:


RussellSpitzer:
I guess we need to open this up again? I'm still not super happy with the api, I wonder if we can brainstorm a better way to specify the collection behavior.

schwartzmx:
Just curious if this is still in the works?  I need this functionality but have to resort to going from dataframe to rdd to use `saveToCassandra`

PR #1044, opened 2016-10-24
SPARKC-435: Allow Custom Conf Options for CustomPushdowns

RussellSpitzer:
Previously we removed all options to be passed through to the
CassandraSource that were not defined in the connector. This made it
impossible for a user supplied pushdown class to use per Table options
or Spark Conf settings.

This commit also breaks the custom pushdown rules api by passing through the
consolidated SparkConf which includes TableOptions and the Spark Conf.


alexliu68:
+1

forainychen:
Hi, Anyone have some documents on how to use such custom predicate pushdown feature? Thanks.

PR #1046, opened 2016-11-08
Reduce data copy and use unsafe rows

artem-aliev:


alexliu68:
Since Cassandra java driver row is converted to spark types, using InternalRow can reduce the double copy/converting, my only concern is https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/sources/interfaces.scala#L222.  Can we add a per table setting to allow use to choose InternalRow implementation or public API row implementation?

PR #1051, opened 2016-11-17
SPARKC-441 Fix CassandraConnector sessionCache remote and race condit…

alexliu68:
…ion issue

If set up a remote C* cluster and submit a job from spark cluster to
connect to remote C* cluster. The CassandraConnectorConf is serialized
and deserialized into different CassandraConnectorConf object for
the same original CassandraConnectorConf. Then each time sessionCache
creates new session for each deserialized CassandraConnectorConf.

Too many new sessions results too many cluster objects, and the
following errors are found.

ResourceLeakDetector: LEAK: You are creating too many HashedWheelTimer instances

To fix the issue, we change sessionCache key from CassandraConnectorConf
to String which is the serialized CassandraConnectorConf in String format.

The second issue is acquire method of RefCountedCache is not synchronized.
Even the cache itself is TriMap, the acquire method still creates two
sessions at certain moment when there are two concurrent requests to
acquire a new session.

To fix the issue, the code of creating new session in acquire method
when the session is not found in the cache should be synchronized.

It does have some performance overhead when new session is created
and serialize/deserialize CassandraConnectorConf

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectorConf.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectorConf.scala
@@ -44,7 +44,7 @@ case class CassandraConnectorConf(
 
   override def equals(obj: Any): Boolean = {
     obj match {
-      case that: CassandraConnectorConf => that.hashCode == serializedConfString.hashCode
+      case that: CassandraConnectorConf => that.serializedConfString == serializedConfString
       case _ => false
     }
   }
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectorConf.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectorConf.scala
@@ -1,6 +1,8 @@
 package com.datastax.spark.connector.cql
 
 import java.net.InetAddress
+import java.io.{ObjectOutputStream, ObjectInputStream, ByteArrayOutputStream, ByteArrayInputStream}
+import java.util.Base64
 
 import scala.language.postfixOps
 import scala.util.control.NonFatal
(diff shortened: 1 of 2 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/RefCountedCache.scala
```diff
@@ -53,6 +53,18 @@ final class RefCountedCache[K, V](create: K => V,
...
+  }
+
+  private[this] def syncAcquire(key: K): V = synchronized {
+    cache.get(key) match {
+      case Some(value) =>
+        if (refCounter.acquireIfNonZero(value) > 0)
+          value
+        else
+          acquire(key)
+      case None =>
```
artem-aliev: Hm, I can not comment the code that out of diff,so:
76 line: The code should not be reachable in sync block as I understand. SO let replace that block with the
ConcurentModificationException.

alexliu68: Sure

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectorConf.scala
```diff
@@ -27,7 +29,26 @@ case class CassandraConnectorConf(
...
+    oos.writeObject(this);
+    oos.close;
+    Base64.getEncoder.encodeToString(baos.toByteArray)
+  }
+
+  override def hashCode: Int = serializedConfString.hashCode
+
+  override def equals(obj: Any): Boolean = {
+    obj match {
+      case that: CassandraConnectorConf => that.hashCode == serializedConfString.hashCode
```
artem-aliev: String hashCode is not crypto hash, so unfortunately that line should compare string:
that = serializedConfString

artem-aliev:
+1 with the equals fix

PR #1064, opened 2016-12-19
SPARKC-363: Classloader issues with UDTs on different Classloaders than SCC

RussellSpitzer:
9e111e1 (Russell Spitzer, 3 hours ago)
    SPARKC-363: Find Converters for Classes on Different Classloaders

    The current code assumes that if you are converting a UDT in a collection
    or some other nested structure that all subclasses can be loaded on the
    same classloader than the SCC has been loaded on. For some runtime
    enivronments (like DSE) this is not true and the SCC is on a different
    classloader than the application code. This can lead to the lookup on the
    contained class failing with a ClassNotFound Exception.

    To avoid this situation we allow for the first class (which will be tuned
    to the correct classloader in it's runtime mirror) to force all
    sub-converters created to enable it's construction to be built with the
    same Classloader.

    This fixes situations like

    case class Row(k: Int, v: Seq[UdtCaseClass])

    so that

    The Classloader which loads "Row" will be used to load "UdtCaseClass".


RussellSpitzer:
NOTE TO REVIEWERS: This can only be triggered by having the classes to be converted in a different classloader (and not present on the SCC classloader). 

The easiest way to repo the issue is to do the spark-shell example from the Jira.

fabienmifsud:
Hi there, is there any planned release version for this issue ?
Do you have a workaround to wait for the release ?

PR #1070, opened 2017-01-10
SPARKC-441 Fix CassandraConnector sessionCache remote and race condition

alexliu68:
If set up a remote C* cluster and submit a job from spark cluster to
connect to remote C* cluster. The CassandraConnectorConf is serialized
and deserialized into different CassandraConnectorConf object for
the same original CassandraConnectorConf. Then each time sessionCache
creates new session for each deserialized CassandraConnectorConf.

Too many new sessions results too many cluster objects, and the
following errors are found.

ResourceLeakDetector: LEAK: You are creating too many HashedWheelTimer instances

To fix the issue, we change sessionCache key from CassandraConnectorConf
to String which is the serialized CassandraConnectorConf in String format.

The second issue is acquire method of RefCountedCache is not synchronized.
Even the cache itself is TriMap, the acquire method still creates two
sessions at certain moment when there are two concurrent requests to
acquire a new session.

To fix the issue, the code of creating new session in acquire method
when the session is not found in the cache should be synchronized.

It does have some performance overhead when new session is created
and serialize/deserialize CassandraConnectorConf


RussellSpitzer:
👍

alexliu68:
test it please

alexliu68:
retest this please

PR #1072, opened 2017-01-24
Requesting a pull to datastax:master from datastax:SPARKC-459

RussellSpitzer:
Write a message for this pull request. The first block
of text is the title and the rest is the description.

Changes:

550eed4 (Russell Spitzer, 2 minutes ago)
   SPARKC-459: Creation of CustomTableScan Extension Point

   Users can now extend the CustomTableMethod trait to provide a new method of
   executing statements. The given method is passed in via classname in the
   SparkConf.

08d95e9 (Russell Spitzer, 2 hours ago)
   SPARKC-459: Allow for a different Session Interface in SessionProxy

   Previously the interface of Session proxy was always the default OSS
   Datastax Driver session. A new connection parameter is added
   spark.cassandra.connection.session_interface which lets an end user specify
   a different session interface to be used.

e905ccb (Russell Spitzer, 3 hours ago)
   SPARKC-459: Seperate Execute Method out of FetchTokenRange Code

   The initial refactor moves the logic for actually executing statements out
   of the fetchTokenRange method. This is now done via a method which is
   determined at runtime. At the moment this just falls back to the original
   method.

acca81f (Russell Spitzer, 4 hours ago)
   SPARKC-459: Deprecate and Rename ThroughputJoinCassandra

   To give the parameter a more generic name and purpose the parameter has
   been changed to reads_per_sec and placed under input. The old parameter can
   still be used but the user will be warned and any values larger than
   Int.MaxValue will be rounded to Int.MaxValue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,18 @@
 language: scala
 jdk:
   - oraclejdk8
+
 sudo: required
+
 dist: trusty
 scala:
   - 2.10.6
   - 2.11.8
 
-cache:
-  directories:
-    - $HOME/.ivy2/cache
-    - $HOME/.sbt/boot
-    - $HOME/.coursier-cache
-
-before_cache:
-  # Tricks to avoid unnecessary cache updates
-  - find $HOME/.ivy2 -name "ivydata-*.properties" -delete
-  - find $HOME/.sbt -name "*.lock" -delete
+env:
+ - CASSANDRA_VERSION=2.1.15
+ - CASSANDRA_VERSION=3.6
 
 script:
-  - "sbt ++$TRAVIS_SCALA_VERSION -Dtravis=true test"
-  - "sbt ++$TRAVIS_SCALA_VERSION -Dtravis=true it:test"
-  - "sbt ++$TRAVIS_SCALA_VERSION -Dtest.cassandra.version=2.1.15 -Dtravis=true test"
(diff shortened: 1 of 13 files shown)


RussellSpitzer:
Test failure ... only on jenkins ... :/ :/ :/

RussellSpitzer:
Note Discussion with @artem-aliev :

We are going to move everything into the ConnectionFactory to eliminate the additional class and to make it more clear that the extension points are related to the connecting driver. Additionally we'll see if we can drop the optional interface for the session parameter and instead do session.getClass

artem-aliev:
@RussellSpitzer Did you try to move all new methods into ConnectionFactory?

RussellSpitzer:
I wish I knew :/

RussellSpitzer:
Two confs in the cache

| Conf1 |  Conf 2| 
|--------|---------|
| (Set(localhost/127.0.0.1) | (Set(localhost/127.0.0.1) |
| 10000 | 10000 |
| NoAuthConf | NoAuthConf |
| None | None |
| 5000 | 5000 |
| 1000 | 1000 |
| 60000 | 60000 |
|            |   |
| 10       | 10 |
| 30000 | 30000 |
| 120000 | 120000 |
| DummyFactory$@7ea4f3c0| DefaultConnectionFactory$@647d8066 |
| CassandraSSLConf | CassandraSSLConf |


So it looks like the DummyFactor is getting timed out in all the other tests except this env. So this is a testing only issue and i'll clean it up by clearing out the cache before those tests run.

PR #1079, opened 2017-02-03
Spark 2.0 is incompatible with Scala 2.10.x, so does the connector

yiu31802:
The details is described in the commit message, thus I here share some errors which I encountered with a build binary compiled with Scala-2.10.
```
     val df = sqlContext
      .read
      .options(Map("table" -> "mytab", "keyspace" -> "mykey"))
      .load()`
     println(df.show())
```
This returns `java.lang.ClassNotFoundException` of `scala.collection.GenTraversableOnce`. The issue disappeared when I re-built the connector with `sbt -Dscala-2.11=true assembly`.


RussellSpitzer:
Spark 2.0 still is compatible with Scala 2.10 it's just no longer the default distribution. See
https://mvnrepository.com/artifact/org.apache.spark/spark-core_2.10/2.0.2

RussellSpitzer:
We also provide both 2.10 and 2.11 artifacts see
https://mvnrepository.com/artifact/com.datastax.spark/spark-cassandra-connector_2.10
and
https://mvnrepository.com/artifact/com.datastax.spark/spark-cassandra-connector_2.11

yiu31802:
My fault. Sorry about this.

RussellSpitzer:
It's not a problem :) We'll get rid of it at Spark 2.2 if they actually decide to go through with removing it then. Scala 2.10 is officially deprecated in Spark 2.1 but we'll probably keep making 2.10 artifacts for that.

PR #1080, opened 2017-02-07
SPARKC-412: Optimize Spark SQL count(*) performance by calling C* cou…

artem-aliev:
…nt()


alexliu68:
Two questions: 
1) what to handle if  optimizer returns empty column for query other than count query?
2) Is it possible to create a virtual column called count and returns the C* count result directly?

artem-aliev:
1. Optimiser pass empty column set by any reason:
I do not see problem here. The connector should return row RDD with no columns. That is exactly what it did. count is called internally
2. passing count directly to Spark
I did not find how to make it.

PR #1085, opened 2017-02-16
SPARKC-472 Add confirm.truncate to source table option

alexliu68:


RussellSpitzer:
@alexliu68 Reminder on adding a test to this pr :)

PR #1086, opened 2017-02-16
SPARKC-472-b1.6 Always use the codec cache when reading from Driver Rows

RussellSpitzer:
SPARKC-383 added a cache of codecs so reads would not look up codecs for
every row. This patch extends that so that all RowReaders will use the
cache instead of just some.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/ValueRowReader.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/ValueRowReader.scala
@@ -14,7 +14,10 @@ class ValueRowReader[T: TypeConverter](columnRef: ColumnRef) extends RowReader[T
     * @param row row fetched from Cassandra
     * @param rowMetaData: column names available in the `row` */
   override def read(row: Row, rowMetaData: CassandraRowMetadata): T =
-    converter.convert(GettableData.get(row, columnRef.cqlValueName))
+    converter.convert(GettableData.get(
+      row,
+      columnRef.cqlValueName,
+      rowMetaData.codecs(columnRef.cqlValueName)))
 
   /** List of columns this `RowReader` is going to read.
     * Useful to avoid fetching the columns that are not needed. */
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,9 @@
+1.6.5
+* Optimize Spark Sql Count(*) with CassandraCount (SPARKC-412)
+* Fix Incomplete Shading of Guava (SPARKC-461)
+* Replace all Logging Dependencies from Spark-Core with Internal Class (SPARKC-460)
+* Remove Java8 Dependency in CassandraConnectorConf (SPARKC-462)
+
 1.6.4
 * Find converters for classes on different classloaders (SPARKC-363)
 * Fix CassandraConnector and Session Race Condition (SPARKC-441)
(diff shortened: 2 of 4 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/ValueRowReader.scala
```diff
@@ -14,7 +14,10 @@ class ValueRowReader[T: TypeConverter](columnRef: ColumnRef) extends RowReader[T
     * @param row row fetched from Cassandra
     * @param rowMetaData: column names available in the `row` */
   override def read(row: Row, rowMetaData: CassandraRowMetadata): T =
-    converter.convert(GettableData.get(row, columnRef.cqlValueName))
+    converter.convert(GettableData.get(
+      row,
+      columnRef.cqlValueName,
+      rowMetaData.codecs(columnRef.cqlValueName)))
```
artem-aliev: We can do it even faster.
Now you have to hash table look up per row.
Looks here 
https://github.com/datastax/spark-cassandra-connector/blob/86f0de138f0ce3a11aedd2c400898398a38bcce0/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/ValueRowReader.scala#L31-L31
Value reader is always read the first column.
and documentation support this statement:
https://github.com/datastax/spark-cassandra-connector/blob/44de124012698ab6bb985e5c45af5eaba4146cac/spark-cassandra-connector/src/main/java/com/datastax/spark/connector/japi/CassandraJavaUtil.java#L346-L346

So you can simple change it to:
converter.convert(GettableData.get(row, 0), rowMetaData.codecs(0))

RussellSpitzer: So this breaks for a few things, The following tests
[info] - should allow to transform rows into KV pairs of two single-column types *** FAILED *** (145 milliseconds)
[info] - should allow to use of keyByAndApplyPartitioner *** FAILED *** (419 milliseconds)

Because the ValueReader is also used by KeyValueReader
https://github.com/datastax/spark-cassandra-connector/blob/44de124012698ab6bb985e5c45af5eaba4146cac/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/KeyValueRowReader.scala#L13-L15

RussellSpitzer: I think what we really want to do may be to set all the row getting machinery once in all RowReaders. So each RowReader gets to have it's own cached list of what columns will be read and in what order ... I do think that's a bit out the scope of this ticket which is just to reduce the massive amount of garbage being made by the full codec lookup

RussellSpitzer: I feel like having an interface for all the rowreader to have an initial "Setup" may be the key here so we don't have to do something like

```
  private var codec: Option[TypeCodec[AnyRef]] = None
  private var columnIndex: Option[Int] = None

  /** Reads column values from low-level `Row` and turns them into higher level representation.
    * @param row row fetched from Cassandra
    * @param rowMetaData: column names available in the `row` */
  override def read(row: Row, rowMetaData: CassandraRowMetadata): T = {
    if (codec.isEmpty) {
      codec = Some(rowMetaData.codecs(rowMetaData.namesToIndex(columnRef.cqlValueName)))
      columnIndex = Some(rowMetaData.namesToIndex(columnRef.cqlValueName))
    }
    converter.convert(GettableData.get(
      row,
      columnIndex.get,
      codec.get))
  }```

PR #1089, opened 2017-02-28
SPARKC-476: SessionProxy Should proxy all Runtime Interfaces

RussellSpitzer:
Previously Session proxy would only proxy the Session interface. Here we
look up all available interfaces and proxy all of them.


RussellSpitzer:
retest this please

PR #1093, opened 2017-02-28
SPARKC-474: Upgrade Java Driver Dependency For 2.0 Release

RussellSpitzer:


RussellSpitzer:
interesting, Travis doesn't seem to mind on 2.11. Investigating

RussellSpitzer:
@jtgrabowski so this is part of the 3.1.X line of the Java driver. It no longer blocks when unable to add an async request to the request queue. Instead it just returns a future immediately with a noHostAvailable exception when it resolves. Since this only effects one of our tests which attempts to simultaneously add 10K+ requests to a queue which by default is only 1k elements deep i've just expanded the queue for that test.

PR #1101, opened 2017-03-21
SPARKC-480 Support load(path) for datasource

alexliu68:


RussellSpitzer:
My big question on this would be what do we gain from having another way of choosing the Cassandra Table? We already have one way so why add a second parallel method?

alexliu68:
It's from one of DSE ticket to make it more easy to load the table similar way as load a file.

RussellSpitzer:
My concern is that we would now have 3 ways to load a dataframe

```scala
spark.read.format("org.apache.spark.sql.cassandra").options(Map("table"-> "tab", "keyspace" -> "ks")
spark.read.cassandraFormat("tab", "ks")
spark.read.format("org.apache.spark.sql.cassandra").load("tab.ks")

```

alexliu68:
load method is a public API from Spark, the pull is an implementation of the public API. As OSS Spark, there are multiple ways to load data sources. 

`spark.read.cassandraFormat("tab", "ks")` is more a custom api from SCC.
It's ok to have all different ways to load a table.

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataFrameSpec.scala
```diff
@@ -81,13 +81,7 @@ class CassandraDataFrameSpec extends SparkCassandraITFlatSpecBase with Eventuall
...
       .read
       .format("org.apache.spark.sql.cassandra")
-      .options(
-        Map(
-          "table" -> "kv",
-          "keyspace" -> ks
-        )
-      )
-      .load()
+      .load(s"$ks.kv")
```
RussellSpitzer: I think we should have a test where we override the "options" , maybe there should be a message if you do that?

alexliu68: I added one for it.

PR #1102, opened 2017-03-21
Requesting a pull to datastax:b2.0 from datastax:SPARKC-479-2.0  83ab41e (Russell Spitzer, 7 days ago)     SPARCK-479: No parameters with Capital Letters

RussellSpitzer:
The Datasource API automatically lower cases all incoming configuration
    options. This means any configuration options with Capitals in them cannot
    actually be pased to a dataframe. To fix this we will deprecate all
    parameters with Capitals and provide automatic translations of them from
    their old usage to the new one.

    This is done by each "fromSparkConf" method of ours autoamtically iterating
    over all DeprecatedConfigParameter objects declared by that configuration
    and replacing them if they exist. If we deprecate a parameter and there is
    no replacement it will throw an error.

 3bfa0ea (Russell Spitzer, 3 weeks ago)
    SPARKC-476: SessionProxy Should proxy all Runtime Interfaces

    Previously Session proxy would only proxy the Session interface. Here we
    look up all available interfaces and proxy all of them.


RussellSpitzer:
@jacek-lewandowski one additional comment for stepping up to 2.0

PR #1108, opened 2017-03-28
SPARKC-481: Refactor Custom Scan Method

RussellSpitzer:
The original implmentation of the CustomScanMethod made it difficult to
extend the functionality inside of CassandraTableScan RDD since it
provided no ability to perform a "close" action on termination. This
meant external users were unable to properly cleanup their
implementations.

To fix this we change getScanMethod to getScanner which returns a
Scanner object which provides two methods, Scan and Close. One Scanner
is built per Spark Partition so the the initializer will only be called
once per Spark Partition and the close method only once at task
Completion.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Scanner.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Scanner.scala
@@ -1,0 +1,41 @@
+package com.datastax.spark.connector.cql
+
+import com.datastax.driver.core.{Row, Session, Statement}
+import com.datastax.spark.connector.CassandraRowMetadata
+import com.datastax.spark.connector.rdd.ReadConf
+import com.datastax.spark.connector.rdd.reader.PrefetchingResultSetIterator
+
+/**
+  * Object which will be used in Table Scanning Operations.
+  * One Scanner will be created per Spark Partition, it will be
+  * created at the beginning of the compute method and Closed at the
+  * end of the compute method.
+  */
+trait Scanner {
+  def close(): Unit
+  def getSession(): Session
+  def scan(statement: Statement): ScanResult
+}
+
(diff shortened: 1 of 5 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/Scanner.scala
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+  * Object which will be used in Table Scanning Operations.
+  * One Scanner will be created per Spark Partition, it will be
+  * created at the beginning of the compute method and Closed at the
+  * end of the compute method.
+  */
+trait Scanner {
+  def close(): Unit
+  def getSession(): com.datastax.driver.core.Session
```
jacek-lewandowski: do you really need full name here?

RussellSpitzer: nope fixed

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableScanRDD.scala
```diff
@@ -325,10 +325,11 @@ class CassandraTableScanRDD[R] private[connector](
...
 
   private def fetchTokenRange(
-    session: Session,
+    scanner: Scanner,
     range: CqlTokenRange[_, _],
-    inputMetricsUpdater: InputMetricsUpdater,
-    scanExecuteMethod: Statement => (Iterator[Row], CassandraRowMetadata)): Iterator[R] = {
+    inputMetricsUpdater: InputMetricsUpdater): Iterator[R] = {
+
+    val session = scanner.getSession()
```
jtgrabowski: it looks like `createStatement` could be internal for Scanner. This way we wouldn't have to expose session. Session would be a private thing

PR #1110, opened 2017-05-15
SPARKC-491: add java.time classes support to converters

artem-aliev:


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
```diff
@@ -489,6 +489,77 @@ object TypeConverter {
...
+      case x: String => java.time.LocalDate.parse(x)
+      case x: Int => java.time.LocalDate.ofEpochDay(x)
+      case x: Long => java.time.LocalDate.ofEpochDay(x)
+      case x: JodaLocalDate => java.time.LocalDate.of(x.getYear, x.getMonthOfYear, x.getDayOfMonth)
+      case x: DateTime => {
+        val ld = x.toLocalDate
+        java.time.LocalDate.of(x.getYear, x.getMonthOfYear, x.getDayOfMonth)
+      }
+      case x: Date => {
+        val a = JodaLocalDate.fromDateFields(x)
```
RussellSpitzer: Why intermediate variables here? I don't think it makes it much more clear than if you just did 
`convert(JodaLocalDate.fromDateFields(x)`

PR #1111, opened 2017-05-16
SPARKC 492: Protect against Size Estimate Overflows

RussellSpitzer:
```scala
val tokenRangeSizeInBytes = (totalDataSizeInBytes /
ringFraction).toLong
```

In datasize estimates has the possibility of overflowing when
getting the size_estimates from a datacenter whose primary range
contains a very tiny range BUT still also contains data. In these
cases a tiny size in bytes will be divided by a miniscule range
fraction leading to a huge value. This when converted toLong becomes
Long.MaxValue.

When these situations arise our size_estimates are no longer
reliable so we need to fall back to a safer estimate. In 2.0 we can
use the min Spark Nodes * 2 + 1 but in 1.6 we don't have that
information. To keep our changes small, here we will just use the
endpoints present in the target datacenter * 2 + 1.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraPartitionGenerator.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraPartitionGenerator.scala
@@ -30,6 +30,9 @@ private[connector] class CassandraPartitionGenerator[V, T <: Token[V]](
   private val keyspaceName = tableDef.keyspaceName
   private val tableName = tableDef.tableName
 
+  if (splitSize == 0)
+    throw new IllegalArgumentException(s"Split Size Cannot be 0, Currently set to $splitSize")
+
   private val totalDataSize: Long = {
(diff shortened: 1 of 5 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/partitioner/CassandraPartitionGenerator.scala
```diff
@@ -30,14 +30,34 @@ private[connector] class CassandraPartitionGenerator[V, T <: Token[V]](
   private val keyspaceName = tableDef.keyspaceName
   private val tableName = tableDef.tableName
 
-  private val totalDataSize: Long = {
+  if (splitSize == 0)
+    throw new IllegalArgumentException(s"Split Size Cannot be 0, Currently set to $splitSize")
+
+  val totalDataSize: Long = {
```
jtgrabowski: why public?

RussellSpitzer: That was a mistake from when I was testing things, setting it back to private

PR #1118, opened 2017-05-31
DSP-12259: add support for date string parsing

artem-aliev:
timestamp has one, to have a consistent story the date parser is added


RussellSpitzer:
LGTM 👍

PR #1122, opened 2017-06-11
SPARKC-495 Add missing type conversions for java.time.LocalDate

jrglee:


RussellSpitzer:
Thanks so much for your commit, I"ll be back-porting it to 2.0

PR #1124, opened 2017-06-14
SPARKC-494: MultipleRetryPolicy should retry with null

RussellSpitzer:
Null means using the statement's original query CL. This is important
because with LWT the CL returned by C* will not match the original
serial CL if the query is in it's non-serial portion. This forces the
original CL to be used regardless of server actions.


jtgrabowski:
I'm not sure about this. [The change](https://github.com/datastax/java-driver/pull/588/commits/e3307d950b0aa43f55aef260b2ef8c8cffd5499c#diff-8e73687a7d5321915aec32d7886f05b7R385) seams to be introduced in C* [2.1.10](https://datastax-oss.atlassian.net/browse/JAVA-764) but we support as old ones as [2.1.5](https://github.com/datastax/spark-cassandra-connector#version-compatibility). WDYT?

RussellSpitzer:
I don't think we make any promises about the driver versions? After all it's shaded by default here as well

jtgrabowski:
ok than :) :+1:

PR #1137, opened 2017-08-14
SPARKC-503: Retry PoolBusy Exceptions, Throttle JWCT Calls

RussellSpitzer:
A change in the Cassandra Java driver has caused the executeAsync method
to become fully asynchronous. This means that executeAsync calls can
sometimes return an exception of BusyPoolException, meaning that queries
were issued and no open execution pathway could be found. These
exceptions are always retryable because they indicate the local driver
state is temporarly busy. So we will retry these exceptions.

To make sure we don't have too many retries we will also make sure to
limit the number of max concurrent requests to the current
"requestsPerConnection" as specified in the pooling options. We are
already setting the maxPool size to the number of cores per executor so
this should prohibit having more requests in flight than we have open
connections.

### DIFF
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/writer/TableWriterSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/writer/TableWriterSpec.scala
@@ -200,7 +200,7 @@ class TableWriterSpec extends SparkCassandraITFlatSpecBase {
         .saveToCassandra(ks, "key_value")
     }
 
-    ioException.getMessage should include ("Invalid null value in condition for column group")
+    ioException.getMessage should include ("Invalid null value")
   }
 
   it should "write to a table with only partition key and static columns without clustering" in {
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/writer/TableWriterSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/writer/TableWriterSpec.scala
@@ -6,12 +6,12 @@ import scala.collection.JavaConversions._
 import scala.concurrent.Future
 import com.datastax.driver.core.ProtocolVersion
 import com.datastax.driver.core.ProtocolVersion._
-
 import com.datastax.spark.connector.{SomeColumns, _}
 import com.datastax.spark.connector.cql._
 import com.datastax.spark.connector.embedded.YamlTransformations
 import com.datastax.spark.connector.mapper.DefaultColumnMapper
 import com.datastax.spark.connector.types._
+import org.apache.spark.SparkException
 
 case class Address(street: String, city: String, zip: Int)
 case class KV(key: Int, value: String)
@@ -192,6 +192,15 @@ class TableWriterSpec extends SparkCassandraITFlatSpecBase {
     )
     sc.parallelize(col).saveToCassandra(ks, "key_value")
     verifyKeyValueTable("key_value")
+  }
+
+  it should "pass the exception back to the driver on a write failure" in {
+    val ioException = intercept[SparkException] {
+      sc.parallelize[(Int, Option[Long], String)](Seq((1, None, "Hello")))
+        .saveToCassandra(ks, "key_value")
+    }
+
+    ioException.getMessage should include ("Invalid null value in condition for column group")
   }
 
   it should "write to a table with only partition key and static columns without clustering" in {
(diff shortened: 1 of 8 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/AsyncExecutor.scala
```diff
@@ -28,28 +30,43 @@ class AsyncExecutor[T, R](asyncAction: T => ListenableFuture[R], maxConcurrentTa
...
+            case nHAE: NoHostAvailableException if nHAE.getErrors.asScala.values.exists(_.isInstanceOf[BusyPoolException]) =>
+              logWarning("BusyPoolException ... Retrying")
+              tryFuture()
+
+            case otherException =>
+              logError("Failed to execute: " + task, throwable)
+              if (_successful) _successful = false
+              release()
+              settable.setException(throwable)
+              latestExeception = Some(throwable)
```
jtgrabowski: Since `_successful` volatile write is conditional, we could make _successful not volatile and make latestException volatile. This may make Travis happy :)

jtgrabowski: Also it looks like it should be placed in critical section.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/TableWriter.scala
```diff
@@ -204,7 +204,13 @@ class TableWriter[T] private (
       queryExecutor.waitForCurrentlyExecutingTasks()
 
       if (!queryExecutor.successful)
-        throw new IOException(s"Failed to write statements to $keyspaceName.$tableName.")
+        throw new IOException(
+          s"""Failed to write statements to $keyspaceName.$tableName. The
+             |latest exception was
+             |  ${queryExecutor.latestExeception.map( _.getMessage).getOrElse("Unable to determine")}
```
jtgrabowski: It looks like `_successful` (and `successful`) is redundant. The if could be changed to `latestException.map(...)`

jtgrabowski:
@RussellSpitzer could you also elaborate a bit about 

> To make sure we don't have too many retries we will also make sure to
limit the number of max concurrent requests to the current
"requestsPerConnection" as specified in the pooling options

If I understand correctly, max concurrent requests while saving to C* is limited by `spark.cassandra.output.concurrent.writes` and in case of joining, it is bound to whatever is found in `poolingOptions.getMaxRequestsPerConnection`. This is not configurable (unless custom ConnectionFactory is provided). Is this intended?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/AsyncExecutor.scala
RussellSpitzer: i'm not sure what you mean in "the critical section"

RussellSpitzer:
@jtgrabowski MaxConcurrentRequests is set only for the apply method of the query executor. The Write method is still throttled as before based on "max_concurrent_writes" I didn't want to add a new tuning parameter in a point release and the driver team told me this should be safe :)

RussellSpitzer:
retest this please

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/writer/TableWriterSpec.scala
```diff
@@ -194,6 +194,15 @@ class TableWriterSpec extends SparkCassandraITFlatSpecBase {
     verifyKeyValueTable("key_value")
   }
 
+  it should "pass the exception back to the driver on a write failure" in {
```
RussellSpitzer: fixed

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/AsyncExecutor.scala
```diff
@@ -6,18 +6,22 @@ import com.datastax.spark.connector.util.Logging
...
 class AsyncExecutor[T, R](asyncAction: T => ListenableFuture[R], maxConcurrentTasks: Int,
     successHandler: Option[Handler[T]] = None, failureHandler: Option[Handler[T]]) extends Logging {
 
-  @volatile private var _successful = true
-
   private val semaphore = new Semaphore(maxConcurrentTasks)
   private val pendingFutures = new TrieMap[ListenableFuture[R], Boolean]
+  @volatile private var latestException: Option[Throwable] = None
+
+  def getLatestException(): Option[Throwable] = synchronized {
```
jtgrabowski: why do we synchronize here?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/AsyncExecutor.scala
jtgrabowski: I'm meant putting it before `release()`, it's all good now

jtgrabowski:
LGTM but please take a look at `synchronized` and see if we need it

PR #1139, opened 2017-08-23
Support partially specified writes from case classes

aashishs101:
Currently, we cannot use `rdd.saveToCassandra` on an RDD[CaseClass] if `CaseClass` does not contain all of the fields from the target table.  This will currently result in the Error: 

```java.lang.IllegalArgumentException: requirement failed: Columns not found in CaseClass: [missing_field1, missing_field2]```

This can be limiting if the table has had columns added to it from an external process.  If this is the case, a job that writes to the table will automatically fail.  This PR proposes a resolution for this issue by making the failure case more permissive (if all of the first _n_ fields of the target table are matched by the case class, then we will attempt the write).

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/mapper/DefaultColumnMapper.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/mapper/DefaultColumnMapper.scala
@@ -102,10 +102,10 @@ class DefaultColumnMapper[T : TypeTag](columnNameOverride: Map[String, String] =
       } yield (getterName, columnRef)
     }.toMap
 
-    // Check if we have all the required columns:
+    // Check if all columns at start of table description are present in the case class:
     val mappedColumns = getterMap.values.toSet
     val unmappedColumns = selectedColumns.filterNot(mappedColumns)
-    require(unmappedColumns.isEmpty, s"Columns not found in $tpe: [${unmappedColumns.mkString(", ")}]")
+    require(selectedColumns.endsWith(unmappedColumns), s"Unmapped columns nust be at end of table definition: [${unmappedColumns.mkString(", ")}]")
 
     SimpleColumnMapForWriting(getterMap)
   }
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/DefaultRowWriter.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/DefaultRowWriter.scala
@@ -19,13 +19,12 @@ class DefaultRowWriter[T : TypeTag : ColumnMapper](
 
   override def readColumnValues(data: T, buffer: Array[Any]) = {
     val row = converter.convert(data)
-    for (i <- columnNames.indices)
+    for (i <- row.columnValues.indices)
       buffer(i) = row.getRaw(i)
   }
 }
 
 object DefaultRowWriter {
-
   def factory[T : ColumnMapper : TypeTag] = new RowWriterFactory[T] {
     override def rowWriter(tableDef: TableDef, selectedColumns: IndexedSeq[ColumnRef]) = {
       new DefaultRowWriter[T](tableDef, selectedColumns)
--- a/spark-cassandra-connector/src/test/scala/com/datastax/spark/connector/writer/MappedToGettableDataConverterSpec.scala
+++ b/spark-cassandra-connector/src/test/scala/com/datastax/spark/connector/writer/MappedToGettableDataConverterSpec.scala
@@ -17,6 +17,7 @@ class MappedToGettableDataConverterSpec extends FlatSpec with Matchers {
   val addressType = UserDefinedType("address", IndexedSeq(streetColumn, numberColumn))
 
   val loginColumn = ColumnDef("login", PartitionKeyColumn, VarCharType)
+  val logoutColumn = ColumnDef("logout", RegularColumn, VarCharType)
   val passwordColumn = ColumnDef("password", RegularColumn, VarCharType)
   val addressColumn = ColumnDef("address", RegularColumn, addressType)
   val addressesColumn = ColumnDef("addresses", RegularColumn, ListType(addressType))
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: spark-cassandra-connector/src/test/scala/com/datastax/spark/connector/writer/MappedToGettableDataConverterSpec.scala [hunk 1]
```diff
@@ -278,18 +279,17 @@ class MappedToGettableDataConverterSpec extends FlatSpec with Matchers {
     row.getTupleValue(1).getUDTValue(2).getInt("number") shouldEqual 3
   }
 
-  class UnknownType
-  case class UserWithUnknownType(login: String, address: UnknownType)
+  case class UserWithUnknownType(login: String, address: String)
 
   it should "throw a meaningful exception when a column has an incorrect type" in {
     val userTable = newTable(loginColumn, addressColumn)
-    val user = UserWithUnknownType("foo", new UnknownType)
```
RussellSpitzer: Why was this test changed?

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/mapper/DefaultColumnMapper.scala
```diff
@@ -102,10 +102,10 @@ class DefaultColumnMapper[T : TypeTag](columnNameOverride: Map[String, String] =
       } yield (getterName, columnRef)
     }.toMap
 
-    // Check if we have all the required columns:
+    // Check if all columns at start of table description are present in the case class:
     val mappedColumns = getterMap.values.toSet
     val unmappedColumns = selectedColumns.filterNot(mappedColumns)
-    require(unmappedColumns.isEmpty, s"Columns not found in $tpe: [${unmappedColumns.mkString(", ")}]")
+    require(selectedColumns.endsWith(unmappedColumns), s"Unmapped columns nust be at end of table definition: [${unmappedColumns.mkString(", ")}]")
```
RussellSpitzer: Not sure what this means, seems like it would only match if unmappedColumns was in the same order as selectedColumns

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/DefaultRowWriter.scala
```diff
@@ -19,13 +19,12 @@ class DefaultRowWriter[T : TypeTag : ColumnMapper](
 
   override def readColumnValues(data: T, buffer: Array[Any]) = {
     val row = converter.convert(data)
-    for (i <- columnNames.indices)
+    for (i <- row.columnValues.indices)
```
RussellSpitzer: This returns values in the order of the data and not in the order of the selected columns?

aashishs101: It shouldn't be changing the order, since these are the indices of the Seq and all of the actual data access is being done on the next line where members of the row and buffer are being accessed/modified.  In most cases, the number of table columns will be equal to the number of values in the row.  However, in the case that we're interested in (the underspecified case class case), the number of columns in the table will be greater than the number of values in the row, and so we want to use the smaller number.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/mapper/DefaultColumnMapper.scala
```diff
@@ -102,11 +102,16 @@ class DefaultColumnMapper[T : TypeTag](columnNameOverride: Map[String, String] =
...
+      * the table columns.  Otherwise, if this is a nested type, then require that there are be no unmapped columns.
+      */
     val mappedColumns = getterMap.values.toSet
     val unmappedColumns = selectedColumns.filterNot(mappedColumns)
-    require(unmappedColumns.isEmpty, s"Columns not found in $tpe: [${unmappedColumns.mkString(", ")}]")
 
+    if (!isTopLevel)
+      require(unmappedColumns.isEmpty, s"Columns not found in nested $tpe: [${unmappedColumns.mkString(", ")}]")
+    else
+      require( selectedColumns.endsWith(unmappedColumns), s"Unmapped columns nust be at end of table definition: [${unmappedColumns.mkString(", ")}]")
```
aashishs101: @RussellSpitzer the comment you had here was removed after my most recent push, but I still wanted to address it.  Based on the way that `unmappedColumns` is created (as a filter on `selectedColumns`on line 109), the two sequences should be in the same order.  So, this requirement is essentially asserting that we should map the first _n_ columns from the table with the case class, but subsequent columns can go unmapped.

aashishs101:
@RussellSpitzer, any word on this?

RussellSpitzer:
Do we have a SPARKC ticket for this yet? I couldn't find one in jira?
https://datastax-oss.atlassian.net/issues/

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/DefaultRowWriter.scala
RussellSpitzer: makes sense

aashishs101:
Sorry, I hadn't created one until now.  Here it is: https://datastax-oss.atlassian.net/browse/JAVA-1676

RussellSpitzer:
Going to try to trigger jenkins from here

RussellSpitzer:
test this please

RussellSpitzer:
Tests running!

RussellSpitzer:
So are we sure we want to aim this patch at Master? I don't think we are going to do another big public release. So if we want to have this capability it would be nice if we could find a way of back-porting it to b1.6 or b2.0 I know we slightly change one method signature but since it's an internal api it should be ok even on the earlier branches... thoughts?

aashishs101:
@RussellSpitzer, since it is an internal method with a default parameter, I'm guessing it shouldn't change much?  I'm also not sure about all of the ramifications of doing a backport vs. a normal release.  Would I need to make a new PR that reapplies these changes against b2.0?

aashishs101:
Hey @RussellSpitzer, any update on this?

RussellSpitzer:
Sorry yeah, you need to change the PR target to b2.0

aashishs101:
test this please

PR #1140, opened 2017-09-05
LocalNodeFirstLoadBalancingPolicy not only one DC for connections

zulk666:
Maybe this approach ?


premuditha:
Why isn't this PR merged yet?

PR #1143, opened 2017-09-29
SPARKC-507: Infinite Retries

RussellSpitzer:
Changes our retry policy to retry all exceptions which can possible be
temporary. This will move burden to users to detect when the system is
beyond hope but will allow a momentarily troubled cluster recover from
issues.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectorConf.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectorConf.scala
@@ -186,8 +186,10 @@ object CassandraConnectorConf extends Logging {
   val QueryRetryParam = ConfigParameter[Int](
     name = "spark.cassandra.query.retry.count",
     section = ReferenceSection,
-    default = 10,
-    description = """Number of times to retry a timed-out query""")
+    default = 60,
+    description =
+      """Number of times to retry a timed-out query,
+        |Setting this to -1 means unlimited retries""".stripMargin)
 
   @deprecated("delayed retrying has been disabled; see SPARKC-360", "1.2.6, 1.3.2, 1.4.3, 1.5.1")
   val QueryRetryDelayParam = ConfigParameter[RetryDelayConf](
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/MultipleRetryPolicy.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/MultipleRetryPolicy.scala
@@ -1,28 +1,30 @@
 package com.datastax.spark.connector.cql
 
-import com.datastax.driver.core.exceptions.DriverException
+import com.datastax.driver.core.exceptions._
 import com.datastax.driver.core.policies.RetryPolicy
 import com.datastax.driver.core.policies.RetryPolicy.RetryDecision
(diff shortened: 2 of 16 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectorConf.scala
```diff
@@ -186,8 +186,10 @@ object CassandraConnectorConf extends Logging {
   val QueryRetryParam = ConfigParameter[Int](
     name = "spark.cassandra.query.retry.count",
     section = ReferenceSection,
-    default = 10,
-    description = """Number of times to retry a timed-out query""")
+    default = -1,
```
tolbertam: I think if you get the point that a request is retried 10 times, it is probably not going to succeed or the system is in bad state, so I think it is more reasonable as a default than infinite.   That said I do like that it can be configurable in some way ("Just try until it works").

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/cql/MultipleRetryPolicy.scala
```diff
@@ -46,7 +52,17 @@ class MultipleRetryPolicy(maxRetryCount: Int)
       ex: DriverException,
       nbRetry: Int): RetryDecision = {
 
-    RetryDecision.rethrow()
+    ex match {
+      case _: NoHostAvailableException |
```
tolbertam: Other than `OperationTimedoutException`, `OverloadedException`, `ReadFailureException`, and `WriteFailureException` I don't believe the other exceptions are evaluated in `RetryPolicy` since they are automatically tied on next host (except for `NoHostAvailableException` which just means there are no more hosts to try)

Another issue is that without knowing why a request failed (`ReadFailure`/`WriteFailure`), it's hard to know whether or not to retry.  My experience is that if you get a failure is that it's usually just going to fail again (i.e. tombstone threshold on reads, batch size failures on writes), but hard to know for sure without looking at the reasons.

RussellSpitzer:
@tolbertam Can you check this out again?
@jtgrabowski Calling on you for the SCC POV review :)

RussellSpitzer:
I just wanted more chances, is there any way we can have delays between retries? I think we talked about have some thread issues there? @tolbertam

tolbertam:
> is there any way we can have delays between retries?

Unfortunately there isn't really a way.  You could sleep in the `on*` methods, but that would block worker threads.

RussellSpitzer:
How long do you think we get with 60 retries? Think worst case scenario. We could always have our Unavailable exceptions try for a much larger amount?

RussellSpitzer:
Is there a way to tell when a request was first made? we could always retry for a certain amount of time?

tolbertam:
> How long do you think we get with 60 retries? Think worst case scenario. We could always have our Unavailable exceptions try for a much larger amount?

Since unavailable are raised almost immediately by coordinators, you can multiply the number of tries by the network latency to get a low threshold for how fast it will give up.  i.e. say if latency is 5ms, with a retry count of 60, you'll get 300ms of time (on the low end) before it gives up.

tolbertam:
> Is there a way to tell when a request was first made? we could always retry for a certain amount of time?

Not without wrapping the statement and adding that information unfortunately.

RussellSpitzer:
We already wrap our statements, so maybe this is possible ...

PR #1147, opened 2017-10-31
Update the WriteTime error exception for reader

moein7tl:
Error message for WriteTime was invalid, I've tried to fix it

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableRowReaderProvider.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableRowReaderProvider.scala
@@ -64,7 +64,7 @@ trait CassandraTableRowReaderProvider[R] {
               s"but column $columnName is not a regular column in table $keyspaceName.$tableName.")
         case WriteTime(columnName, _) =>
           if (!regularColumnNames.contains(columnName))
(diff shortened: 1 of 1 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableRowReaderProvider.scala
```diff
@@ -64,7 +64,7 @@ trait CassandraTableRowReaderProvider[R] {
               s"but column $columnName is not a regular column in table $keyspaceName.$tableName.")
         case WriteTime(columnName, _) =>
           if (!regularColumnNames.contains(columnName))
-            throw new IOException(s"TTL can be obtained only for regular columns, " +
+            throw new IOException(s"Write time can be obtained only for regular columns, " +
```
RussellSpitzer: Looks good but you should probably aim at the earliest branch which is 1.4? or something like that :)

moein7tl: @RussellSpitzer pardon me, I couldn't understand what you mean. What should I do?

moein7tl: @RussellSpitzer Am I do it write in #1148 PR?

PR #1154, opened 2017-11-29
SPARKC-426: Allow POJOS to be set as NULL

RussellSpitzer:
Previously a null would hit trigger a `CannotConvert` error before
getting to the "nulls allowed" check in gettable mapped data. Now we
attempt to pass through the null and catch the exception if we cannot
cast the Null into the acceptable type.


RussellSpitzer:
test this please

PR #1155, opened 2017-12-01
SPARKC-516: Add column name to the error message in case of conversion error

donderom:
The change adds new exception on top of the actual conversion exception with debug information about column name and expected type.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/SqlRowWriter.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/SqlRowWriter.scala
@@ -5,7 +5,10 @@ import java.util.UUID
 
 import com.datastax.spark.connector.ColumnRef
 import com.datastax.spark.connector.cql.TableDef
+import com.datastax.spark.connector.types.TypeConversionException
 import org.apache.spark.sql.Row
+
+import scala.util.{Failure, Success, Try}
 
 
 /** A [[RowWriter]] that can write SparkSQL `Row` objects. */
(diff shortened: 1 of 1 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/SqlRowWriter.scala
```diff
@@ -23,9 +26,16 @@ class SqlRowWriter(val table: TableDef, val selectedColumns: IndexedSeq[ColumnRe
     require(row.size == columnNames.size, s"Invalid row size: ${row.size} instead of ${columnNames.size}.")
     for (i <- 0 until row.size) {
       val colValue = row(i)
-      buffer(i) = converters(i).convert(colValue)
+      buffer(i) = Try(converters(i).convert(colValue)) match {
+        case Success(value) => value
+        case Failure(ex: IllegalArgumentException) =>
```
donderom: The most exceptions I've seen that were hard to track down were either `IllegalArgumentException` or its subclass.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/writer/SqlRowWriter.scala
```diff
@@ -23,9 +26,16 @@ class SqlRowWriter(val table: TableDef, val selectedColumns: IndexedSeq[ColumnRe
...
+          val columnType = columnTypes(i).scalaTypeName
+          throw new TypeConversionException(
+            s"Cannot convert value $colValue of column ${columnNames(i)} to type $columnType", ex)
```
donderom: The message will look like `Cannot convert value somevalue of column somecolumn to type Int` without any wrapping of `somevalue` (like single or double quotes). Any suggestions on how to improve readability?

RussellSpitzer:
We probably will want to do some perf testing around this change since this is adding a try catch into a pretty tight loop.

donderom:
Would it be ok to change it from using `Try` to good ol' `try/catch`? Or is it the very `try/catch` performance?

RussellSpitzer:
Mostly I was worried about the Try/Catch block in the very inner loop of the conversion here. Sorry I've been so long in responding, I do think that for fixing this issue better maybe we should force Spark to do the conversion IE not accepting a String for a numeric type ... or something like that so Spark does the conversion instead of us in the type converter code ... Gotta think more on this ...

PR #1156, opened 2017-12-19
Fix minor error in configuration README

ndamclean:


RussellSpitzer:
This would need to be fixed in the property itself, since the reference.md file is automatically generated from the property's class

PR #1159, opened 2018-01-23
SPARKC-522: Add Timestamp converter, improve Log converter for better…

artem-aliev:
… SparkSQL support

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
@@ -2,6 +2,7 @@ package com.datastax.spark.connector.types
 
 import java.net.InetAddress
 import java.nio.ByteBuffer
+import java.sql.Timestamp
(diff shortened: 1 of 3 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
```diff
@@ -378,6 +383,18 @@ object TypeConverter {
...
 
+  private val TimestampTypeTag = TypeTag.synchronized {
+    implicitly[TypeTag[Timestamp]]
+  }
+
+  implicit object TimestampConverter extends NullableTypeConverter[Timestamp] {
+    def targetTypeTag = TimestampTypeTag
+    override def convertPF = {
+      case x: Timestamp => x
+      case x => Timestamp.from(DateConverter.convert(x).toInstant())
```
RussellSpitzer: shouldn't this only be for Strings?

RussellSpitzer: ah nvm I get it

PR #1160, opened 2018-01-28
Bump version of sbt-coursier plugin

alexott:
Scalaz version used by previous version of that plugin wasn't compatible with sbtclipse
plugin, so bumping version of sbt-coursier plugin that uses newer version of scalaz,
resolved the issue


RussellSpitzer:
Thanks!

PR #1161, opened 2018-02-01
SPARKC-527: Allow Split Count to be Passed to CassandraSourceRelation

RussellSpitzer:
Previous to this it was impossible to choose the target split count for
a CassandraSourceRelation (SparkSql, Dataframes). This patch allows
for case insensitive parameter passing as well as adding a new parameter
for `splitCount`. The extra lowerCase in `consolidateConfs` is for
making the forward merge easier.

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -236,6 +236,14 @@ OSS Cassandra this should never be used.</td>
   <td>64</td>
   <td>Approx amount of data to be fetched into a Spark partition</td>
 </tr>
+<tr>
+  <td><code>splitCount</code></td>
+  <td>None</td>
+  <td>Specify the number of Spark partitions to
+read the Cassandra table into. This parameter is
+used in SparkSql and DataFrame Options.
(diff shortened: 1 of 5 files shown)


doc/reference.md
```diff
@@ -236,6 +236,14 @@ OSS Cassandra this should never be used.</td>
   <td>64</td>
   <td>Approx amount of data to be fetched into a Spark partition</td>
 </tr>
+<tr>
+  <td><code>splitCount</code></td>
+  <td>None</td>
+  <td>Specify the number of Spark partitions to
+read the Cassandra table into. This parameter is
+used in SparkSql and DataFrame Options.
```
jtgrabowski: This comment is a bit confusing. RDD like CassandraTableScanRDD also respect this parameter, right?

RussellSpitzer:
They would but you wouldn't be able to set it since it doesn't start with
spark.

PR #1163, opened 2018-02-14
Fixed Doc Issue

rs017991:
Removed an extra copy of splitCount that was added in 6c81a54d59154c7fd84ab749df688327dc6f759e


bcantoni:
Thanks for noticing and fixing this.

PR #1168, opened 2018-03-16
SPARKC-532: Change Str Literal Match to Be Greedy

RussellSpitzer:
Previous non greedy match would StackOverflow on long string literals.
New version does not.


RussellSpitzer:
@artem-aliev could please review?

PR #1169, opened 2018-03-20
SPARKC-533: restore support for various timezone formats to Timestamp…

artem-aliev:
…Parser

    ":" time separate time zone should be parsed
    "Z" timezone is supported
    The parser fail back to previous iterative solution in case of any other problem


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/TimestampParser.scala
```diff
@@ -32,30 +33,59 @@ object TimestampParser {
...
     try {
-      val format = if (containsTAtFixedPosition(date)) formatsWithTBySize(date.length) else formatsBySize(date.length)
+      implicit def bool2int(b: Boolean) = if (b) 1 else 0
+      // trying guess string format by length and two symbols
+      val format: DateTimeFormatter = converterMatrix(containsTAtFixedPosition(date))(endsWithZ(date))(date.length)
       format.parseDateTime(date).toDate
     } catch {
-      case _: Throwable => throw new IllegalArgumentException(s"Invalid date: [$date]")
+      // guess fail, iterate over all parsers
+      case _: Throwable => slowParse(date)
```
RussellSpitzer: If we can't find one in the conversion how will any of those match? If we couldn't tell by the hints then don't we know none of our parsers will work?

artem-aliev: During development I hit new TimeZone format regression two times. SO to make sure we do not create new regressions I put all old code into the slowParser. SO If I miss some exotic case, user will not get better performance but it will work as it work before (including performace)

artem-aliev:
retest this please

PR #1170, opened 2018-03-20
Created a local reflection lock in order to work with spark 2.3.0.

nikolovivan:
Patch to make the connector work with Spark 2.3.0.

In Spark, this commit https://github.com/apache/spark/commit/425c4ada4c24e338b45d0e9987071f05c5766fa5 removes Scala 2.10 support. One of the things removed is the workarounds using `ScalaReflectionLock`, which causes a `java.lang.NoClassDefFoundError` when using the connector.

This pull request can potentially help resolve https://datastax-oss.atlassian.net/projects/SPARKC/issues/SPARKC-530.

While removing Scala 2.10 support might be the right way to go, as Spark seems to be going in this direction, it turned out to be a much more intrusive change.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/org/apache/spark/sql/catalyst/ReflectionLock.scala
+++ b/spark-cassandra-connector/src/main/scala/org/apache/spark/sql/catalyst/ReflectionLock.scala
@@ -2,5 +2,7 @@ package org.apache.spark.sql.catalyst
 
 object ReflectionLock {
 
-  val SparkReflectionLock = ScalaReflectionLock
+  protected object ScalaReflectionLockFallback
+
+  val SparkReflectionLock = ScalaReflectionLockFallback
 }


RussellSpitzer:
Please post PR's on the corresponding Jira Ticket as we do not check Github Pull requests without corresponding Jiras.

Below is mostly copied from other PR for this workaround. 

The main issue with this PR is that the fix is a workaround for the missing class. This will break compatibility with Spark 2.0-2.2. I think the correct thing to do here is to completely remove the SparkReflectionLock from the Spark Cassandra Connector as well as the Scala 2.10 compatibility code. This would bring our testing back up to green and we wouldn't have to worry about this issue in the future.

So in my eyes a Spark 2.3.0 branching event requires

Removal of SparkReflectionLock
Removal of Scala 2.10 support
All supported integration tests run green (Hopefully this is just free after removing Scala 2.10)

nikolovivan:
Yeah, that other pull request came after mine. As you can see it has my commit in there and the guy thought he wanted to improve it. I raised my concerns about modifying dependencies unrelated to what the PR is doing when he emailed me to inform me about it.

I completely agree with how the branching has to be approached. I've even stated it in my description.

There is a branch in my fork which does exactly that, but the change seems quite intrusive and I abandoned it for the time being. I actually needed some kind of confirmation by someone from DS to make sure I don't waste my time going that route.

It's a shame that you do not check Github - doesn't look like there is much noise in here anyway. If you had, that would have already been done.

I will see if I get some time to complete the work in my fork. If not, it will have to be someone else.

nikolovivan:
Created https://github.com/datastax/spark-cassandra-connector/pull/1175.

PR #1171, opened 2018-04-03
Upgrade to spark2.3.0 and cassandra driver 3.4

schernolyas:
Hi

PR #1172, opened 2018-04-09
SPARKC-536 UDT converters optimization

glutamatt:
### [SPARKC-536](https://datastax-oss.atlassian.net/browse/SPARKC-536)

Avoid overhead of repeated `columnTypes(i).converterToCassandra` which lead to synchronized reflection costly methods.

Profiling a spark job wich prepare a sstable from dataframe with many udt into collections, with project https://github.com/jparkie/Spark2Cassandra , up to 90% of the time is spent into `converterToCassandra` for the udt columns conversion

I just expand the memoization of converters we already have in `DriverUDTValueConverter` to the case class UDTFieldDef converter generator in converterToCassandra

I wasn't able to register on the jira project (in order to create a issue to associate to the PR) 
> datastax-oss.atlassian.net has reached its license limit.

[Build Status](https://travis-ci.org/glutamatt/spark-cassandra-connector)

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/UserDefinedType.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/UserDefinedType.scala
@@ -33,6 +33,8 @@ case class UserDefinedType(name: String, columns: IndexedSeq[UDTFieldDef])
   def scalaTypeTag = SparkReflectionLock.synchronized { implicitly[TypeTag[UDTValue]] }
   def cqlTypeName = name
 
+  val fieldConverters = columnTypes.map(_.converterToCassandra)
+
   def converterToCassandra = new NullableTypeConverter[UDTValue] {
     override def targetTypeTag = UDTValue.TypeTag
     override def convertPF = {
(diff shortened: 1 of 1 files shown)


glutamatt:
CPU sample **before** the optimization

image

**After** => **72 % faster** on the writeRowsToSSTables method

image

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/UserDefinedType.scala
```diff
@@ -33,14 +33,16 @@ case class UserDefinedType(name: String, columns: IndexedSeq[UDTFieldDef])
   def scalaTypeTag = SparkReflectionLock.synchronized { implicitly[TypeTag[UDTValue]] }
   def cqlTypeName = name
 
+  val fieldConverters = columnTypes.map(_.converterToCassandra)
+
```
abenicho: remove empty line

RussellSpitzer:
I think this is a great fix, could you please make a corresponding Jira? We don't actively check github pullrequests which is why no one from DS has looked at this.

ManuelWeiss:
@RussellSpitzer That's a big mistake. Can you also please get https://github.com/datastax/spark-cassandra-connector/pull/1171 merged?

glutamatt:
> could you please make a corresponding Jira?

Done -> [SPARKC-536](https://datastax-oss.atlassian.net/browse/SPARKC-536)

RussellSpitzer:
Excellent! Thank you so much

bcantoni:
@glutamatt thanks for your note about the https://datastax-oss.atlassian.net JIRA instance being at its limit -- we've cleaned out some inactive users to make room for new accounts if you'd like to create one now.

PR #1175, opened 2018-05-11
Add support for Spark 2.3.0

nikolovivan:
After finally getting a response to https://github.com/datastax/spark-cassandra-connector/pull/1170 confirming that it's fine to lose Scala 2.10 support, here it is...

In Spark, this commit apache/spark@425c4ad removes Scala 2.10 support. One of the things removed is the workarounds using ScalaReflectionLock, which causes a java.lang.NoClassDefFoundError when using the connector.

This pull request potentially resolves https://datastax-oss.atlassian.net/projects/SPARKC/issues/SPARKC-530.


ManuelWeiss:
👍 nice work!

fahman:
+1

nikolovivan:
test this please

Fatmali:
+1

RussellSpitzer:
I'll need to put a check in travis to stop doing 2.10 checks

RussellSpitzer:
@nikolovivan There are some compilation errors in the data source scan spec, probably because it uses some Spark internals (or at least internalish) code to check query plan builds. Do you want to take a crack at fixing it or want me to rewire that? Once the compilation is fixed I think it may hit SPARKC-537 which I think is due to how 2.2+ spark changed IsNotNull predicates

RussellSpitzer:
I think with the mods made the 2.10 build is just going to do the 2.11 build ... so we'll all just remember that it's the same on this ticket, and I"ll put a version gate after we tag and branch.

nikolovivan:
I updated the code. Any chance we could try the Jenkins build again?

PS. I'm having some issues triggering the integration tests locally. I think it has to do with ports being closed on my local machine, but I could try getting a different machine if things don't work out and we end up having to go back and forth.

RussellSpitzer:
test this please

RussellSpitzer:
Local runs are probably failing because of this https://issues.apache.org/jira/browse/CASSANDRA-14173 if your java is up to date

RussellSpitzer:
```
Error Message
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 5.0 failed 1 times, most recent failure: Lost task 0.0 in stage 5.0 (TID 14, localhost, executor driver): java.lang.NoSuchMethodError: net.jpountz.lz4.LZ4BlockInputStream.<init>(Ljava/io/InputStream;Z)V  at org.apache.spark.io.LZ4CompressionCodec.compressedInputStream(CompressionCodec.scala:122)  at 
```

I'll get a integration test fix in so it's easier to pick this stuff up locally

RussellSpitzer:
https://github.com/datastax/spark-cassandra-connector/pull/1176
I'm gonna let integration tests run on this branch then merge forward to master, then that should be good to rebase on for local testing.

nikolovivan:
Looks like a dependency issue. I'll try and set up a different environment in a container where I can have a bit more control and run those integration tests.

Let me know when/if I can help. Keen to get this over the line.

nikolovivan:
I managed to run the intergration tests on a Docker container which was pretty nice. Happy to provide it if needed.

There are quite a few commits here, but not much has changed - I've updated some versions to be in line with what Spark 2.3.0 uses and I've added an exclusion in test that fixes the class not found error.

The issue now is that 6 of those tests fail for me (actually 4, but 1 pair is ran twice with different input parameters). Another pair has to do with the hints for wrong CF or KS names not coming back as expected for some reason, even though my changes shouldn't have affected this.

There is a possibility that it is due to me not setting up the container image exactly the same way as the build server, so maybe worth trying the Jenkins build again. One of the failures looks like it could be related to SPARKC-537 mentioned above.

RussellSpitzer:
I'm almost done with my own mods to bring everything up to 3.11 and 2.2, once I get that in we should be ok on local and jenkins builds.

RussellSpitzer:
@nikolovivan the name mismatch thing is because apache commons changed their underlying algorithm for JaroWinkler, the spark upgrade updates commons so now the threshold needs recalibrating. 

I'm working on other test failures as well for 2.2 so hopefully once fixed on 2.2 they'll all pass on your branch

nikolovivan:
@RussellSpitzer, thanks for the update. I was looking into the failures yesterday:
* for JaroWinkler I was printing the distances and saw that the threshold needs to go down to maybe `0.6` in order to get the tests to pass again. I'll recalibrate it by going down the least possible amount to get the tests to pass. I guess it's safe enough.
* one of the tests seems to fail because an additional configuration might be needed to get the logical plan to get propagated - `spark.sql.constraintPropagation.enabled`. Should be OK to get it to work.
* the last one might get fixed by your changes.

I will wait for your work to be done and then merge in into this branch and then bring it to green.

RussellSpitzer:
@nikolovivan have you looked over https://github.com/datastax/spark-cassandra-connector/pull/1176

Our 2.10 jenkins build seems to be broken but I did similar fixes to what you did to get the JW fixing, i actually did both lowercasing and lowering the threshold.

I haven't seen the test that is failing because of constraint propagation? But basically i have a clean run on 2.2 now and I also switched C* to 3.11 and logging to logback. I'll have the tests run on your current branch since I don't know what other fixes might be needed.

RussellSpitzer:
test this please

nikolovivan:
@RussellSpitzer Just saw your PR. I didn't have to lower the threshold for JW. Just lowercasing made my test pass. Maybe we should leave it like this and keep it `0.85`?

Hopefully your test will fix the pushdown filter.

The constraint propagation is in `CassandraDataSourceSpec` - `it should "pass through local conf properties"`. Seems like `qp.constraints should not be empty` doesn't hold for me. I looked in the Spark code and it checks if `spark.sql.constraintPropagation.enabled`. So hopefully this will be just about setting this config in test. Tried quickly, but didn't manage, so I might be missing something simple.

RussellSpitzer:
@nikolovivan that's fine with me, for some reason the naming tests still fails for me but only on jenkins. Not in travis, and not locally.

Lets replace the "local conf test" I have another version I'm working on that doesn't so strictly depend on Spark internals.

```
 it should "pass through local conf properties" in {

    sc.setLocalProperty(
      CassandraSourceRelation.AdditionalCassandraPushDownRulesParam.name,
      "com.datastax.spark.connector.sql.PushdownUsesConf")

    val df = sparkSession
      .read
      .options(Map("keyspace" -> ks, "table" -> "test1", PushdownUsesConf.testKey -> "Don't Remove"))
      .load().filter("g=1 and h=1")

    df.queryExecution.executedPlan // Will throw an exception if local key is not set

    val df2 = sparkSession
      .read
      .options(Map("keyspace" -> ks, "table" -> "test1"))
      .load().filter("g=1 and h=1")

    intercept[IllegalAccessException] {
      df2.explain() //
    }


/**
  * Throws an exception if test key is not set
  */
object PushdownUsesConf extends CassandraPredicateRules {
  val testKey = "testkey"
  val notSet = "notset"
  override def apply(
                      predicates: AnalyzedPredicates,
                      tableDef: TableDef,
                      conf: SparkConf): AnalyzedPredicates = {
    if (conf.contains(testKey)){
      predicates
    } else {
      throw new IllegalAccessException(s"Conf did not contain $testKey")
    }
  }
}
```

I may just skip my pr since you are fixing all these things on master anyway

nikolovivan:
@RussellSpitzer, I took your changes to `CassandraDataSourceSpec` and it's now all green locally.

RussellSpitzer:
test this please

RussellSpitzer:
If jenkins complains, i'll turn off the parallel build framework since that's the only difference between Jenkins and travis

RussellSpitzer:
Jenkins approves!!
Ok, I think we just need to clean up our commits a bit here and we should be good to go. Maybe we rebase this into, 
Remove scala2.10 commit
Spark 2.3 Upgrade and Test Fixes commit

Then merge this

Unless you have another structure?

nikolovivan:
Sounds good to me. I'll do it now.

RussellSpitzer:
Looks clean!

PR #1190, opened 2018-07-06
SPARKC-541: Ignore Solr_query when setting up Writing

RussellSpitzer:
DSE includes a virtual column "solr_query" which was used for
interacting with Solr in DSE. This column cannot actually be written too
so we should remove references to the column when writing.


RussellSpitzer:
please test this

RussellSpitzer:
test this please

PR #1196, opened 2018-09-07
SPARKC-550: Fix Thread Safety Issue in AnyObjectFactory

RussellSpitzer:
If multiple threads attempt to convert an object at the same time
there is a chance of racing on filling the temporary arguement buffer.
The buffer is used to add in extra arguments that may be required if the
class has certain external references. To fix this we remove the buffer
entirely and add in the extra args (if they exist) by adding them
to the passed args with sequence concatination.

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/AnyObjectFactory.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/AnyObjectFactory.scala
@@ -43,8 +43,6 @@ class AnyObjectFactory[T: TypeTag] extends Logging with Serializable {
   // the right constructor exists or not, before the job is started
   val argCount: Int = constructorParamTypes.length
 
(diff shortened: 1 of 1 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/reader/AnyObjectFactory.scala
```diff
@@ -75,12 +73,12 @@ class AnyObjectFactory[T: TypeTag] extends Logging with Serializable {
...
 
   @transient
-  private lazy val argBuffer = {
-    val buffer = Array.ofDim[AnyRef](argOffset + argCount)
+  private lazy val outerInstanceArg: Seq[AnyRef] = {
     if (isRealMemberClass(javaClass)) {
-      buffer(0) = resolveDirectOuterInstance()
+      Seq(resolveDirectOuterInstance)
+    } else {
+      Seq.empty
```
pkolaczk: You could have just used `Option`, but it is ok as well.

PR #1197, opened 2018-09-13
SPARKC-551: Fix CassandraSourceRelation Option Parsing in Spark 2.0

RussellSpitzer:
A previous change to deal with the CaseInsensitiveMap in Spark 2.2
caused issues for users of Spark 2.0 which behaves slightly differently
in passing it's map of operation specific parameters to the Datasource.
To avoid issues we will check for both mixed-case and lower-cased
versions of all properties.

In addition I updated the C* test version so that it will work with more
modern Java runtimes.


RussellSpitzer:
test this please

PR #1198, opened 2018-09-17
Correct travis URL

spring1843:
Fixes the URL for the build status from Travis CI. Seems like the old links shows a wrong broken build while master build is actually working.

I got the correct link from https://travis-ci.org/datastax/spark-cassandra-connector


spring1843:
done.

spring1843:
Looks like it's fixed.

PR #1200, opened 2018-11-02
Add support for spark configuration spark.sql.caseSensitive

wankunde:
Now spark-cassandra-connector tool is case sensitive by default when we create a cassandra table.
We can use spark sql configuration parameter **spark.sql.caseSensitive**  to determine if cassandra table is case sensitive or not !


wankunde:
I had signed the CLA,but the status is still cla-missing ?
Can any one help me?
[FWD__Spark_Cassandra_Connector_CLA.pdf](https://github.com/datastax/spark-cassandra-connector/files/2618376/FWD__Spark_Cassandra_Connector_CLA.pdf)

PR #1205, opened 2019-01-17
SPARKC-558: RegularStatements not Cached by SessionProxy

RussellSpitzer:
Previously the case matching in the SessionProxy incorrectly stated that
the RegularStatement class was a String, this meant that Regular
Statements would never actually get matched. To fix this we correct the
class type and in addition we add some tests to actually check the
cache. This still can't avoid all reprepares since the TrieMap is a "at
least once" guarentee and can generate extra calls sometimes.

### DIFF
--- a/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/cql/CassandraConnectorSpec.scala
+++ b/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/cql/CassandraConnectorSpec.scala
@@ -1,6 +1,7 @@
 package com.datastax.spark.connector.cql
 
-import com.datastax.driver.core.{HostDistance, ProtocolOptions}
+import com.datastax.driver.core.querybuilder.QueryBuilder
+import com.datastax.driver.core.{HostDistance, ProtocolOptions, RegularStatement, SimpleStatement}
 import com.datastax.spark.connector.{SparkCassandraITFlatSpecBase, _}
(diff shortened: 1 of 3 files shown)


RussellSpitzer:
@jtgrabowski

spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/cql/CassandraConnectorSpec.scala
```diff
@@ -56,14 +57,30 @@ class CassandraConnectorSpec extends SparkCassandraITFlatSpecBase with BeforeAnd
...
+  it should "cache String PreparedStatements" in {
     conn.withSessionDo { session =>
+      PreparedStatementCache.remove(session.getCluster)
       session.execute(createKeyspaceCql)
       session.execute(s"DROP TABLE IF EXISTS $ks.pstmt")
       session.execute(s"CREATE TABLE $ks.pstmt (key INT PRIMARY KEY, value TEXT)")
-      val stmt1 = session.prepare(s"INSERT INTO $ks.pstmt (key, value) VALUES (?, ?)")
-      val stmt2 = session.prepare(s"INSERT INTO $ks.pstmt (key, value) VALUES (?, ?)")
-      assert(stmt1 eq stmt2)
+      for (_ <- 1 to 10) {
```
jtgrabowski: we could have an assertion for 0 size here

PR #1211, opened 2019-04-09
Load Yaml Template from Proper ClassLoader

rs017991:
Loading from the system ClassLoader makes unnecessary (and in my situation, incorrect) assumptions about the classpath hierarchy.

It is more reliable to stick with the ClassLoader that loaded this class (and thus the spark-cassandra-connector-embedded Jar)


RussellSpitzer:
This sounds good to me, let me see if I can trigger off the tests

RussellSpitzer:
test this please

RussellSpitzer:
Merged, Thanks for your contribution

PR #1214, opened 2019-05-31
Create spark partitions equivalent to Cassandra partitions instead of 1

sambekar15:
Currently even if predicates are pushed down , only 1 spark partition is created. 
So u loose parallelism if your query is scanning across multiple cassandra partitions.
For eg - For a table with partitionKey = (dt,billing_country,offering,category,name) 
**current design creates only 1 spark partition for a query =>**
val results = spark.sql("select * from event where dt in (to_date('2019-03-07'),to_date('2019-03-08')) " +"and billing_country in ('us') and  offering in ('qbn subscription','qbo') and category in " +
          "('billing','buy','trial','product','ipd') and name " +
          "in ('wwsuiwidgetinit','billing_success','billing_failure','pageview.app_homepage','im')")
Modified the code base to create **Spark partitions = Cassandra partitions when predicates are pushed down** 
So for above query , it creates = 2*1*2*5*5 = 100 partitions. This results in query running faster.
Have created a pull request to get initial feedback but there is still refactoring required in coed base.
Opportunities  I see is 1) creating Spark partitions = Cassandra partitions or specify the number of spark partitions in case too may partitions are created. 
2)  Cassandra Partitioner accepts Key Type as tuple. Modified the CassandraTableScanRdd to determine spark partitions based on where clause. Keys are represented as tuples. Modified the CassandraPartitionGenerators to take PKs and generate token range based on that .
Please take a look and let me know your feedback

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableScanRDD.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraTableScanRDD.scala
@@ -21,7 +21,7 @@ import org.apache.spark.{Partition, Partitioner, SparkContext, TaskContext}
 import scala.collection.JavaConversions._
 import scala.language.existentials
 import scala.reflect.ClassTag
-
+import scala.reflect.runtime.universe._
 
 /** RDD representing a Table Scan of A Cassandra table.
   *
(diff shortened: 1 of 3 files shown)


RussellSpitzer:
A similar feature was included in our merge from the DS Fork, check the code in CassandraSourceRelation dealing with "InPredicates" to see if that is similar to what you wanted. If not please file a ticket on our OSS Jira and we can discuss how we can address it.

PR #1216, opened 2019-07-08
SPARKC-458 Scala 2.12 support

YannMoisan:
# Motivation

As Spark now supports Scala 2.12 ([SPARK-14220](https://issues.apache.org/jira/browse/SPARK-14220)), this connector should also support Scala 2.12, to not prevent users from upgrading.

Here is the corresponding Jira ticket : [SPARKC-458](https://datastax-oss.atlassian.net/projects/SPARKC/issues/SPARKC-458?filter=allopenissues&orderby=priority%20DESC&keyword=2.12)

# Status of this PR
- The submodule `spark-cassandra-connector` compiles
- Unit tests are ok
- Integration tests don't work because it tries to run them against Spark 1.4.0 and Spark 2.4.x is required for Scala 2.12
- scoverage has been deactived (cf https://github.com/datastax/spark-cassandra-connector/pull/1216/commits/8169a61180ffe743166b9d5484a248642470f5c1)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,6 @@
+2.4.2
+  * Support for Scala 2.12 (SPARKC-458)
+
 2.4.1
  * Includes all up to 2.3.3
 
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -45,7 +45,8 @@ object Settings extends Build {
   lazy val buildSettings = Seq(
     organization         := "com.datastax.spark",
     version in ThisBuild := currentVersion,
-    scalaVersion         := Versions.scalaVersion,
+    scalaVersion         := Versions.scala211,
+    crossScalaVersions   := Seq(Versions.scala211, Versions.scala212),
     crossVersion         := CrossVersion.binary,
     versionStatus        := Versions.status(scalaVersion.value, scalaBinaryVersion.value)
   )
--- a/project/Versions.scala
+++ b/project/Versions.scala
@@ -18,10 +18,11 @@ import scala.util.Properties
 object Versions {
 
 
-  lazy val scalaVersion = "2.12.8"
+  lazy val scala211 = "2.11.7"
+  lazy val scala212 = "2.12.8"
 
   /* For `scalaBinaryVersion.value outside an sbt task. */
-  lazy val scalaBinary = scalaVersion.dropRight(2)
+  lazy val scalaBinary = scala212.dropRight(2)
 
   val Akka            = "2.3.4"
   val Cassandra       = "3.11.3"
--- a/project/Versions.scala
+++ b/project/Versions.scala
@@ -18,7 +18,7 @@ import scala.util.Properties
 object Versions {
 
 
-  lazy val scalaVersion = "2.11.12"
+  lazy val scalaVersion = "2.12.8"
 
   /* For `scalaBinaryVersion.value outside an sbt task. */
   lazy val scalaBinary = scalaVersion.dropRight(2)
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -20,8 +20,6 @@ addSbtPlugin("com.scalapenos" % "sbt-prompt" % "1.0.0")
 
 addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "0.8.0")
 
-addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.0.4")
-
 //SbtAssembly 0.12.0 is included in sbt-spark-package
 resolvers += "Spark Packages Main repo" at "https://dl.bintray.com/spark-packages/maven" 
 addSbtPlugin("org.spark-packages" %% "sbt-spark-package" % "0.2.5")
--- a/spark-cassandra-connector/scala-2.12/src/main/scala/com/datastax/spark/connector/util/Reflect.scala
+++ b/spark-cassandra-connector/scala-2.12/src/main/scala/com/datastax/spark/connector/util/Reflect.scala
@@ -1,0 +1,21 @@
+package com.datastax.spark.connector.util
+
+import scala.reflect.runtime.universe._
+
+private[connector] object Reflect {
+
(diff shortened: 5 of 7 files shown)


RussellSpitzer:
Sounds good, we'll definitely want to get the integration tests running as well. I aiming the connector at a 2.12 version of spark for IT tests seems reasonable?

YannMoisan:
@RussellSpitzer I have some questions :
- is it ok to add scala 2.12 support in version 2.4.2 ?
- do we want to cross compile the connector for 2.11 and 2.12 ?
- Where comes from the dependency to Scala 2.10 for scoverage (cf below). If I understand well, the support for Scala 2.10 has been dropped ?

```
[warn] 	Note: Some unresolved dependencies have extra attributes.  Check that these dependencies exist with the requested attributes.
[warn] 		org.scoverage:sbt-scoverage:1.3.0 (scalaVersion=2.10, sbtVersion=0.13)
```

> I aiming the connector at a 2.12 version of spark for IT tests seems reasonable?
yes, I just didn't find yet how to do that.

RussellSpitzer:
Adding support would be fine but we would require cross compiling since the default build of Spark is still 2.11 (except for the mistaken 2.4.2 build - 2.4.3 is back to 2.11) We can drop 2.10 related libs, that must have just been forgotten.

YannMoisan:
I'm still wandering why integration tests are trying to run against Spark 1.4.0 (cf below)

Any help will be appreciated

```
[SPARKC-458-scala-212-support] root/spark-cassandra-connector: ++ 2.12.8
[info] Setting version to 2.12.8
[info] Reapplying settings...

  Scala: 2.12.8
  Scala Binary: 2.12
  Java: target=1.7 user=1.8.0_131
  Cassandra version for testing: 3.11.3 [can be overridden by specifying '-Dtest.cassandra.version=<version>']

[info] Set current project to spark-cassandra-connector (in build file:/Users/yann.moisan/projects/spark-cassandra-connector/)
[SPARKC-458-scala-212-support] root/spark-cassandra-connector: it:test
[info] Updating cassandra-server
[info] Resolved cassandra-server dependencies
[trace] Stack trace suppressed: run last cassandra-server/*:coursierResolutions for the full output.
[error] (cassandra-server/*:coursierResolutions) coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
[error]     org.apache.spark:spark-core_2.12:1.4.0:
```

jrciii:
@YannMoisan I wonder if the mysterious 1.4.0 is coming from here: https://github.com/databricks/sbt-spark-package/blob/a8fd991f9b68e1961b997ee11c3516eb714cbd69/src/main/scala/sbtsparkpackage/SparkPackagePlugin.scala#L60

jrciii:
@YannMoisan Looks like this issue is tracked here https://github.com/databricks/sbt-spark-package/issues/45

But the project might be dead.

jrciii:
Adding "sparkVersion := Versions.Spark" to the cassandraServerProject's settings in SparkCassandraConnector build will get you past the spark 1.4.0 issue, but then some integration tests are failing (on my environment anyway. I noticed some tests even failing with the vanilla repo though)

simpadjo:
Thank you for your work.
I missed this PR and fixed it independently: https://github.com/datastax/spark-cassandra-connector/pull/1219/files

When this one gets merged I can rebase and reopen my PR since it also contains fixes for scripts.

YannMoisan:
hello @simpadjo . Have you solved integration tests issue in your PR ?

YannMoisan:
@jrciii with the suggested modification, ITs don't compile anymore.

```
[error] /Users/yann.moisan/projects/spark-cassandra-connector/spark-cassandra-connector/src/it/scala/com/datastax/spark/connector/repl/CassandraRDDReplSpec.scala:24: not found: value SparkRepl
[error]     val output = SparkRepl.runInterpreter(
[error]                  ^
```

simpadjo:
@YannMoisan no, it fails. We could run tests only for 2.11 for now and it would be much better than having no scala-2.12 build at all. But seems that this issue is not Datastax's priority.

YannMoisan:
@simpadjo do you have publish to a private repo ?
I'm trying to do that and publish triggers compile:doc and that command fails
```
[error] /Users/yann.moisan/projects/spark-cassandra-connector/spark-cassandra-connector/src/main/java/com/datastax/spark/connector/japi/CassandraJavaUtil.java:149: not found: value TypeTags
[error]     public static <T> TypeTags.TypeTag<T> safeTypeTag(Class<T> main) {
[error]                       ^
```

simpadjo:
Hm, didn't try it. I ran only `compile` and `test`.

hurelhuyag:
Any updates on this?
btw seems spark 3.0 will release compatible with only scala 2.12.

RussellSpitzer:
Looked into this, the issue is the spark package plugin automatically add's itself to all projects, since the cassandra-server project is a project, it auto got Spark packages added to it. Adding the spark package settings to that project fixes the issue but there is still a compile error. I'll fix that and then push a patch if that's ok

RussellSpitzer:
The repl error is related to the Embedded Cassandra module having scala version based source trees, but the above pr does not include a scala 2.12 source folder so you get the missing class issue.

RussellSpitzer:
Added commits to new PR, based in this repo since I can't modify @YannMoisan 's original 
https://github.com/datastax/spark-cassandra-connector/pull/1223

bergmark:
Thanks for working on this!

FWIW we have been running this PR (it was our last 2.11 dependency) and it is working well!

RussellSpitzer:
There is definitely at least one incompatibility in the Java Type Conversion api (no issues in the scala api tests), it looks (from the test results) like java types are getting replaced with Scala ones. I'll investigate this today to see if it's actually an issue (since it happens internally I believe)

RussellSpitzer:
Done, I've triggered the Sonatype Release 2.4.2 and will make the doc changes tomorrow.

PR #1219, opened 2019-08-16
Support for scala 2.12

simpadjo:
I've been using my own for of spark-cassandra-connector with scala-2.12 from the day one and it works without any problems. This PR brings back cross-version build and adds scala-2.12 to it.
`sbt test` and `sbt assembly` succeed for both version of scala.

If it looks good then documentation should be tweaked as well.

One problem that I faced: when I try to load the project into Intellij or run integration tests it tries to pull 'spark-1.4.0' for some reason:

> [error] (cassandra-server/*:coursierResolutions) coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
[error]     org.apache.spark:spark-core_2.12:1.4.0:

Not sure where it comes from. Is 'cassandra-server' project still relevant?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ sudo: required
 dist: trusty
 scala:
   - 2.11.12
+  - 2.12.9
 
 env:
  - CASSANDRA_VERSION=2.1.15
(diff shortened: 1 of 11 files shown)


simpadjo:
Sorry, missed the existing PR. Closing this one.

PR #1221, opened 2019-10-07
Fixed quick_start documentation

HadesArchitect:
0_quick_start.md uses incorrect version 2.4.1-s_2.11, although it should be 2.4.0-s_2.11

**Before**: the docs misguide to use commands like `bin/spark-shell --conf spark.cassandra.connection.host=127.0.0.1 --packages datastax:spark-cassandra-connector:2.4.1-s_2.11` - _fails_, `module not found`
**After**: works correctly, runs shell with the connector enabled


RussellSpitzer:
While 2.4.1 is out, it seems like the databricks packages website is done. I'll change the coordinate to point to maven instead.

RussellSpitzer:
Thanks for pointing this out, the coordinates are all updated on master now

PR #1223, opened 2019-11-21
Yann moisan sparkc 458 scala 212 support

RussellSpitzer:


RussellSpitzer:
Bunch of tests failing, i'll have to dig in tomorrow but they don't look related to the change ... although except for maybe the reflection change and type conversion errors in the java tests.

RussellSpitzer:
There seems to still be an odd "table not found error in some tests" seems to be flaky, i'll look into this later but I think we are good for now.

RussellSpitzer:
Not actually flakey ... something is going wrong in the DataSource configuraiton

RussellSpitzer:
```
implicit val cassandraConnector/*: CassandraConnector*/= connector
  override def insert(data: DataFrame, overwrite: Boolean):     
...
    val columns = SomeColumns(data.columns.map(x => x: ColumnRef): _*)
    data.rdd.saveToCassandra(tableRef.keyspace, tableRef.table, columns, writeConf)
}
```
Passes


 ```
 override def insert(data: DataFrame, overwrite: Boolean):     
...
    val columns = SomeColumns(data.columns.map(x => x: ColumnRef): _*)
    data.rdd.saveToCassandra(tableRef.keyspace, tableRef.table, columns, writeConf)
}
implicit val cassandraConnector/*: CassandraConnector*/= connector
```
falls back to default (fails test)


So it looks like scala 2.12 has changed the scoping rules slightly? I can't repo this in the repl so maybe it ignores it... anyway quick fix

**Comment by RussellSpitzer** [comment]
``` scala
case class Foo ( val x: Int = 5)
object F {
  def implicitFun(implicit foo: Foo = Foo()) { println(foo)}
}
object BNA {
  println("Before No Annotation")
  implicit val x = Foo(20)
  def f() {F.implicitFun}
}
object ANA {
  println("After No Annotation")
  def f() {F.implicitFun}
  implicit val x = Foo(20)
}
object BWA {
  println("Before With Annotation")
  implicit val x: Foo = Foo(20)
  def f() {F.implicitFun}
}
object AWA {
  println("After With Annotation")
  def f() {F.implicitFun}
  implicit val x: Foo = Foo(20)
}
BNA.f
ANA.f
BWA.f
AWA.f
```

Scala 2.12
```
Before No Annotation
Foo(20)
After No Annotation
Foo(5)
Before With Annotation
Foo(20)
After With Annotation
Foo(20)
```

Scala 2.11
```
Before No Annotation
Foo(20)
After No Annotation
Foo(20)
Before With Annotation
Foo(20)
After With Annotation
Foo(20)
```

rs017991:
Probably this: https://www.scala-lang.org/news/2.12.0/#inferred-types-for-fields
> In particular, an implicit val that did not need an explicitly declared type in 2.11 may need one now.

RussellSpitzer:
> Probably this: https://www.scala-lang.org/news/2.12.0/#inferred-types-for-fields
> 
> > In particular, an implicit val that did not need an explicitly declared type in 2.11 may need one now.

Interesting, but apparently just the placement fixes it, ie having no explicit type, but moving it above the reference fixes the issue as well :/

RussellSpitzer:
Ok we are clean now (the jenkins failure is do to the Repl test not being setup correctly on our jenkins which I don't feel like fixing since it isn't even running scala 2.12) merging

Thanks @YannMoisan for the patch and sorry for the delay! Hopefully we can get back to more active contributions in the near future.

YannMoisan:
Thanks a lot @RussellSpitzer for terminating the work !

PR #1228, opened 2020-02-06
SPARKC-504: Save null UDTs

acustiq:


RussellSpitzer:
Sorry, I really messed up the master branch so a rebase will be necessary. If you could please rebase on b3.0 and I will review as soon as possible.

driver/src/main/scala/com/datastax/spark/connector/mapper/MappedToGettableDataConverter.scala
```diff
@@ -208,7 +208,7 @@ object MappedToGettableDataConverter extends StrictLogging {
           for (i <- columnValues.indices)
             columnValues(i) = converters(i).convert(columnValues(i))
           struct.newInstance(columnValues: _*)
-        case None =>
+        case _ =>
```
RussellSpitzer: Maybe i'm just paranoid, but should we do None | null? I may just be paranoid (this wouldn't be the first time...)

RussellSpitzer:
All tests pass! Thanks so much for your pr, please let me know you response to my above comment. I'll change the target of this pr to b3.0 to make sure it gets in the next release.

RussellSpitzer:
Ah we have a few of the master commits (just docs) got dragged in, could you please rebase on b3.0?

driver/src/main/scala/com/datastax/spark/connector/mapper/MappedToGettableDataConverter.scala
acustiq: I would think that the _ wildcard symbol catches both None and null as it matches anything:

> case _ =>         // Wild card pattern -- matches anything

Source: https://docs.scala-lang.org/tutorials/FAQ/finding-symbols.html

acustiq:
@RussellSpitzer please close this PR in favor of #1241  which has this change rebased on b3.0 .

PR #1230, opened 2020-03-11
SPARKC-570 Add ReadConf parameter to Cassandra table join methods

mathgeek:


mathgeek:
https://datastax-oss.atlassian.net/projects/SPARKC/issues/SPARKC-570

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala
```diff
@@ -152,7 +152,8 @@ class RDDFunctions[T](rdd: RDD[T]) extends WritableToCassandra[T] with Serializa
   def joinWithCassandraTable[R](
     keyspaceName: String, tableName: String,
     selectedColumns: ColumnSelector = AllColumns,
-    joinColumns: ColumnSelector = PartitionKeyColumns)(
+    joinColumns: ColumnSelector = PartitionKeyColumns,
+    readConf: ReadConf = ReadConf.fromSparkConf(rdd.sparkContext.getConf))(
```
edeesis: should these be implicits like SparkContextFunctions#cassandraTable?

mathgeek: I tried to keep this consistent with other functions in this file, which pass a defaulted WriteConf (RDDFunctions#saveToCassandra)

RussellSpitzer:
I think we are also gonna need at least one integration test making sure we respect the readconf passed in, but other than that this seems very reasonable to me.

mathgeek:
I found out that I apparently need to double check with my company's legal department as well before contributing 😅 Once I get that approval I'll add a test. Thanks for looking!

RussellSpitzer:
Ha, well I can always do a black box implementation of this as well :) I promise I won't look too close

RussellSpitzer:
Since I just clobbered master this will need a slight rebase (just needs to be relocated to the new module location)

RussellSpitzer:
https://github.com/datastax/spark-cassandra-connector/pull/1242 Working on rebase here

PR #1237, opened 2020-03-25
Sparkc 574

RussellSpitzer:


connector/src/it/scala/com/datastax/spark/connector/doc/DocCheck.scala
```diff
@@ -5,19 +5,20 @@ import com.datastax.spark.connector.util.{ConfigCheck, RefBuilder}
 import java.io.File
 
 import com.datastax.spark.connector.cluster.DefaultCluster
+import org.scalatest.{FlatSpec, Matchers}
 
-class DocCheck extends SparkCassandraITFlatSpecBase with DefaultCluster {
+class DocCheck extends FlatSpec with Matchers {
```
absurdfarce: Probably more of a question than a comment: can you elaborate on the dropping of DefaultCluster here?  I can't see any relationship between what it provides and your changes here, although it's entirely possible I'm missing something obvious.

connector/src/main/scala/com/datastax/spark/connector/util/ConfigCheck.scala
```diff
@@ -28,8 +27,7 @@ object ConfigCheck {
     CassandraSourceRelation,
     CassandraConnectorConf,
     AuthConfFactory,
-    CassandraConnectionFactory,
-    DseConnectionParams)
+    CassandraConnectionFactory)
```
absurdfarce: I'm not sure I understand the point of ConfObjects.  It doesn't appear to be used anywhere... is it just here to force realization of the various config parameter collections in order to achieve some kind of effect?

connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
```diff
@@ -558,7 +557,7 @@ object CassandraSourceRelation extends Logging {
 
   val SearchPredicateOptimizationRatioParam = ConfigParameter[Double] (
     name = "spark.sql.dse.search.autoRatio",
-    section = DseReferenceSection,
+    section = ReferenceSection,
```
absurdfarce: So, just to make sure I have this straight... we're keeping the "dse" bits in the name even though we're going out of our way to move the docs for these props out of sections that are clearly tied to DSE.  Is this... just a backwards-compatibility issue?

connector/src/it/scala/com/datastax/spark/connector/doc/DocCheck.scala
RussellSpitzer: The "DefaultCluster" means that it starts up a C* CCM Cluster in this test suite, since this test suite doesn't actually use a C* cluster I removed it so I could run the test faster ;)

connector/src/main/scala/com/datastax/spark/connector/util/ConfigCheck.scala
RussellSpitzer: The comment says
  /** Force creation of all Parameters. Any New Parameter Holding Objects should be added here **/

Which I honestly do not remember, maybe there was something about lazy initialization at some point in the past, where you wouldn't get certain parameters if the class was up yet ...

This codebase has gone through a  lot of changes, but I think it's to explicitly make sure we don't get a weird NPE below because we are trying to check properties and an object has init'd yet. Pretty sure that's it

connector/src/main/scala/org/apache/spark/sql/cassandra/CassandraSourceRelation.scala
RussellSpitzer: I think we should change all of the names on these parameters, I was going to do a big rename ticket later.

connector/src/main/scala/com/datastax/spark/connector/util/ConfigCheck.scala
absurdfarce: I was just wondering if this is something we could get rid of then, but I don't think it is.  The various param containers are using ConfigParameter.apply() to create their instances and that's what's used to populate the staticParameters collection.

:sadpanda:

RussellSpitzer: I'll delete it and see if everything still works :)

PR #1239, opened 2020-03-30
[SPARKC-524] Generate a CQL Create Table from DataFrame with options

alexott:
Generate a CQL Create Table statement using a DataFrame and all the parameters required to
construct the CREATE Table statement, like, `IF NOT EXISTS`, `WITH CLUSTERING ORDER BY`,
etc.

### DIFF
--- a/connector/src/main/scala/com/datastax/spark/connector/DatasetFunctions.scala
+++ b/connector/src/main/scala/com/datastax/spark/connector/DatasetFunctions.scala
@@ -17,11 +17,11 @@ class DatasetFunctions[K: Encoder](dataset: Dataset[K]) extends Serializable {
   }
 
   /**
-    *  Creates a C* table based on the Dataset Struct provided. Optionally
-    *  takes in a list of partition columns or clustering columns names. When absent
-    *  the first column will be used as the partition key and there will be no clustering
-    *  keys.
-    */
+   *  Creates a C* table based on the Dataset Struct provided. Optionally
+   *  takes in a list of partition columns or clustering columns names. When absent
+   *  the first column will be used as the partition key and there will be no clustering
+   *  keys.
+   */
   def createCassandraTable(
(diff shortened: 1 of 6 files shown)


RussellSpitzer:
This looks good to me, we had a discussion of a few things though to expand it on slack:

* Maybe use the ClusteringOrdering class for ASC Desc? (not Boolean)
* Possible to add TableOptions Map<String, String> for generic other options

alexott:
Changes are made as discussed, plus added a function for generation of table from DataFrame - previous version was only for RDDs...

connector/src/main/scala/com/datastax/spark/connector/DatasetFunctions.scala
```diff
@@ -54,18 +76,20 @@ class DatasetFunctions[K: Encoder](dataset: Dataset[K]) extends Serializable {
...
           .zipWithIndex
-          .map { case (col, index) => col.copy(columnRole = ClusteringColumn(index))}
+          .map { case (col, index) => col._1.copy(columnRole = ClusteringColumn(index, col._2))}
       ,
       regularColumns = regularColumnNames
           .map(regularColumnName =>
              columnMapping.getOrElse(regularColumnName,
                   throw missingColumnException(regularColumnName, "Regular Column")))
           .map(_.copy(columnRole = RegularColumn))
+      , ifNotExists = ifNotExists, tableOptions = tableOptions
```
RussellSpitzer: usually we would have a carriage return between args

alexott: let me push the change with CR between args...

RussellSpitzer:
Thanks! I was going to just do it for you and merge but this makes my life
easier :P

RussellSpitzer:
Thanks @alexott

PR #1240, opened 2020-03-30
SPARKC-575: Change Dse Prefixed Parameters

RussellSpitzer:
The DS Merge brought in a few parameters for DSE Specific features which
didn't follow the spark.cassandra pattern of everything else. We changed
all the parameters which don't work specifically with dse to generic
spark.cassandra.

Options for Datasource Tables exclusively are also moved into their own
section.


RussellSpitzer:
Now that you mention it, I think that's a good idea, it will break some mild backwards compatibility with DSE SparkConnector fork but i'm not sure it will really effect anyone unless you have mixed versions which I don't want to support anyway.

PR #1241, opened 2020-03-31
SPARKC-504: Allow null UDTs against b3.0 branch

acustiq:
Same as PR #1228 rebased on b3.0

### DIFF
--- a/driver/src/main/scala/com/datastax/spark/connector/mapper/MappedToGettableDataConverter.scala
+++ b/driver/src/main/scala/com/datastax/spark/connector/mapper/MappedToGettableDataConverter.scala
@@ -208,7 +208,7 @@ object MappedToGettableDataConverter extends StrictLogging {
           for (i <- columnValues.indices)
             columnValues(i) = converters(i).convert(columnValues(i))
           struct.newInstance(columnValues: _*)
-        case None =>
+        case _ =>
           null.asInstanceOf[struct.ValueRepr]
       }
     }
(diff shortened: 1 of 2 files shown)


driver/src/main/scala/com/datastax/spark/connector/mapper/MappedToGettableDataConverter.scala
```diff
@@ -208,7 +208,7 @@ object MappedToGettableDataConverter extends StrictLogging {
           for (i <- columnValues.indices)
             columnValues(i) = converters(i).convert(columnValues(i))
           struct.newInstance(columnValues: _*)
-        case None =>
+        case _ =>
```
RussellSpitzer: Yeah my comment on _ was just that I thought we would explicitly match None and null just incase some other object got to this point in the match. I know that's impossible I just worry sometimes.

RussellSpitzer: If you see the case on line 204, "case obj if obj == null" Why doesn't this case match?

RussellSpitzer: I tried 

`  it should "convert null UDTs to null" in {
    val userTable = newTable(loginColumn, passwordColumn)
    val converter = MappedToGettableDataConverter[UserWithOption](userTable, userTable.columnRefs)
    val row = converter.convert(null)
    row shouldEqual null
  }`

And this test passes without this code change, I think I may be missing something here

PR #1242, opened 2020-03-31
SPARKC-570 Add ReadConf parameter to Cassandra table join methods

RussellSpitzer:


edeesis:
Thanks for putting this together for us cc @mathgeek.

Any possibility of this going in a 2.4.4?

RussellSpitzer:
I could do that, but is there any reason you wouldn't want it in the 3.0.0 release? I probably would do them at the same time.

edeesis:
Ah I saw the base branch and thought this wouldn’t go to 2.4.4.

I’m good with either 2.4.4 or waiting for 3.0, just didn’t know what the time frame of 3.0 is.

RussellSpitzer:
My hope was 3.0.0 is out by the end of the month, but I'll probably cut
another alpha end of this week

edeesis:
Hey @RussellSpitzer. I've noticed you've cut a 2.5 release that contains this change. However, we since been made aware of an incompatibility with the Cassandra 4.5 driver, so we cannot upgrade to that version. Any chance of this getting put on the 2.4.x branch?

In case it wasn't obvious, @mathgeek and I work at the same company lol

RussellSpitzer:
Ah are you using the driver elsewhere in your app? I think we could
possibly do some backports to 2.4 as well if that's required.

@Jarek what do you think?

edeesis:
Not exactly. Parts of our application are reading from an older Cassandra Ring (2.0 I think) that other testing shows that the upgrade to 4.5 breaks. It wasn’t officially supported in 3.0 but it still worked, so until we get off of that, we’re stuck where we are.

jtgrabowski:
Sure, we can include that in 2.4. I'll create a PR soon.

PR #1244, opened 2020-04-29
re explicitly mark some features as available since SCC 2.5

alexott:
# Description

Fixes one place marked as SCC 3.0 instead of 2.5, and adds similar markers to other places


RussellSpitzer:
@alexott Small thing to do to fix the tests, you changed a parameter in the code but didn't run the reference file generation script. So the tests are failing because they do not match the programmed parameters.

] - should match a freshly created reference file *** FAILED ***
[info]   ******* re-run sbt connector/run to regenerate properties file
[info]   *******.
[info]       "...n the number of rows[ ]to be returned by th..." was not equal to "...n the number of rows[]to be returned by th..." (DocCheck.scala:32)

alexott:
fixed...

PR #1245, opened 2020-04-30
SPARKC-577: Removal of Driver Duplicate Classes

absurdfarce:
# Description

Replaced internal refs with Java driver types directly wherever possible.  Internal classes within the connector were used for several different functions:

1. Metadata retrieval/access

In some cases (1) above could be handled with direct replacement, but we were also storing multiple layers of metadata within a single object (i.e. some table level information was stored in ColumnDef).  Unfortunately the 4.x driver doesn't allow traversing the metadata tree in this way.  In order to make this information available without too much breakage to the existing API some of the old classes are preserved with a new role: containers for all metadata types on a "branch" of this tree.  Thus TableDef stores keyspace + table metadata, ColumnDef stores keyspace + table + column metadata, etc.

# How Has This Been Tested?

# Checklist:


connector/src/main/scala/com/datastax/spark/connector/rdd/AbstractCassandraJoin.scala
```diff
@@ -108,8 +107,8 @@ private[rdd] trait AbstractCassandraJoin[L, R] {
       .filter(cc => colNames.contains(cc.columnName))
     if (!tableDef.clusteringColumns.startsWith(chosenClusteringColumns)) {
       val maxCol = chosenClusteringColumns.last
-      val maxIndex = maxCol.componentIndex.get
-      val requiredColumns = tableDef.clusteringColumns.takeWhile(_.componentIndex.get <= maxIndex)
+      val maxIndex = tableDef.clusteringColumns.indexOf(maxCol)
+      val requiredColumns = tableDef.clusteringColumns.take(maxIndex + 1)
```
absurdfarce, line 111: Wanted to highlight this for review.  I'm pretty sure the logic I have in there now mirrors what was being done but wanted to make sure this was looked at more closely.

absurdfarce:
Will likely wind up closing this in favor of https://github.com/datastax/spark-cassandra-connector/pull/1250

PR #1246, opened 2020-05-07
Sparkc 590 - Adding DSV2 Support (Built on SPARKC-589)

RussellSpitzer:
# Description

This is a ticket to Add Datastax V2 Support to the Spark Cassandra Connector

Spark 3.0 introduces the Datasource V2 api which allows for a lot of great new possibilities. In particular it has some great avenues for taking advantage of C* partitioning automatically and working better in conjunction with writing and reading.

This is also most likely going to be the key way to interact with datasources in the future for Spark so it would be great if we could support it.

This PR plus the Catalog PR allows Spark SQL to natively read and write from C* tables as well as preform CreateTable as select queries and lots of other nifty things.

To accomplish this we have to refactor out code from both CassandraTableScan and TableWriter. Both of these classes have a bunch of things that need to be refactored out into pure functions because the new DSV2 api no longer just wraps user provided RDDs. To make sure that RDD and Datasourcev2 code are both still using nearly identical code paths, I removed most of the functions into helper classes, and then made RDD code call these new helpers.

Fixes: [SPARKC-590](https://datastax-oss.atlassian.net/projects/SPARKC-590)

# How Has This Been Tested?

I have only added minimal tests so far for the Catalog Integration. All DSV1 tests now pass as well without changing their apis (with a few exceptions).

# Checklist:

- [x] I have a ticket in the [SPARKC-590](https://datastax-oss.atlassian.net/projects/SPARKC-590)


connector/src/main/scala/org/apache/spark/sql/cassandra/DefaultSource.scala
```diff
@@ -27,93 +32,46 @@ import org.apache.spark.sql.streaming.OutputMode
...
-               |Example: df.write.format.options.mode(SaveMode.Append).save()" '""".stripMargin)
-        }
-      case Ignore =>
-        if (table.buildScan().isEmpty()) {
-          table.insert(data, overwrite = false)
-        }
-    }
-
-    CassandraSourceRelation(tableRef, sqlContext, options, None)
+  override def shortName(): String = "cassandra"
```
RussellSpitzer, line 41: DOC NOTE: "cassandra" is valid format, can't believe we didn't turn this on when it was added (Spark 1.5! )

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogTableReadSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+  it should "handle count pushdowns" in {
+    setupBasicTable()
+    val request = spark.sql(s"""SELECT COUNT(*) from $defaultKs.$testTable""")
+    val reader = request
+      .queryExecution
+      .executedPlan
+      .collectFirst { case batchScanExec: BatchScanExec=> batchScanExec.readerFactory.createReader(EmptyInputPartition)}
+
+    reader.get.isInstanceOf[CassandraCountPartitionReader] should be (true)
+    request.collect()(0).get(0) should be (101)
```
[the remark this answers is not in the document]
RussellSpitzer: There isn't really a V2 read api, but added a .table("tablename") example for programmatic reading. All the rest of the api is covered by DSV1 tests

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,434 @@
...
+
+    connector.withSessionDo(session =>
+      session.execute(alterWithReplication.asCql())
+    )
+  }
+
+  override def loadNamespaceMetadata(namespace: Array[String]): java.util.Map[String, String] = {
+    val ksMetadata = getKeyspaceMeta(connector, namespace)
+
+    (Map[String, String](
```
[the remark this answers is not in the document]
RussellSpitzer: removed the extra casting hopefully it is cleaner to read now

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraWriteBuilder.scala
```diff
@@ -0,0 +1,118 @@
...
+  inputSchema: StructType)
+  extends WriteBuilder with SupportsTruncate {
+
+  val consolidatedConf = consolidateConfs(
+    session.sparkContext.getConf,
+    session.conf.getAll,
+    catalogName,
+    tableDef.keyspaceName,
+    options.asScala.toMap)
+
```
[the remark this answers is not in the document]
RussellSpitzer: Added in test

connector/src/main/scala/com/datastax/spark/connector/datasource/InternalRowWriterFactory.scala
```diff
@@ -0,0 +1,65 @@
...
+  private val columnTypes = columns.map(_.columnType)
+  private val schemaRequestedIndexes = {
+    val indexedSchema = schema.map(_.name).zipWithIndex.toMap
+    selectedColumns.map(column => indexedSchema(column.selectedAs))
+  }
+
+  private val dataTypes = schemaRequestedIndexes.map(index => typeForSchemaIndex(index))
+
+  /*
+  Todo Is this Needed? With Unsafe Rows we had to be more careful about internal representations but
```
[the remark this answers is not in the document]
RussellSpitzer: Conversions needed

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+
+  val OnlyOneNamespace = "Cassandra only supports a keyspace name of a single level (no periods in keyspace name)"
+
+  //Table Support
+  def getTableMetaData(connector: CassandraConnector, ident: Identifier) = {
+    val namespace = ident.namespace
+    checkNamespace(namespace)
+    val tableMeta = getMetadata(connector)
+      .getKeyspace(fromInternal(namespace.head)).asScala
+      .getOrElse(throw nameSpaceMissing(getMetadata(connector), namespace))
```
jtgrabowski, line 389: -> getKeyspaceMeta

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogNamespaceSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+  it should "give no keyspace found when looking up a keyspace which doesn't exist" in {
+    intercept[NoSuchNamespaceException]{
+      spark.sql(s"SHOW NAMESPACES FROM $catalogName.fakenotreal").collect
+    }
+  }
+
+  it should "be able to create a new keyspace" in {
+    dropKeyspace(defaultKs)
+    waitForKeyspaceToExist(defaultKs, false)
+    spark.sql(s"CREATE DATABASE $defaultKs WITH DBPROPERTIES (class='SimpleStrategy',replication_factor='5')")
```
absurdfarce, line 37: Nit: worth validating that defaultKs no longer exists here?  I'm not sure it buys you much but _might_ be useful.

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogNamespaceSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+    waitForKeyspaceToExist(defaultKs, false)
+    val exception = intercept[CassandraCatalogException]{
+      spark.sql(s"CREATE DATABASE $defaultKs WITH DBPROPERTIES (class='foobar',replication_factor='5')")
+    }
+    exception.getMessage should include("Unknown keyspace replication strategy")
+  }
+
+  it should "throw a keyspace exists exception when creating a keyspace that already exists" in {
+    spark.sql(s"CREATE DATABASE IF NOT EXISTS $defaultKs WITH DBPROPERTIES (class='SimpleStrategy',replication_factor='5')")
+    intercept[NamespaceAlreadyExistsException]{
```
absurdfarce, line 62: Need a waitForKeyspaceToExist() here to confirm that we should get the exception we expect below?

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+  override def loadNamespaceMetadata(namespace: Array[String]): java.util.Map[String, String] = {
+    val ksMetadata = getKeyspaceMeta(connector, namespace)
+
+    (ksMetadata.getReplication.asScala + (DurableWrites -> ksMetadata.isDurableWrites.toString))
+      .asJava
+  }
+
+  override def dropNamespace(namespace: Array[String]): Boolean = {
+    checkNamespace(namespace)
+    val keyspace = getMetadata(connector).getKeyspace(fromInternal(namespace.head)).asScala
```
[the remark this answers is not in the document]
RussellSpitzer: FromInternal takes the unquoted mixed case version of the name and quotes it if necessary. So I believe this will NOT match an improperly cased table in SparkSql but I think if spark sql naming is case sensitive we should respect that ... ?

build.sbt
```diff
@@ -3,9 +3,8 @@ import sbt.Keys.parallelExecution
 import sbt.moduleFilter
 import sbt._
 
-lazy val scala211 = "2.11.12"
 lazy val scala212 = "2.12.10"
-lazy val supportedScalaVersions = List(scala212, scala211)
+lazy val supportedScalaVersions = List(scala212)
```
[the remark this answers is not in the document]
RussellSpitzer: Yep sorry! There are even references to 2.10 :|

My current commit i'm working on should tidy all this up plus add a bunch of new docs

connector/src/test/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
```diff
@@ -7,16 +7,6 @@ import org.apache.spark.SparkConf
 
 object SparkTemplate {
 
-  val HiveMetastoreConfig: Map[String, String] = Map (
-    "spark.hadoop.hive.metastore.rawstore.impl" -> "com.datastax.bdp.hadoop.hive.metastore.CassandraHiveMetaStore",
-    "spark.hadoop.cassandra.autoCreateHiveSchema" -> "true",
```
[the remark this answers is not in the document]
RussellSpitzer: they would get ClassNotFound Exceptions, yeah I should tidy that up

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+
+class CassandraCatalogException(msg: String) extends IllegalArgumentException(msg)
+
+/**
+  * A Spark Sql Catalog for inter-operation with Cassandra
+  *
+  * Namespaces naturally map to C* Keyspaces, but they are always only a single
+  * element deep.
+  */
+class CassandraCatalog extends CatalogPlugin
```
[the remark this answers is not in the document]
RussellSpitzer: We do not, here is my doc on it for clarity (will be in the next commit)

> DatasourceV2 makes connecting to Cassandra now easier than ever. Parameters for configuring your connection
> can be done in the SparkConf, SparkSession, spark-defaults file or individually for the Catalog.  Once a catalog is configured it can 
> be accessed through both SparkSql and DataFrames to read from, write to, create, and drop Cassandra tables.
> 
> While setting up a Catalog can be done before starting your application it can also be done at 
> runtime by setting  ```spark.sql.catalog.anyname``` to  ```com.datastax.spark.connector.datasource.CassandraCatalog```
> In your SparkSession. The "anyname" portion can be any name you would like to specify for this 
> particular catalog.
> 
> Then any parameter for that catalog can be set by just appending the parameter name to the catalog name as in
> ```spark.sql.catalog.anyname.propertykey``` to ```propertyvalue```
> 
> Example: Manually configuring a Catalog named Cass100
> 
> ```scala
> spark.conf.set(s"spark.sql.catalog.cass100", "com.datastax.spark.connector.datasource.CassandraCatalog")
> spark.conf.set(s"spark.sql.catalog.cass100.spark.cassandra.connection.host", "127.0.0.100")
> ```
> 
> Once this is set, we can access the catalog by using the triple part identifier of 
> ```catalog.keyspace.table``` which in this case would be ```cass100.ks.tab```

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogTableSpec.scala
```diff
@@ -0,0 +1,343 @@
...
+  "A Cassandra Catalog Table Support" should "initialize successfully" in {
+    spark.sessionState.catalogManager.currentCatalog.name() should include("Catalog cassandra")
+  }
+
+  val testTable = "testTable"
+
+  it should "create a table with a partition key" in {
+    createDefaultKs()
+    spark.sql(s"CREATE TABLE $defaultKs.$testTable (key Int, value STRING) USING cassandra PARTITIONED BY (key)")
+    getTable(defaultKs, testTable).getPartitionKey.get(0).getName.asInternal() should be ("key")
```
absurdfarce, line 21: Do we need/should we have an explicit wait for table creation here and/or a metadata sync before trying to check results from the metadata?  Same question for other tests within this spec I suppose.

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogTableSpec.scala
```diff
@@ -0,0 +1,343 @@
...
+         |PARTITIONED BY (key_1, key_2, key_3)
+         |TBLPROPERTIES (clustering_key='cc1.asc, cc2.desc, cc3.asc')""".stripMargin)
+
+    spark.sql(
+      s"""ALTER TABLE $defaultKs.$testTable ADD COLUMNS (newCol INT) """.stripMargin
+    )
+
+    getTable(defaultKs, testTable)
+      .getColumn(fromInternal("newCol"))
+      .get().getName.asInternal() shouldBe ("newCol")
```
absurdfarce, line 304: Worth confirming that _all_ columns (including the new one) are present now?

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogTableSpec.scala
```diff
@@ -0,0 +1,343 @@
...
+    createDefaultKs()
+    spark.sql(
+      s"""CREATE TABLE $defaultKs.$testTable (
+         |key_1 Int, key_2 Int, key_3 Int,
+         |cc1 STRING, cc2 String, cc3 String,
+         |value String) USING cassandra
+         |PARTITIONED BY (key_1, key_2, key_3)
+         |TBLPROPERTIES (clustering_key='cc1.asc, cc2.desc, cc3.asc')""".stripMargin)
+
+    val excep = intercept[SparkException] {
```
absurdfarce, line 336: In other cases we have a more precise type to indicate specific error cases.  Worth something like a NoSuchTableException for this case?  SparkException seems... awfully general here.

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+  with TableCatalog
+  with SupportsNamespaces
+  with Logging {
+
+  import CassandraCatalog._
+
+  lazy val sparkSession = SparkSession.active
+
+  val ReplicationClass = "class"
```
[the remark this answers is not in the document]
RussellSpitzer: because I like sharing (jk making private)

build.sbt
```diff
@@ -3,12 +3,12 @@ import sbt.Keys.parallelExecution
...
-lazy val scala211 = "2.11.12"
-lazy val scala212 = "2.12.10"
-lazy val supportedScalaVersions = List(scala212, scala211)
+lazy val scala212 = "2.12.11"
+lazy val supportedScalaVersions = List(scala212)
 
 // factor out common settings
-ThisBuild / scalacOptions += "-target:jvm-1.8"
+ThisBuild / scalaVersion := scala212
+ThisBuild / scalacOptions ++= Seq("-target:jvm-1.8")
```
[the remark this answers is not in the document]
RussellSpitzer: We don't need too... I"ll try out some other versions

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+    val userProperties = tableProps.filter{ case (key, _) => CassandraProperties.contains(key) }
+    val unusedProperties = tableProps -- userProperties.keys
+    logDebug(s"Ignoring non-cassandra properties for table $unusedProperties")
+    val createTableWithProperties = userProperties.foldLeft(createTableWithColumns) {
+      case (createStmt, (key, value)) => createStmt.withOption(key, parseProperty(value)).asInstanceOf[CreateTable]
+    }
+
+    connector.withSessionDo(_.execute(createTableWithProperties.asCql()))
+
+    loadTable(ident)
```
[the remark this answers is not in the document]
RussellSpitzer: The only case where the session wouldn't get the metadata (because all DDL now blocks for the metadata update if you remember some of our timeouts we had to fix before) the only case where this could be an issue is if the session cache fails and a new session has to be made and this new session is made so quickly it doesn't pick up the metadata. I think this is pretty unlikely ... I think

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+    //Check for unsupported table changes
+    changes.foreach {
+      case add: AddColumn =>
+      case del: DeleteColumn =>
+      case setProp: TableChange.SetProperty =>
+      case other: TableChange => throw new CassandraCatalogException(s"Cassandra Catalog does not support Alter operation: ${other.getClass.getSimpleName}")
+    }
+
+    val propertiesToAdd = changes.collect { case setProperty: TableChange.SetProperty =>
+      setProperty
```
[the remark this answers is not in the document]
RussellSpitzer: ah yeah I should do that here as well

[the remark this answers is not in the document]
RussellSpitzer: It's not as much of a problem though since here Spark isn't going to forward reserved properties ... I think

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+      connector.withSessionDo(_.execute(setOptionsStatement.asCql()))
+    }
+
+    if (columnsToRemove.nonEmpty) {
+      val dropColumnsStatement = SchemaBuilder.alterTable(keyspace, table).dropColumns(columnsToRemove: _*)
+      connector.withSessionDo(_.execute(dropColumnsStatement.asCql()))
+    }
+
+    if (columnsToAdd.nonEmpty) {
+      val addColumnStatement = columnsToAdd.foldRight(
```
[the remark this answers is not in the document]
RussellSpitzer: because i'm right handed

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+    consolidatedConf = consolidateConfs(sparkConf, sparkSession.conf.getAll, name, userOptions = options.asScala.toMap)
+    connector = CassandraConnector(consolidatedConf)
+    catalogName = name
+    nameIdentifier = connector.conf.contactInfo.endPointStr()
+  }
+
+  override def name(): String = s"Catalog $catalogName For Cassandra Cluster At $nameIdentifier " //TODO add identifier here
+
+  override def listNamespaces(): Array[Array[String]] = {
+    getMetadata(connector)
```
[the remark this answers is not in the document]
absurdfarce: :notbad:

connector/src/it/scala/com/datastax/bdp/spark/search/SearchAnalyticsIntegrationSpec.scala
```diff
@@ -125,42 +126,25 @@ class SearchAnalyticsIntegrationSpec extends SparkCassandraITFlatSpecBase with D
...
-
-      getUnderlyingCqlWhereClause(df.rdd)
+      getCassandraScan(df.queryExecution.sparkPlan).cqlQueryParts.whereClause
     }
   }
 
 
   "SearchAnalytics" should "be able to use solr_query" in skipIfNotDSE(conn) {
-    val df = spark.sql(s"""SELECT key,a,b,c FROM $table WHERE solr_query = '{"q": "*:*", "fq":["key:[500 TO *]", "a:1"]}' """)
+    val df = spark.sql(s"""SELECT key,a,b,c FROM $table.$ks.$table WHERE solr_query = '{"q": "*:*", "fq":["key:[500 TO *]", "a:1"]}' """)
```
absurdfarce, line 147: Noob question: what's with the $table.$ks.$table select here?  What's the leading $table for?

RussellSpitzer: This is weird because the way the test used to do this was to set options on a per table basis, registering each configuration as a specific table referring to similar underlying tables. To rewrite the tests with least amount of confusion I put each set of configuration into a catalog. So there is a catalog for which solr is turned on, there is a catalog for which solr is turned off, and so on. 

So here the leading identifier is for the catalog (which has the same name as the table)

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+        metadata + (setProperty.property() -> setProperty.value)
+      case (removeProperty: RemoveProperty, metadata: mutable.Map[String, String]) =>
+        metadata - removeProperty.property()
+      case (other, _) => throw new CassandraCatalogException(s"Unable to handle alter namespace operation: ${other.getClass.getSimpleName}")
+    }
+
+    val alterStart = SchemaBuilder.alterKeyspace(namespace.head)
+    val alterWithDurable = alterStart.withDurableWrites(ksMeta.getOrElse(DurableWrites, "True").toBoolean)
+    val replicationClass = ksMeta
+      .getOrElse(ReplicationClass, throw new CassandraCatalogException(s"Creating a keyspace requires a $ReplicationClass option"))
```
absurdfarce, line 136: The exception thrown here should probably reference altering a keyspace rather than creating one... maybe something like "cannot alter keyspace because replication class couldn't be determined" or something along those lines.

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+          val replicationFactor = ksMeta.getOrElse(ReplicationFactor,
+            throw new CassandraCatalogException(s"Need a $ReplicationFactor option with SimpleStrategy"))
+          createStmt.withSimpleStrategy(replicationFactor.toInt)
+        case NetworkTopologyStrategy =>
+          val datacenters = (ksMeta -- IgnoredReplicationOptions).map(pair => (pair._1, pair._2.toInt: java.lang.Integer))
+          createStmt.withNetworkTopologyStrategy(datacenters.asJava)
+        case other => throw new CassandraCatalogException(s"Unknown keyspace replication strategy $other")
+      }
+      val finalCreateStmt = createWithReplication.withDurableWrites(ksMeta.getOrElse(DurableWrites, "True").toBoolean)
+      session.execute(finalCreateStmt.asCql())
```
absurdfarce, line 118: In the alter logic below the CQL statement is built up before we go into withConnectorDo(), effectively minimizing the amount of work we have to do when we have access to the session.  It's not a huge thing but we could do the same here since session isn't actually used until we need to execute the CQL.  I suppose that's probably _slightly_ preferred... but it's not a major thing either way.

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+    val dropResult = connector.withSessionDo(session =>
+      session.execute(SchemaBuilder.dropKeyspace(keyspace.getName).asCql()))
+    dropResult.wasApplied()
+  }
+
+  override def listTables(namespace: Array[String]): Array[Identifier] = {
+    getKeyspaceMeta(connector, namespace)
+      .getTables.asScala
+      .map { case (tableName, _) => Identifier.of(namespace, tableName.asInternal()) }
+      .toArray
```
absurdfarce, line 175: Perhaps this is slightly cleaner:

```scala
  override def listTables(namespace: Array[String]): Array[Identifier] = {
    getKeyspaceMeta(connector, namespace)
      .getTables.asScala
      .keys
      .map { cqlid => Identifier.of(namespace, cqlid.asInternal)}
      .toArray
  }
```

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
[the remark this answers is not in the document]
absurdfarce: I agree this should be quite unlikely.  We could add some notion of a minimal set of retries to completely guard against such a case, right?  I'm not sure it's worth it, though...

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    if (columnsToAdd.nonEmpty) {
+      val addColumnStatement = columnsToAdd.foldRight(
+        SchemaBuilder.alterTable(keyspace, table).asInstanceOf[AlterTableAddColumn]
+      ) { case ((colName, dataType), alterBuilder) =>
+        alterBuilder.addColumn(colName, dataType)
+      }.asInstanceOf[AlterTableAddColumnEnd]
+      connector.withSessionDo(_.execute(addColumnStatement.asCql()))
+    }
```
absurdfarce, line 357: By issuing these alters as a sequence of discrete statements rather than a single monolith don't we introduce the possibility of partial application and thus leaving the underlying table in some unexpected state?

RussellSpitzer: From the JavaDoc I wrote :)
```
    * The API expects that if any change is rejected, we  should not apply
    * any changes. This is basically impossible for us since we cannot batch
    * together all of our DDL changes into a single CQL statement. To ameliorate
    * this we perform as many checks as we can up front to try to avoid beginning
    * our DDL requests which we know will fail.
```

connector/src/it/scala/com/datastax/bdp/spark/search/SearchAnalyticsIntegrationSpec.scala
absurdfarce: Ohhh, I see it now... registerTableWithOptions has all the fun.  Thx for the pointer. ;)

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
absurdfarce: What, like I'm supposed to _read docs_ now?  Whatever... ;)

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraInJoinReaderFactory.scala
```diff
@@ -0,0 +1,167 @@
...
+  override def createReader(partition: InputPartition): PartitionReader[InternalRow] =
+    if (cqlQueryParts.selectedColumnRefs.contains(RowCountRef)) {
+      CassandraInJoinCountReader(connector, tableDef, inClauses, readConf, schema, cqlQueryParts, partition)
+    } else {
+      CassandraInJoinReader(connector, tableDef, inClauses, readConf, schema, cqlQueryParts, partition)
+    }
+}
+
+
+abstract class CassandraBaseInJoinReader(
```
absurdfarce, line 37: CassandraInJoinReader is just a concrete subclass of this class.  Should we perhaps just make the current CassandraBaseInJoinReader into CassandraInJoinReader and make CassandraInJoinCountReader extend that?

doc/14_data_frames.md
```diff
@@ -1,548 +1,285 @@
...
-word count
-cat  30
-fox  40
-```
+```spark.sql.catalog.casscatalog``` to  ```com.datastax.spark.connector.datasource.CassandraCatalog```
 
-There are also some helper methods which can make creating Datasets easier. They can 
-be accessed after importing `org.apache.spark.sql.cassandra` package. In the following 
-example, all the commands used to create the Dataset are equivalent:
+This will set up an identifier of "anyname" to point to the catalog for the default Cassandra Cluster
```
jtgrabowski, line 50: casscatalog

doc/14_data_frames.md
```diff
@@ -1,548 +1,285 @@
...
-spark.sessionState.functionRegistry.registerFunction(FunctionIdentifier("ttl"),
-        (input: Seq[Expression]) => CassandraMetadataFunction.cassandraTTLFunctionBuilder(input))
-spark.sessionState.functionRegistry.registerFunction(FunctionIdentifier("writetime"),
-        (input: Seq[Expression]) => CassandraMetadataFunction.cassandraWriteTimeFunctionBuilder(input))
+```sql
+SELECT TTL(v) as t, WRITETIME(t) FROM casscatalog.ksname.testTable WHERE k = -1"
 ```
 
-#### Example Catalyst Optimization with Cassandra Server Side Pushdowns
+For writing there are specfic write options which can be used to assign WriteTime and TTL
```
jtgrabowski, line 229: specific

doc/14_data_frames.md
```diff
@@ -1,548 +1,285 @@
...
-    clusterkey2   BIGINT,
-    clusterkey3   BIGINT,
-    regularcolumn BIGINT,
-    PRIMARY KEY ((partitionkey1, partitionkey2, partitionkey3), clusterkey1, clusterkey2, clusterkey3)
-);
-```
+These methods are still useful if you need to express certain options only for a specific
+read or write, and not for the entire catalog.
+
+We using these methods you are required to set an option for `table` and `keyspace`
```
jtgrabowski, line 270: When

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraInJoinReaderFactory.scala
RussellSpitzer: So the reason this is like that is because you cannot extend case classes, so the only way to keep them case classes is to extend a common class

connector/src/main/scala/com/datastax/spark/connector/rdd/CassandraCoGroupedRDD.scala
```diff
@@ -239,9 +239,8 @@ class CassandraCoGroupedRDD[T](
     scanRDDs.maxBy(_.partitions.length).partitions
   }
 
-  private lazy val nodeAddresses = new NodeAddresses(connector)
   override def getPreferredLocations(split: Partition): Seq[String] =
-    split.asInstanceOf[CassandraPartition[_, _]].endpoints.flatMap(nodeAddresses.hostNames).toSeq
+    split.asInstanceOf[CassandraPartition[_, _]].endpoints
```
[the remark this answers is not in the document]
RussellSpitzer: yep! So previously this got passed around as a set of hostnames which we would then map back out to a list of aliases which we knew men the same thing `nodeAddresses.hostNames`. This would make sure we would get all known interfaces for a node, just incase spark was bound on one and C* was bound on another, or public/private confusions. 

Rather than doing this over and over at the place of consumption I turned things around so instead we just keep all of that information in the partition itself rather than computing it right before we do "preferredLocations"

Tldr;
Before:  partition.oneHost ----> on read ---> mapToAllHostsIdentifiers
Now: partition.AllHostIdentIfiers

jtgrabowski:
We had a major refactoring in this ticket. Quite a few methods were moved to `datasource` package, to ScanHelper and other classes. I think that in a long run it would be beneficial to mirror the changes in lower branches (as much as possible without having Spark 3.0 in the lower branch). Doing this would make porting changes between branches way easier.

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogNamespaceSpec.scala
RussellSpitzer: That's what the line 36 does, and if it DID exist this would actually fail with NamespaceExistsException :)

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogNamespaceSpec.scala
RussellSpitzer: The intercept will fail if no exception is thrown, so I think that covers it

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogTableSpec.scala
RussellSpitzer: I mentioned this a bit previously with Jarek. Since all catalog sessions share the same underlying session (pool) they all have the same metadata access as well. Which means that any statements which do DDL will block until the metadata is updated internally. So we really don't have to, I added in a bunch of these waits because I wanted to verify that things existed in some places but I don't think they are really necessary.

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogTableSpec.scala
RussellSpitzer: Definitely

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogTableSpec.scala
RussellSpitzer: Yeah I have no Idea why this one is like that

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+
+    val providedPartitionKeyNames = partitions.map(_.references().head.fieldNames().head)
+
+    val partitionKeys = if (providedPartitionKeyNames.nonEmpty) {
+      providedPartitionKeyNames
+    } else {
+      tableProps
+        .getOrElse(PartitionKey, throw new CassandraCatalogException(s"Cassandra Tables need partition keys defined in property $PartitionKey or with 'PARTITIONED BY columns"))
+        .split(",")
+        .map(_.replaceAll("\\s", ""))
```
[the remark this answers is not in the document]
RussellSpitzer: will do

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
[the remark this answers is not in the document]
RussellSpitzer: I think it's fine :/ I don't really think there is a logical scenario where the session expires while simultaneously being active enough to make the table and not update the metadata.

In my head:
Session is abandoned, therefor lots of time has past and the metadata is updated 
Session is not abandoned, it is the same session that did the DDL and we blocked till metadata was updated

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
RussellSpitzer: Yep good call

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
RussellSpitzer: yeah I think perf wise it probably doesn't matter, but that does sound cleaner

RussellSpitzer:
@jtgrabowski That makes sense to me, 2.4 and 2.5?

connector/src/test/scala/com/datastax/spark/connector/embedded/SparkTemplate.scala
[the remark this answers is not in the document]
RussellSpitzer: :| I tried to do this but we can't get to the conf till it's too late. By the time we can check the configuration it's already tried to load the catalog and broken.

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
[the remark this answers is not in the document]
RussellSpitzer: Yeah spark controls this, but I can put a little check in getMetadata which warns about an uninit'd connector

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogTableSpec.scala
RussellSpitzer: Ok having run this again I know why this is like this. The exception thrown is 
SparkException( CassandraCatalogException ) so this is correct. Spark wraps our catalog exception in it's own before surfacing it.

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraSourceUtil.scala
```diff
@@ -0,0 +1,233 @@
...
+
+  def sparkSqlToJavaDriverType(
+    dataType: SparkSqlDataType,
+    protocolVersion: ProtocolVersion = ProtocolVersion.DEFAULT): DataType = {
+
+    def unsupportedType() = throw new IllegalArgumentException(s"Unsupported type: $dataType")
+
+    val pvGt4 = (protocolVersion.getCode >= ProtocolVersion.V4.getCode)
+
+    dataType match {
```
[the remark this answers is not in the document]
RussellSpitzer: Changed so DFColumnMapper uses this
ColumnType.fromDriverType(CassandraSourceutil.sparkSqlToDriverType)

And deprecated the createTable functions and noted you should use the catalog api instead.

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+    val createTableStart: OngoingPartitionKey = SchemaBuilder.createTable(namespace, table)
+
+    val createTableWithPk: CreateTable = partitionKeyNames.foldLeft(createTableStart) { (createTable, pkName) =>
+      val dataType = columnToType.getOrElse(pkName,
+        throw new CassandraCatalogException(s"$pkName was defined as a partition key but it does not exist in the schema ${schema.fieldNames.mkString(",")}"))
+      createTable.withPartitionKey(pkName, dataType).asInstanceOf[OngoingPartitionKey]
+    }.asInstanceOf[CreateTable]
+
+    val createTableWithClustering = clusteringKeyNames.foldLeft(createTableWithPk) { (createTable, ckName) =>
+      val dataType = columnToType.getOrElse(ckName._1,
```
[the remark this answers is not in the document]
RussellSpitzer: I'd rather leave this till Spark gets a Sorted keyword. Since Spark is doing the heavy lifting for us at the moment by actually looking for the matching column identifier for Partitioned By ...

I'll add a little fix for now if that is ok

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+    //There is an implicit for this but it's only accessible in org.apache.spark.sql.catalog (maybe we should use it)
+    val invalidPartitions = partitions.filter(_.name() != "identity")
+    if (invalidPartitions.nonEmpty) {
+      throw new UnsupportedOperationException(s"Cassandra Tables can only by partitioned based on direct references to columns, found: ${invalidPartitions.mkString(",")}")
+    }
+
+    val providedPartitionKeyNames = partitions.map(_.references().head.fieldNames().head)
+
+    val partitionKeys = if (providedPartitionKeyNames.nonEmpty) {
+      providedPartitionKeyNames
```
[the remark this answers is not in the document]
RussellSpitzer: I switched it to throwing an exception, I can change it to a log though if you think that's better.

jtgrabowski:
2.4 and 2.5 sound reasonable

jtgrabowski:
@RussellSpitzer let's merge this puppy :+1:

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraScanBuilder.scala
```diff
@@ -0,0 +1,375 @@
...
+  tableDef: TableDef,
+  catalogName: String,
+  options: CaseInsensitiveStringMap)
+
+  extends ScanBuilder
+    with SupportsPushDownFilters
+    with SupportsPushDownRequiredColumns
+    with Logging {
+
+  val consolidatedConf = consolidateConfs(session.sparkContext.getConf, session.conf.getAll, catalogName, tableDef.keyspaceName, options.asScala.toMap)
```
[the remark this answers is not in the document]
RussellSpitzer: I don't have a problem with "connector" being private, but I think the others are final fields and should be accessible, although currently they are only used in tests/catalyst rules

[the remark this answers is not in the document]
RussellSpitzer: Switched most things over to private

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraScanPartitionReaderFactory.scala
```diff
@@ -0,0 +1,160 @@
...
+  }
+}
+
+
+abstract class CassandraPartitionReaderBase
+  extends PartitionReader[InternalRow]
+    with SupportsReportStatistics
+    with Logging {
+
+  val connector: CassandraConnector
```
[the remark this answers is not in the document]
RussellSpitzer: Has to be protected because it's abstract, but switched

connector/src/main/scala/com/datastax/spark/connector/writer/TableWriter.scala
```diff
@@ -211,43 +202,106 @@ class TableWriter[T] private (
...
+    with Logging {
+
+  //Don't grab a connection or queryExecutor unless we are using this statement writer
+  lazy val session: CqlSession = connector.openSession()
+  val keyspaceName: String = tableDef.keyspaceName
+  val tableName: String = tableDef.tableName
+
+  lazy val queryExecutor = new QueryExecutor(session, writeConf.parallelismLevel, successHandler, failureHandler)
+
+  def batchRoutingKey(session: CqlSession)(bs: RichBoundStatementWrapper): Any = {
```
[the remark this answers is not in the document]
RussellSpitzer: it's unused, removing

connector/src/main/scala/com/datastax/spark/connector/datasource/CasssandraDriverDataWriterFactory.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  val unsafeRowWriterFactory = new InternalRowWriterFactory(inputSchema)
+
+  val columns = SomeColumns(inputSchema.fieldNames.map(name => ColumnName(name)): _*)
+
+  val writer =
+    TableWriter(connector, tableDef.keyspaceName, tableDef.tableName, columns, writeConf)(unsafeRowWriterFactory)
+      .getAsyncWriter()
+
+  override def write(record: InternalRow): Unit = writer.write(record)
```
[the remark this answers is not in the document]
RussellSpitzer: yeah one of our issues here is that the new DSV2 no longer provides a metrics context, the RDD it makes using DSV2 probably has some hooks but I figured it wasn't worth messing with them at the moment.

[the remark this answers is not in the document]
RussellSpitzer: Previously in the "compute" method of the RDD we grabbed the metrics context, we can no longer do that since we no longer are directly implementing "compute"

connector/src/main/scala/com/datastax/spark/connector/writer/TableWriter.scala
```diff
@@ -211,43 +202,106 @@ class TableWriter[T] private (
...
+  preparedStatement: PreparedStatement,
+  groupingBatchBuilderBase: GroupingBatchBuilderBase[T],
+  maybeRateLimit: RichStatement => Unit,
+  successHandler: Option[Handler[RichStatement]] = None,
+  failureHandler: Option[Handler[RichStatement]] = None)
+  extends Closeable
+    with Logging {
+
+  //Don't grab a connection or queryExecutor unless we are using this statement writer
+  lazy val session: CqlSession = connector.openSession()
```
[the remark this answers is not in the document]
RussellSpitzer: i think session and query executor are the only ones that really need to be private but i'll private all of them

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogNamespaceSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+    dropKeyspace(defaultKs)
+    waitForKeyspaceToExist(defaultKs, false)
+    val exception = intercept[CassandraCatalogException]{
+      spark.sql(s"CREATE DATABASE $defaultKs WITH DBPROPERTIES (class='foobar',replication_factor='5')")
+    }
+    exception.getMessage should include("Unknown keyspace replication strategy")
+  }
+
+  it should "throw a keyspace exists exception when creating a keyspace that already exists" in {
+    spark.sql(s"CREATE DATABASE IF NOT EXISTS $defaultKs WITH DBPROPERTIES (class='SimpleStrategy',replication_factor='5')")
```
absurdfarce, line 70: Nit: perhaps a waitForKeyspaceToExist(defaultKs, true) here to validate that the DB does in fact exist before trying the create?

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogNamespaceSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+    intercept[NamespaceAlreadyExistsException]{
+      spark.sql(s"CREATE DATABASE $defaultKs WITH DBPROPERTIES (class='SimpleStrategy',replication_factor='5')")
+    }
+  }
+
+  it should "explain Cassandra specific properties in the describe statement" in {
+    spark.sql(s"CREATE DATABASE IF NOT EXISTS $defaultKs WITH DBPROPERTIES (class='SimpleStrategy',replication_factor='5')")
```
absurdfarce, line 77: Nit: don't we want the dropKeyspace() + waitForKeyspaceToExist(false) sequence here as well?

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,472 @@
...
+  }
+
+  override def name(): String = s"Catalog $catalogName For Cassandra Cluster At $nameIdentifier " //TODO add identifier here
+
+  override def listNamespaces(): Array[Array[String]] = {
+    getMetadata(connector)
+      .getKeyspaces
+      .asScala
+      .map { case (name, _) => Array(name.asInternal()) }
+      .toArray
```
absurdfarce, line 97: Semi-nit: you get a bit nicer symmetry with listTables below via something like:

```scala
  override def listNamespaces(): Array[Array[String]] = {
    getMetadata(connector)
      .getKeyspaces
      .asScala
      .keys
      .map(ident => Array(DriverUtil.toName(ident)))
      .toArray
  }
```

You could also do something like the following which would allow you to pull out the map => Strings of the CqlIdentifiers in it's keys logic found in both functions:

```scala
  override def listNamespaces(): Array[Array[String]] = {
    getMetadata(connector)
      .getKeyspaces
      .asScala
      .keys
      .map(DriverUtil.toName(_))
      .map(Array(_))
      .toArray
  }
```

In theory this application lends itself to functional composition but a few quick attempts to make that work kept causing IntelliJ to yell at me... so I moved on. :)

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,472 @@
...
+    dropResult.wasApplied()
+  }
+
+  override def listTables(namespace: Array[String]): Array[Identifier] = {
+    getKeyspaceMeta(connector, namespace)
+      .getTables.asScala
+      .keys
+      .map { case tableName => Identifier.of(namespace, tableName.asInternal()) }
+      .toArray
+  }
```
absurdfarce, line 206: Semi-nit: I'm not sure the case statement is buying you much here... maybe this instead?

```scala
  override def listTables(namespace: Array[String]): Array[Identifier] = {
    getKeyspaceMeta(connector, namespace)
      .getTables.asScala
      .keys
      .map(DriverUtil.toName)
      .map(Identifier.of(namespace, _))
      .toArray
  }
```

See above re: extracting logic of a map of CqlIdentifier -> Any => Seq[String] from this function and listNamespaces() above.

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,431 @@
...
+    throw new UnsupportedOperationException("Cassandra does not support renaming tables")
+  }
+}
+
+object CassandraCatalog {
+
+  //Add asScala to JavaOptions to make working with the driver a little smoother
+  implicit class ScalaOptionConverter[T](javaOpt: Optional[T]) {
+    def asScala: Option[T] =
+      if (javaOpt.isPresent) Some(javaOpt.get) else None
```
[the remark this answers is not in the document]
absurdfarce: More a question than a comment: what's the thinking on something like the java8-compat lib for this kind of thing?  Clearly this example is pretty trivial, but let's assume it were more involved.  Is there a bias against including a third-party app for something like this or is it just a reflection of the simplicity of this specific case?  Or maybe... something else entirely?

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -0,0 +1,472 @@
...
+  def getTableMetaData(connector: CassandraConnector, ident: Identifier) = {
+    val namespace = ident.namespace
+    checkNamespace(namespace)
+    val tableMeta = getMetadata(connector)
+      .getKeyspace(fromInternal(namespace.head)).asScala
+      .getOrElse(throw namespaceMissing(getMetadata(connector), namespace))
+      .getTable(fromInternal(ident.name)).asScala
+      .getOrElse(throw tableMissing(getMetadata(connector), namespace, ident.name()))
+    tableMeta
+  }
```
absurdfarce, line 433: Semi-nit: perhaps marginally easier to do with the Java Optional API... at least that way you can avoid the creation of the intermediate Scala options just for a single op:

```scala
  //Table Support
  def getTableMetaData(connector: CassandraConnector, ident: Identifier) = {
    val namespace = ident.namespace
    checkNamespace(namespace)
    getMetadata(connector)
      .getKeyspace(fromInternal(namespace.head))
      .orElseThrow(() => namespaceMissing(getMetadata(connector), namespace))
      .getTable(fromInternal(ident.name))
      .orElseThrow(() => tableMissing(getMetadata(connector), namespace, ident.name()))
  }
```

Similar idea for getKeyspaceMeta() as well.

Oh, and perhaps we should add an explicit return type here (and to getKeyspaceMeta) as well?

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogNamespaceSpec.scala
RussellSpitzer: Sure but I think that would only cover the case where we throw an exception when the keyspace doesn't exist otherwise I think we are fine?

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogNamespaceSpec.scala
RussellSpitzer: sure, this would break if another keyspace with different properties was made, I wrote this before I had the NTS test in there too so all the keyspaces were the same anyway

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
RussellSpitzer: Simplified like the first example, I"m not super worried about getting the functional composition here since we are using the Java Driver's classes anyway.

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
RussellSpitzer: ah this was before I switched to scala 2.12, you can't write scala 2.11 lambdas and pass them to java functions. But since we are 2.12 we can do this now

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
RussellSpitzer: yeah I just popped out the case part, switched it to a single line map

RussellSpitzer:
I considered doing a Java8 Compat lib but we want to if at all possible
never include other dependencies, especially those which may be included in
an end user's application. Here I just had a bunch of places where we have
methods which take scala options and we have Java ones coming out of the
Driver or vice versa. Most of those we can actually eliminate now that we
can use Scala lambda's  in java8 anonymous functions (which we couldn't do
with scala 2.11)

doc/0_quick_start.md
```diff
@@ -35,59 +35,65 @@ This driver does not depend on the Cassandra server code.
...
-
 ### Loading up the Spark-Shell
 
-Run the `spark-shell` with the packages line for your version. To configure
+Run the `spark-shell` with the packages line for your version. This will include the connector
+and *all* of its dependencies on the Spark Class PathTo configure
 the default Spark Configuration pass key value pairs with `--conf`
 
     $SPARK_HOME/bin/spark-shell --conf spark.cassandra.connection.host=127.0.0.1 \
                                 --packages com.datastax.spark:spark-cassandra-connector_2.11:2.5.0
```
absurdfarce, line 45: This should now be an artifact with the "_2.12" name now, right?

RussellSpitzer:
Can of worms there :)

So the 2.X line of Spark is 2.11 (default) and 2.12 optional.

So we don't want to change these instructions until 3.0 is released, then
the artifact for default Spark 3.0 usage will be the 2.12 one

doc/0_quick_start.md
absurdfarce: Ah, nevermind, I missed the 2.5.0 part... thanks for the clarification @RussellSpitzer

PR #1252, opened 2020-06-22
SPARKC-599 exclude org.apache.tinkerpop

jtgrabowski:
As it is not used in SCC and it's transitive dependencies
cause ERRORs during spark-submit bootstrap (only when executed
with --packages argument).

### DIFF
--- a/project/Versions.scala
+++ b/project/Versions.scala
@@ -6,7 +6,7 @@ object Versions {
   val Paranamer       = "2.8"
   val ScalaLogging    = "3.5.0"
 
-  val DataStaxJavaDriver = "4.5.0"
+  val DataStaxJavaDriver = "4.7.2"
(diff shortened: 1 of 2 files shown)


project/Versions.scala
```diff
@@ -6,7 +6,7 @@ object Versions {
   val Paranamer       = "2.8"
   val ScalaLogging    = "3.5.0"
 
-  val DataStaxJavaDriver = "4.5.0"
+  val DataStaxJavaDriver = "4.7.2"
```
absurdfarce, line 9: Small note, mainly historical; if something less than 4.6.0 is used here another dependency (jnr-ffi) comes into play.  That lib was removed as an explicit dependency of the Java driver in 4.6.0; see JAVA-2662 for more details.

PR #1253, opened 2020-06-23
SPARKC-601 remove SCC - Cloud Envs dependency conflicts 

jtgrabowski:


build.sbt
```diff
@@ -121,3 +122,26 @@ lazy val driver = (project in file("driver"))
...
+  * It's a shaded version of the main artifact. It contains all of the dependent classes, some of them are shaded. */
+lazy val assembledJar = project
+  .enablePlugins(AssemblyPlugin)
+  .settings(assemblySettings)
+  .settings(
+    publish / skip := true,
+    assembly / assemblyOption := (assemblyOption in assembly).value.copy(includeScala = false),
+    assembly / assemblyShadeRules := {
+      Seq(
+        ShadeRule.rename("com.typesafe.config.**" -> s"shade.com.datastax.spark.connector.com.typesafe.config.@1").inAll
```
alexott, line 136: we also need to shade the `com.typesafe.scalalogging` that is used by `com.typesafe.config`

project/Dependencies.scala
```diff
@@ -75,11 +75,10 @@ object Dependencies
     val driverCore = "com.datastax.oss" % "java-driver-core-shaded" % DataStaxJavaDriver driverCoreExclude()
```
alexott, line 75: theoretically, because everything will be in assembly, we won't need to exclude `ffi` dependencies.

build.sbt
jtgrabowski: scala-logging was our dependency, I removed it in one of the previous commits

alexott: ah, ok... I messed it up :-)

project/Dependencies.scala
jtgrabowski: I would like to keep things consistent between the main artifact and the shaded one. Do you think it would be beneficial to have ffi in the shaded version?

alexott: My thought is following - we have problem with ffi when we're using connector with Spark directly via `--packages`. If instead we'll use shaded version with `--packages`, and leave the `ffi` dependencies for connector that will be used for building applications, then we won't have problem with Spark.

jtgrabowski:
@alexott could you take another look? As discussed we have two jars now, the main (unshaded, modular) and the assembly jar (shaded, fat). I adjusted build section of the docs. In the near future I'm planning to do a bigger, separate section explaining differences in detail.

README.md
```diff
@@ -44,8 +44,8 @@ development for the next connector release in progress.
 
 | Connector | Spark         | Cassandra | Cassandra Java Driver | Minimum Java Version | Supported Scala Versions |
 | --------- | ------------- | --------- | --------------------- | -------------------- | -----------------------  |
-| 3.0       | 3.0           | 2.1.5*, 2.2, 3.x, 4.0 | 4.5              | 8             | 2.11, 2.12               |
-| 2.5       | 2.4           | 2.1.5*, 2.2, 3.x, 4.0 | 4.5              | 8             | 2.11, 2.12
+| 3.0       | 3.0           | 2.1.5*, 2.2, 3.x, 4.0 | 4.7              | 8             | 2.11, 2.12               |
```
alexott, line 47: btw, Spark 3.0 supports only Scala 2.12, maybe will support Scala 2.13: https://issues.apache.org/jira/browse/SPARK-26132

doc/12_building_and_artifacts.md
```diff
@@ -3,84 +3,41 @@
...
+    sbt/sbt assembly
 
-    spark-submit --jars spark-cassandra-connector-assembly.jar
+A fat jar will be generated to:
 
+    spark-cassandra-connector/assembledJar/target/scala-{binary.version}/
 
-This driver is also compatible with Spark distribution provided in
-[DataStax Enterprise](https://www.datastax.com/products/datastax-enterprise).
+The jar contains Spark Cassandra Connector and it's dependencies. Some of the dependencies are  shaded to avoid
```
alexott, line 39: `it's` -> `its`

PR #1263, opened 2020-07-14
Sparkc 577 round three

absurdfarce:
# Description

Case classes TableDef, ColumnDef currently serve two distinct roles:

1. Metadata retrieval/access

Majority of operations leverage the first role above.  This PR begins the process of converting these usages to the types provided by the Java driver now that they're serializable.  The existing TableDef, ColumnDef etc. case classes are preserved as-is to support future table creation.

This PR reflects only the changes necessary to support conversion of one fairly high-level class (CassandraScanBuilder).  It certainly isn't complete and the changes to various method signatures broke other things in the code which will also need to be fixed.  The intent of this PR is to demonstrate the basic process in order to get agreement on this approach going forward.

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/connector/src/main/scala/com/datastax/spark/connector/ColumnSelector.scala
+++ b/connector/src/main/scala/com/datastax/spark/connector/ColumnSelector.scala
@@ -1,49 +1,70 @@
 package com.datastax.spark.connector
 
-import scala.language.implicitConversions
+import com.datastax.oss.driver.api.core.metadata.schema.TableMetadata
 
-import com.datastax.spark.connector.cql.TableDef
+import scala.language.implicitConversions
+import com.datastax.spark.connector.cql.{ColumnDef, StructDef, TableDef}
 
 sealed trait ColumnSelector {
   def aliases: Map[String, String]
   def selectFrom(table: TableDef): IndexedSeq[ColumnRef]
+  def selectFrom(table: TableMetadata): IndexedSeq[ColumnRef]
 }
 
 case object AllColumns extends ColumnSelector {
   override def aliases: Map[String, String] = Map.empty.withDefault(x => x)
   override def selectFrom(table: TableDef) =
     table.columns.map(_.ref)
+  override def selectFrom(table: TableMetadata) =
+    TableDef.columns(table).map(ColumnDef.toRef(_))
 }
 
 case object PartitionKeyColumns extends ColumnSelector {
   override def aliases: Map[String, String] = Map.empty.withDefault(x => x)
   override def selectFrom(table: TableDef) =
     table.partitionKey.map(_.ref).toIndexedSeq
+  override def selectFrom(table: TableMetadata) =
+    TableDef.partitionKey(table).map(ColumnDef.toRef(_)).toIndexedSeq
 }
 
 case object PrimaryKeyColumns extends ColumnSelector {
   override def aliases: Map[String, String] = Map.empty.withDefault(x => x)
   override def selectFrom(table: TableDef) =
     table.primaryKey.map(_.ref)
+  override def selectFrom(table: TableMetadata) =
+    TableDef.primaryKey(table).map(ColumnDef.toRef(_))
 }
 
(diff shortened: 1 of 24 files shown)


connector/src/main/scala/com/datastax/spark/connector/ColumnSelector.scala
```diff
@@ -1,49 +1,70 @@
...
   def selectFrom(table: TableDef): IndexedSeq[ColumnRef]
+  def selectFrom(table: TableMetadata): IndexedSeq[ColumnRef]
 }
 
 case object AllColumns extends ColumnSelector {
   override def aliases: Map[String, String] = Map.empty.withDefault(x => x)
   override def selectFrom(table: TableDef) =
     table.columns.map(_.ref)
+  override def selectFrom(table: TableMetadata) =
+    TableDef.columns(table).map(ColumnDef.toRef(_))
```
absurdfarce, line 19: Instances of this trait are used widely throughout the code base so in order to avoid broad breakage here I just overloaded selectFrom()

connector/src/main/scala/com/datastax/spark/connector/ColumnSelector.scala
```diff
@@ -1,49 +1,70 @@
...
+    case RowCountRef => Seq.empty //Filters RowCountRef from the column list
+    case other => Seq(other)
+  }
+
+  /** Compute the columns that are not present in the structure. */
+  private def missingColumns(table:StructDef, columnsToCheck: Seq[ColumnRef]): Seq[ColumnRef] =
+    for (c <- columnsToCheck if !table.columnByName.contains(c.columnName)) yield c
+
+  private def missingColumns(table:TableMetadata, columnsToCheck: Seq[ColumnRef]): Seq[ColumnRef] =
+    for (c <- columnsToCheck if !TableDef.containsColumn(c.columnName)(table)) yield c
```
absurdfarce, line 51: Formerly in StructDef.  Pulled out because it's only really used here so it seemed better to localize the functionality where it's needed.  If this becomes more broadly used it can always be moved back to StructDef

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraInJoinReaderFactory.scala
```diff
@@ -27,33 +28,32 @@ case class CassandraInJoinReaderFactory(
...
-  schema: StructType,
-  cqlQueryParts: ScanHelper.CqlQueryParts,
-  partition: InputPartition)
+abstract class CassandraBaseInJoinReader(connector: CassandraConnector,
+                                         tableMetadata: TableMetadata,
+                                         inClauses: Seq[In],
+                                         readConf: ReadConf,
+                                         schema: StructType,
+                                         cqlQueryParts: ScanHelper.CqlQueryParts,
+                                         partition: InputPartition)
```
absurdfarce, line 44: IntelliJ kept _insisting_ on reformatting arg lists in this way and nothing I did could convince it to stop.  This is a relatively minor thing that can be cleaned up later but my apologies for the noise while reviewing.

connector/src/main/scala/com/datastax/spark/connector/rdd/ClusteringOrder.scala
```diff
@@ -1,22 +1,32 @@
...
-import com.datastax.spark.connector.cql.TableDef
+import com.datastax.oss.driver.api.core.metadata.schema.TableMetadata
+import com.datastax.spark.connector.cql.{ColumnDef, TableDef}
 
 sealed trait ClusteringOrder extends Serializable {
   private[connector] def toCql(tableDef: TableDef): String
+  private[connector] def toCql(table: TableMetadata): String
 }
 
+// TODO: Keeping both TableDef and TableMetadata impls here to minimize intrusiveness of Java driver metadata changes
```
absurdfarce, line 11: Another case where overloading seemed to make sense, at least for now

PR #1265, opened 2020-09-01
SPARKC-615 temp workaround for inconsistent C* versioning

jtgrabowski:
remove this workaround once C* 4.0.0 is released

### DIFF
--- a/test-support/src/main/scala/com/datastax/spark/connector/ccm/mode/StandardModeExecutor.scala
+++ b/test-support/src/main/scala/com/datastax/spark/connector/ccm/mode/StandardModeExecutor.scala
@@ -44,8 +44,9 @@ private[mode] trait DefaultExecutor extends ClusterModeExecutor {
     * 4.0-beta1 which breaks versioning convention used in integration tests.
     */
   private def adjustCassandraBetaVersion(version: String): String = {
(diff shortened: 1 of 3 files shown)


test-support/src/main/scala/com/datastax/spark/connector/ccm/mode/StandardModeExecutor.scala
```diff
@@ -37,12 +37,25 @@ private[mode] trait DefaultExecutor extends ClusterModeExecutor {
     throw new IllegalStateException(s"Unable to complete function in $timeoutInSeconds: $hint")
   }
 
+  /**
+    * Remove this once C* 4.0.0 is released.
+    *
+    * This is a workaround that allows running it:test against 4.0.0-beta1. This version of C* is published as
+    * 4.0-beta1 which breaks versioning convention used in integration tests.
+    */
+  private def adjustCassandraBetaVersion(version: String): String = {
```
justinchuch, line 46: Question: just wondering, how many beta versions may need to adjust? Would there be beta2, beta3, etc.?
If so, would it be more flexible if we define a mapping in a config file, and load the mapping accordingly?

justinchuch (approved):
LGTM :+1:

PR #1267, opened 2020-09-03
SPARKC-598 return proper name of the catalog

jtgrabowski:
The CatalogPlugin requires the actual name in name() as it is
used in matching with spark.sql.catalog.<name> setting.

### DIFF
--- a/connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogNamespaceSpec.scala
+++ b/connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogNamespaceSpec.scala
@@ -7,7 +7,7 @@ import scala.collection.JavaConverters._
 class CassandraCatalogNamespaceSpec extends CassandraCatalogSpecBase {
 
   "A Cassandra Catalog Namespace Support" should "initialize successfully" in {
-    spark.sessionState.catalogManager.currentCatalog.name() should be(catalogName)
+    spark.sessionState.catalogManager.currentCatalog.name() should be(defaultCatalog)
   }
 
   it should "list available keyspaces" in {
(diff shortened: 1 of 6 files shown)


connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogNamespaceSpec.scala
```diff
@@ -7,7 +7,7 @@ import scala.collection.JavaConverters._
 class CassandraCatalogNamespaceSpec extends CassandraCatalogSpecBase {
 
   "A Cassandra Catalog Namespace Support" should "initialize successfully" in {
-    spark.sessionState.catalogManager.currentCatalog.name() should include("Catalog cassandra")
+    spark.sessionState.catalogManager.currentCatalog.name() should be(catalogName)
```
RussellSpitzer, line 10: I may or may not have ... added this test in a ton of other catalog specs

[info]   "cassandra" did not include substring "Catalog cassandra" (CassandraCatalogTableReadSpec.scala:22)
2292
You'll probably need to hit them all

RussellSpitzer: Other than that, Patch lgtm :)

RussellSpitzer (approved):
There are a few more tests with the "should init" that check the nam that need to be fixed, but 👍  other than that

connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala
```diff
@@ -440,7 +440,7 @@ object CassandraCatalog {
     * is used to quickly bail out if the namespace is not compatible with Cassandra
     */
   def checkNamespace(namespace: Array[String]): Unit = {
-    if (namespace.length != 1) throw new NoSuchNamespaceException(s"$OnlyOneNamespace : $namespace")
+    if (namespace.length != 1) throw new NoSuchNamespaceException(s"""$OnlyOneNamespace: ${namespace.mkString(".")}""")
```
RussellSpitzer, line 443: Unrelated?

jtgrabowski: yeah, the original version was not human readable

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogNamespaceSpec.scala
jtgrabowski: indeed! thanks, fixed.

PR #1272, opened 2020-09-15
SPARKC-619 restore PrefetchingResultSetIterator

jtgrabowski:
The iterator tries to prefetch the next result page. Once the
current page is exhausted, the next page (hopefully
materialized at this point) becomes the current page and a
fetch request is sent.

fetchSize argument was removed, the supplied statement should
have pageSize parameter set by the caller.

### DIFF
--- a/connector/src/main/scala/com/datastax/bdp/util/ScalaJavaUtil.scala
+++ b/connector/src/main/scala/com/datastax/bdp/util/ScalaJavaUtil.scala
@@ -7,10 +7,11 @@
 package com.datastax.bdp.util
 
 import java.time.{Duration => JavaDuration}
-import java.util.concurrent.Callable
+import java.util.concurrent.{Callable, CompletionStage}
 import java.util.function
-import java.util.function.{Consumer, Predicate, Supplier}
+import java.util.function.{BiConsumer, Consumer, Predicate, Supplier}
 
+import scala.concurrent.{ExecutionContext, Future, Promise}
 import scala.concurrent.duration.{Duration => ScalaDuration}
 import scala.language.implicitConversions
 
@@ -45,4 +46,19 @@ object ScalaJavaUtil {
   }
 
   def asScalaFunction[T, R](f: java.util.function.Function[T, R]): T => R = x => f(x)
+
+  def asScalaFuture[T](completionStage: CompletionStage[T])
+                      (implicit context: ExecutionContext): Future[T] = {
+    val promise = Promise[T]()
+    completionStage.whenCompleteAsync(new BiConsumer[T, java.lang.Throwable] {
+      override def accept(t: T, throwable: Throwable): Unit = {
+        if (throwable == null)
+          promise.success(t)
+        else
+          promise.failure(throwable)
+
+      }
+    })
+    promise.future
+  }
 }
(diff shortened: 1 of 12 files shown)


connector/src/main/scala/com/datastax/spark/connector/util/Threads.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import scala.concurrent.ExecutionContext
+
+object Threads {
+
+  implicit val BlockingIOExecutionContext = {
+    val threadFactory = new ThreadFactoryBuilder()
+      .setDaemon(true)
+      .setNameFormat("spark-cassandra-connector-io" + "%d")
+      .build
```
jacek-lewandowski, line 15: perhaps it would be good to add exception handler which logs failures

connector/src/main/scala/com/datastax/spark/connector/cql/Scanner.scala
```diff
@@ -35,21 +39,25 @@ class DefaultScanner (
...
-          rateLimiter.maybeSleep(getRowBinarySize(row))
-          row
-        }
-      case None =>
-        prefetchingIterator
-    }
+    import com.datastax.spark.connector.util.Threads.BlockingIOExecutionContext
 
-    ScanResult(rateLimitingIterator, columnMetaData)
+    val rs = session.executeAsync(maybeExecutingAs(statement, readConf.executeAs))
```
jacek-lewandowski, line 44: do you swallow fetch size on purpose?

jtgrabowski: yes, now the caller is responsible for setting the page size on the statement. In most cases the parameter was already set, I had to add it in one or two places.

connector/src/main/scala/com/datastax/spark/connector/util/Threads.scala
jtgrabowski: done

connector/src/main/scala/com/datastax/spark/connector/cql/Scanner.scala
```diff
@@ -35,21 +39,25 @@ class DefaultScanner (
...
+          prefetchingIterator.map { row =>
+            rateLimiter.maybeSleep(getRowBinarySize(row))
+            row
+          }
+        case None =>
+          prefetchingIterator
+      }
+      ScanResult(rateLimitingIterator, columnMetaData)
+    }
+    Await.result(scanResult, Duration.Inf)
```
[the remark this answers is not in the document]
jtgrabowski: yes, the driver is responsible for timeouting the query: 
https://github.com/datastax/spark-cassandra-connector/blob/9b1609c94c37032bf0337cac113df459eb73ee00/connector/src/main/scala/com/datastax/spark/connector/cql/CassandraConnectionFactory.scala#L57

connector/src/main/scala/com/datastax/bdp/util/ScalaJavaUtil.scala
```diff
@@ -45,4 +46,19 @@ object ScalaJavaUtil {
   }
 
   def asScalaFunction[T, R](f: java.util.function.Function[T, R]): T => R = x => f(x)
+
+  def asScalaFuture[T](completionStage: CompletionStage[T])
+                      (implicit context: ExecutionContext): Future[T] = {
```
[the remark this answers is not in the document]
jtgrabowski: my intention was to run the callback in the `context`. Fixed.

jacek-lewandowski (approved):

PR #1275, opened 2020-09-18
SPARKC-620 closed classloader workaround

jtgrabowski:
During shutdown hive-exec (used in spark-sql) closes context class
loader that was used to load all the SCC classes.
This happens before shutdown hooks invocation. It results in errors
when an additional SCC/Java Driver/Netty class needs to be loaded
during shutdown hook execution.

Unfortunately the class loading errors cause our hook to hang forever
(we use the hook to close all the remaining Driver sessions and
session.close() invocation simply doesn't ever return).

Since there is no way to avoid class loader closing, we need to fix
things on our end.

This PR introduces a workaround for this problem. An attempt to load
a marker class is made just before the shutdown hook invocation. In
case of an error the process exits without calling `close` on the
remaining sessions.

The proper solution would require removing the hook completely.


RussellSpitzer:
As we discussed on slack, I think we should do the try catch in the shutdown hook itself rather than in the general class, then we can hopefully avoid the special case handling

RussellSpitzer (approved):
I think this is a fine "for now" solution but we really should figure out a better way to handle this. I wish we could just do a try with time limit or something ... but even that's not very satisfying

jtgrabowski:
I created this https://datastax-oss.atlassian.net/browse/SPARKC-622. The initial idea would require a contribution to Spark - a close method in `CatalogPlugin` interface.

PR #1279, opened 2020-10-07
SPARKC-613 fix direct join for spark sql

jtgrabowski:
Expression strict matching used in CDJExec and CDJStrategy failed when
one of the expressions contained a qualifier. This failed Direct Join
recognition.
Since we care only about expression names and types, this commit
introduces a loosened expression comparison.


jacek-lewandowski (approved):

RussellSpitzer (approved):
Thanks for back porting this! I think this is the right thing to do

jacek-lewandowski:
+1

jacek-lewandowski (approved):

PR #1282, opened 2020-10-27
 SPARKC-627 fix join rowsPerSec throttling 

jtgrabowski:
Before this change each maybeSleep was invoked on a new instance
of the RateLimiter which broke throttling.

### DIFF
--- a/connector/src/it/scala/org/apache/spark/sql/cassandra/execution/JoinThrottlingSpec.scala
+++ b/connector/src/it/scala/org/apache/spark/sql/cassandra/execution/JoinThrottlingSpec.scala
@@ -1,0 +1,81 @@
+package org.apache.spark.sql.cassandra.execution
+
+import com.datastax.spark.connector.SparkCassandraITFlatSpecBase
+import com.datastax.spark.connector.cluster.DefaultCluster
+import com.datastax.spark.connector.cql.CassandraConnector
+import com.datastax.spark.connector.embedded.SparkTemplate
+import com.datastax.spark.connector.rdd.ReadConf
+import org.apache.spark.sql.cassandra.CassandraSourceRelation.DirectJoinSettingParam
+import org.apache.spark.sql.cassandra._
+import org.scalatest.concurrent.Eventually
+
+import scala.concurrent.Future
+
+class JoinThrottlingSpec extends SparkCassandraITFlatSpecBase with DefaultCluster with Eventually {
+
+  override lazy val conn = CassandraConnector(defaultConf)
+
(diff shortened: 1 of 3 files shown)


connector/src/it/scala/org/apache/spark/sql/cassandra/execution/JoinThrottlingSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+      .option(ReadConf.ReadsPerSecParam.name, rowsPerSecondPerCore)
+      .load()
+    val join = left.join(right, left("k") === right("id"))
+
+    val startMillis = System.currentTimeMillis()
+    join.write.format("org.apache.spark.sql.cassandra")
+      .options(Map("keyspace" -> ks, "table" -> "kvtarget"))
+      .mode("append")
+      .save()
+    val durationSeconds = (System.currentTimeMillis() - startMillis).toInt / 1000
```
RussellSpitzer, line 59: Seems like a long test to check this, but logic seems sound to me

RussellSpitzer: Maybe put an assert in here that minimalDurationSeconds > 2 or 3 ? Just to make sure the test isn't doing nothing?

RussellSpitzer: Maybe just loop over a few thresholds and make sure each is slower than the previous? I dunno, ignore me if I'm just paranoid

jtgrabowski: Thanks for the review :) It takes ~15 seconds on my laptop. We could lower this if this too long.

RussellSpitzer: My worry is that, what if unthrottled it takes 15 seconds (or greater) and so the test passes just because the CI environment is slow, not because of the throttle.

jtgrabowski: I see, without throttling the test is an order of magnitude slower, around 2 seconds.

RussellSpitzer: But what about when I run the tests on my raspberry pi :) If you are sure it will be faster than I'm :+1:

PR #1286, opened 2020-11-03
SPARKC-621 SAI support

jtgrabowski:
SAI predicates pushdown rules:

  1. IN clause for a partition key part prevents SAI push downs
  2. OR clause prevents any push down
  3. At most one equality is pushed
  4. Only equality is pushed if equality and range are defined
  5. Supported predicates:
   - numeric types: >, <, >=, <=, =
   - text types: =
   - set and list types: contains (non-frozen collections), = (frozen collections)
   - map type: contains key, contains value, key lookup result equality (map(key) == value), = (frozen collections)
 
   NOTE that collection types support is going to be introduces in SPARKC-630


connector/src/main/scala/org/apache/spark/sql/cassandra/DsePredicateRules.scala
```diff
@@ -53,6 +65,79 @@ object DsePredicateRules extends CassandraPredicateRules {
...
+    *  - map type: contains key, contains value, key lookup result equality (map(key) == value), = (frozen collections)
+    *
+    *  NOTE that collection types support is going to be introduces in SPARKC-630
+    */
+  private def storageAttachedIndexPredicateRules(predicates: AnalyzedPredicates, table: TableDef): Seq[PredicateRule] = {
+    val partitionKeyColumns = table.partitionKey.map(_.columnName).toSet
+    val inClauseInPartitionKey = predicates.handledByCassandra
+      .find(f => FilterOps.isInPredicate(f) && f.references.exists(partitionKeyColumns.contains))
+
+    if (inClauseInPartitionKey.isDefined) {
```
jacek-lewandowski, line 127: i might have misunderstood that but can we do push down sai predicates when all partition key columns are defined in `in` clause?

jacek-lewandowski (approved):
great job, as usual 👍 
left one question, other than that +1

connector/src/main/scala/org/apache/spark/sql/cassandra/DsePredicateRules.scala
jtgrabowski: no we can't, in this case we get
```
InvalidRequest: Error from server: code=2200 [Invalid query] message="Select on indexed columns and with IN clause for the PRIMARY KEY are not supported"
```

jtgrabowski:
test this please

jtgrabowski:
test this please

PR #1287, opened 2020-11-09
SPARKC-629 fix invalid clustering key pushdown

jtgrabowski:
Before this change clustering keys were pushed down without
considering partition key pushdowns. This enabled a scenario
where a clustering key was pushed down and the partition key
pushdown was missing. This kind of query ends with a timeout.
It was not rejected by C* only because we add ALLOW FILTERING
to our queries.

This PR introduces pk check to ck pushdowns. CK predicates
are pushed down only if pk pushdowns are defined.


jtgrabowski:
After giving this a bit more thought, I think we should leave this as is. C* does better job on filtering wide partitions than Spark side filtering would do.

PR #1288, opened 2020-11-27
fix SPARKC-634

wForget:
[SPARKC-634](https://datastax-oss.atlassian.net/browse/SPARKC-634) The AsyncStatementWriter.wirte method should check for exceptions after each execution

### DIFF
--- a/connector/src/main/scala/com/datastax/spark/connector/writer/TableWriter.scala
+++ b/connector/src/main/scala/com/datastax/spark/connector/writer/TableWriter.scala
@@ -264,6 +264,7 @@ case class AsyncStatementWriter[T](
     groupingBatchBuilderBase.batchRecord(record).foreach{ stmt =>
       queryExecutor.executeAsync(stmt.executeAs(writeConf.executeAs))
       maybeRateLimit(stmt)
+      assertExecutorException(queryExecutor)
     }
   }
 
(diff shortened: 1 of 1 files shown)


dheerajk (approved):

RussellSpitzer:
So the goal here is that we check for possible failed executions of previous requests every time we make a new request? This means we have to touch a volatile on every request which sounds expensive to me. 

I know the current logic just waits for either, the whole thing to fail IE a query.execute just fails all by itself, or we wait for the till the entire task is submitted, and check for any failures from in flight async requests.

Is the idea here we would just fail faster since each request would be checking to see if any previous requests failed?

wForget:
> So the goal here is that we check for possible failed executions of previous requests every time we make a new request? This means we have to touch a volatile on every request which sounds expensive to me.
> 
> I know the current logic just waits for either, the whole thing to fail IE a query.execute just fails all by itself, or we wait for the till the entire task is submitted, and check for any failures from in flight async requests.
> 
> Is the idea here we would just fail faster since each request would be checking to see if any previous requests failed?

It is my understanding that the volatile operation only lock during write operations, and a large number of reads may not cause too much performance loss.
The current logic is to check the exception only after all operations have been executed. If an exception occurs, it will still be executed for a long time.

RussellSpitzer:
Ah so the goal is just to reduce the time to failure with the hope that if a future failed for another query, we will catch it when attempting to execute a subsequent query. That sounds pretty reasonable to me

jtgrabowski:
This is a reasonable change, would you mind preparing a PR for b2.5?

wForget:
> This is a reasonable change, would you mind preparing a PR for b2.5?

okay, thank you. Do I need to close this PR?

PR #1291, opened 2020-12-10
fix SPARKC-634 for b2.5

wForget:
[SPARKC-634](https://datastax-oss.atlassian.net/browse/SPARKC-634) The TableWriter.writeInternal Method should check for exceptions after each execution of queryExecutor

### DIFF
--- a/connector/src/main/scala/com/datastax/spark/connector/writer/TableWriter.scala
+++ b/connector/src/main/scala/com/datastax/spark/connector/writer/TableWriter.scala
@@ -229,26 +229,31 @@ class TableWriter[T] private (
       for (stmtToWrite <- batchBuilder) {
         queryExecutor.executeAsync(stmtToWrite.executeAs(writeConf.executeAs))
         maybeRateLimit(stmtToWrite)
+        assertExecutorException(queryExecutor)
       }
 
       queryExecutor.waitForCurrentlyExecutingTasks()
-
-      queryExecutor.getLatestException().map {
-        case exception =>
-          throw new IOException(
-            s"""Failed to write statements to $keyspaceName.$tableName. The
-               |latest exception was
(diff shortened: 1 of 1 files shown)


connector/src/main/scala/com/datastax/spark/connector/writer/TableWriter.scala
```diff
@@ -229,26 +229,31 @@ class TableWriter[T] private (
       for (stmtToWrite <- batchBuilder) {
         queryExecutor.executeAsync(stmtToWrite.executeAs(writeConf.executeAs))
         maybeRateLimit(stmtToWrite)
+        assertExecutorException(queryExecutor)
```
jtgrabowski, line 232: hi @wForget , thank you for your contribution! I finally had some time to look into this. I agree with the idea to failing fast here but this implementation is not deterministic. In the worst case scenario all `assertExecutorException` invocations in this line will fire before any of the asynchronous statement execution ends. Would you be willing to implement the fail fast approach in the QueryProcessor? (each task completion checks for errors and breaks further execution if an error is found)

wForget: hi @jtgrabowski, thank you for your reply! Due to the control of maximum concurrency, the submission speed of all tasks may not be so fast. So I think the judgment here is still needed.
In the executeAsync method, I don't think we should judge the failure of the last task. Is it possible to break if a failure is found in the for loop of the waitForCurrentlyExecutingTasks method?

jtgrabowski: Yeah, I agree with the general idea. We need to break the execution as soon as the first error appears. We can do that in QueryExecutor

PR #1295, opened 2021-01-07
Add initial docs for sparkR with Cassandra Connector

harrison-tarr:
# Description
Documentation update for SparkR compatibility

It was hard to get started without an intimate knowledge of the SparkR API. These instructions should make it easier to pick it up and start experimenting.

It is a new documentation file with a unique integer number at the beginning of the file name, following the pattern of the others.

# How Has This Been Tested?

Manually verified every step that was added in a sparkR shell running against a local Cassandra.

# Checklist:

- [ ] I have a ticket in the [OSS JIRA](https://datastax-oss.atlassian.net/projects/SPARKC) (possibly n/a, i don't want to clutter the jira with doc tickets - happy to add one if that is the right approach)
- [X] I have performed a self-review of my own code
- [ ] Locally all tests pass (make sure tests fail without your patch) (i dont' think this would need to be tested?)

### DIFF
--- a/doc/17_quick_start_with_sparkr.md
+++ b/doc/17_quick_start_with_sparkr.md
@@ -1,0 +1,87 @@
+# Documentation
+
+## 5-minute quick start guide for Spark 3.0
+
+In this tutorial, you'll learn how to setup a very simple Spark application for reading and writing data from/to Cassandra.
+Before you start, you need to have basic knowledge of Apache Cassandra and Apache Spark.
+Refer to [Datastax](https://docs.datastax.com/en/) and [Cassandra documentation](https://cassandra.apache.org/doc/latest/getting_started/index.html)
+and [Spark documentation](https://spark.apache.org/docs/latest/). 
+
+### Prerequisites
+
+Install and launch a Cassandra cluster
+
+### Loading up the Spark-Shell
+
+Run `sparkr` with the packages line for your version. This will include the connector
+and *all* of its dependencies on the Spark Class PathTo configure
+the default Spark Configuration pass key value pairs with `--conf`
+
+    $SPARK_HOME/bin/sparkr --conf spark.cassandra.connection.host=127.0.0.1 \
+                                --packages com.datastax.spark:spark-cassandra-connector_2.12:3.0.0
+                                --conf spark.sql.extensions=com.datastax.spark.connector.CassandraSparkExtensions
+
+This command would set the Spark Cassandra Connector parameter 
+`spark.cassandra.connection.host` to `127.0.0.1`. Change this
+to the address of one of the nodes in your Cassandra cluster.
+
+The extensions configuration option enables Cassandra Specific Catalyst
+optimizations and functions.
+ 
+Create a Catalog Reference to your Cassandra Cluster
+
+```R
+sparkR.session(sparkConfig=list(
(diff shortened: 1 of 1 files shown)


harrison-tarr:
this doc update is accurate for all the first steps. I wasn't sure about some of the subsequent DSL examples that work in Scala. Does anyone know if there is a corresponding SparkR DSL that can be used to make those same calls or is it all `sql()`?

harrison-tarr:
For reference, this was discussed in the google group yesterday/today (Jan 6, 7 2021) https://groups.google.com/u/1/a/lists.datastax.com/g/spark-connector-user/c/KWmLrs1nvBM/m/5zj2qhC-CAAJ

alexott:
Here is more: https://stackoverflow.com/questions/41601715/integration-of-spark2-0-and-cassandra-using-r/57107100#57107100

harrison-tarr:
Hi again Alex,

Thanks for the additional information, it's really helpful. I had seen the
'read.df' syntax and couldn't figure out how to make it work against the
Cassandra Connector so I had been using 'sql' calls, but now I have the
option of both!

I have a new question about an error I'm seeing when joining two
SparkDataFrames after running 'regexp_replace'. I have two dataframes that
I want to join on the ids, but one of the ids has the string "#hidden"
appended. I try to run the 'regexp_replace(dataframeA$column, "#hidden",
"") and then a join(dataframeA, dataframeB,
dataframeA$column==dataframeB$id)

When I try to run any subsequent commands, like head() on the result, it
fails with the following error:
Error in handleErrors(returnStatus, conn) :
  scala.MatchError: regexp_replace(id#175, #hidden,  ) AS id_fixed#187 (of
class org.apache.spark.sql.catalyst.expressions.Alias)
at
org.apache.spark.sql.cassandra.execution.CassandraDirectJoinStrategy$.$anonfun$aliasMap$1(CassandraDirectJoinStrategy.scala:315)
at
scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
at scala.collection.TraversableLike.map(TraversableLike.scala:238)
at scala.collection.TraversableLike.map$(TraversableLike.scala:231)
at scala.collection.AbstractTraversable.map(Traversable.scala:108)
at
org.apache.spark.sql.cassandra.execution.CassandraDirectJoinStrategy$.aliasMap(CassandraDirectJoinStrategy.scala:315)
at
org.apache.spark.sql.cassandra.execution.CassandraDirectJoinStrategy$.allPartitionKeysAreJoin

Do I need to do something special to execute a 'regexp_replace'? When I
execute the join without the 'regexp_replace', I can do a head() on the
result, but obviously, the dataframe is empty.

alexott:
please post the bigger, snippet of the code. Also, this is more a question for StackOverflow, or mailing list, not in the context of the pull request...
I suspect that you need to alias the result of `regex_replace` to give it a name matching to the column in database.

harrison-tarr:
oops, i thought i was replying to the mailing list thread! my mistake. I will continue the discussion there

PR #1302, opened 2021-02-25
SPARKC-642 repartitionByCassandraReplica relocates data to the local node only

yohannrub:
# Description

`repartitionByCassandraReplica` relocates data to the local(-host) node only, whereas it should relocate data over all the nodes of the local DC instead.

When running `repartitionByCassandraReplica` on a machine where:
- a local node exists, then it relocates the entire data to this single node.
- no local node exist (e.g. with nodes being on remote machines, or with a containerized node with an isolated IP on the local machine), then it always returns an empty RDD without throwing any exception.

The problem is caused by the [reconfiguration](https://github.com/datastax/spark-cassandra-connector/commit/3f891f8c25283eb5fb19eac4407d847d5c8d8d94) of `CassandraConnectorConf` that handles new non-IP-based `ContactInfo`, as part of the v2.5.0 pre-release.

In order to determine the local DC nodes, the nodes retrieved from the cassandra session are [filtered](https://github.com/datastax/spark-cassandra-connector/commit/3f891f8c25283eb5fb19eac4407d847d5c8d8d94#diff-ea855a77ea29f5a80329c44d5823660a3f600bd5b4abaf63f9dc3832e1db062fR73) to select only LOCAL-distance nodes. This is, in effect, selecting the localhost node only, not all the nodes of the local DC as would be expected.

The confusion may come from the fact that:
- according to the default load-balancing policy provided by the cassandra java driver, `LOCAL` nodes are nodes in the local DC, `REMOTE` nodes are nodes in other DCs ([BasicLoadBalancingPolicy](https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/BasicLoadBalancingPolicy.java#L401))
- whereas according to the load-balancing policy used by the spark cassandra connector, the `LOCAL` node is the localhost node, `REMOTE` nodes are other nodes in the local DC, all nodes in other DCs are `IGNORED` ([LocalNodeFirstLoadBalancingPolicy](https://github.com/datastax/spark-cassandra-connector/blob/master/driver/src/main/scala/com/datastax/spark/connector/cql/LocalNodeFirstLoadBalancingPolicy.scala#L49))

This difference may be confusing, but is acceptable considering that it is up to each policy implementation to determine how it handles the `NodeDistance` it assigns to nodes.

This PR fixes the issue by selecting non-`IGNORED` (i.e. `LOCAL` + `REMOTE`) nodes as expected, instead of `LOCAL` node only.

Moreover, it puts the determination of the local DC nodes back into the existing method `dataCenterNodes()` of class `CassandraConnector` (it is in fact how it was designed before the problematical commit). This refactoring has two benefits:
- it avoids the repeated calls to the method `LocalNodeFirstLoadBalancingPolicy.determineDataCenter()`, in order to determine the local DC, each time the method `dataCenterNodes()` is called
- it throws an `ConnectorConfigurationException` when no node in the local DC is found

Lastly, it cleans up an unused piece of code, which was instantiating an unused `ReplicaLocator` in the RDD function `repartitionByCassandraReplica` (it seems to be a legacy code copied from the RDD function `keyByCassandraReplica`). `repartitionByCassandraReplica` uses a `ReplicaPartitioner` instead, which is referencing the (currently wrong) set of local DC nodes as described previously.

Fixes: [SPARKC-642](https://datastax-oss.atlassian.net/browse/SPARKC-642)

# How Has This Been Tested?

I discovered this bug by running `repartitionByCassandraReplica` on my local machine with a containerized cassandra node, and noticing that an empty RDD is always returned as described above.

Note that integration tests that are testing method `repartitionByCassandraReplica` in `RDDSpec` currently only check the number of output partitions on the basis of the number of nodes determined by `CassandraConnector`, but they do not check that this number of nodes is actually the one expected, which is why the tests pass, but with a wrong number of nodes / partitions.

Designing suitable integration tests, in order to correctly check this, would probably require digging into some CCM setups, which is beyond my expertise.
Any contribution on this is welcome.

# Checklist:

### DIFF
--- a/connector/src/it/scala/com/datastax/spark/connector/cluster/Fixtures.scala
+++ b/connector/src/it/scala/com/datastax/spark/connector/cluster/Fixtures.scala
@@ -187,6 +187,15 @@ trait TwoClustersWithOneNode extends Fixture {
   override def cluster(c: Int): Cluster = ClusterHolder.get(this)(c)
 }
 
+/** A fixture that bootstraps on single cluster with two nodes. */
+trait TwoNodeCluster extends SingleClusterFixture {
+
+  private[cluster] final override val configs: Seq[CcmConfig] = Seq(defaultConfig.copy(nodes = Seq(1, 2)))
+
+  private[cluster] override def connectionParameters(address: InetSocketAddress): Map[String, String] =
+    DefaultCluster.defaultConnectionParameters(address)
+}
+
 trait CETCluster extends DefaultCluster
 
 trait CSTCluster extends DefaultCluster
--- a/connector/src/it/scala/com/datastax/spark/connector/rdd/ReplicaRepartitionedCassandraRDDSpec.scala
+++ b/connector/src/it/scala/com/datastax/spark/connector/rdd/ReplicaRepartitionedCassandraRDDSpec.scala
@@ -1,0 +1,85 @@
+package com.datastax.spark.connector.rdd
+
+import java.lang.{Long => JLong}
+
+import com.datastax.spark.connector._
+import com.datastax.spark.connector.cluster.TwoNodeCluster
+import com.datastax.spark.connector.cql.CassandraConnector
+import com.datastax.spark.connector.rdd.partitioner.EndpointPartition
+
+import scala.concurrent.Future
+
+class ReplicaRepartitionedCassandraRDDSpec extends SparkCassandraITFlatSpecBase with TwoNodeCluster {
+
+  override lazy val conn = CassandraConnector(defaultConf)
+  val tableName = "key_value"
+  val keys = 0 to 200
+  val total = 0 to 10000
+
+  override def beforeClass {
+    conn.withSessionDo { session =>
+      createKeyspace(session)
+      val startTime = System.currentTimeMillis()
+
+      val executor = getExecutor(session)
+
+      awaitAll(
+        Future {
+          session.execute(
+            s"""
+               |CREATE TABLE $ks.$tableName (
+               |  key INT,
+               |  group BIGINT,
+               |  value TEXT,
+               |  PRIMARY KEY (key, group)
+               |)""".stripMargin)
+          val ps = session
+            .prepare(s"""INSERT INTO $ks.$tableName (key, group, value) VALUES (?, ?, ?)""")
+          awaitAll((for (value <- total) yield
+            executor.executeAsync(ps.bind(value: Integer, (value * 100).toLong: JLong, value.toString))): _*)
+        }
+      )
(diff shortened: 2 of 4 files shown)


jtgrabowski:
Hi! Thanks for the contribution, the changes look good to me.
Could you please add an integration test for this? ccm knowledge is not needed, you could add a new fixture like this
```
/** todo: doc */
trait TwoNodeCluster extends SingleClusterFixture {

  private[cluster] final override val configs: Seq[CcmConfig] = Seq(defaultConfig.copy(nodes = Seq(1, 2)))

  private[cluster] override def connectionParameters(address: InetSocketAddress): Map[String, String] =
    DefaultCluster.defaultConnectionParameters(address)
}
```
And then use it in a new integration test (instead of DefaultCluster).

yohannrub:
Hi @jtgrabowski ! Thanks for your reply !
I was indeed able to add an integration test with the new fixture template that you suggested.
It explicitly checks the number of nodes on which the repartition is working, in addition to the number of partitions.

However, note that it does not quite address the spotted issue, because I can notice that the two nodes created by CCM with this fixture are both `LOCAL` (i.e. on a localhost IP), in a same datacenter `datacenter1`.
Ideally, we would like to test the repartition on a setup with some `LOCAL` (localhost IP), `REMOTE` (non-localhost IP, in local datacenter) and `IGNORED` (non-local datacenter) nodes.

I don't think this would be straightforward to setup with CCM and the test-support classes, wouldn't it ?

jtgrabowski:
Thanks! The test looks good and I'm ok with LOCAL thing. Have you tried running it for DSE 6.8.9? It fails on Travis, other configs pass.

yohannrub:
I can reproduce the test failure locally with DSE 6.8.9

Seems like something is wrong with the node config or an internal issue in DSE, due to the addition of the second node.
Note this warning before the bunch of error logs:
```
WARN 13:27:46,497 [T11] com.datastax.oss.driver.internal.core.metadata.DefaultTopologyMonitor (DefaultTopologyMonitor.java:553) - [s0] Found invalid row in system.peers for peer: /127.11.0.2. This is likely a gossip or snitch issue, this node will be ignored.
```

I didn't even manage to make it skip the test for DSE, because the failure occurs at CCM bridge run time, before any actual test code is run...

What is your recommendation ?

jtgrabowski:
I need to investigate, will get back to you.

jtgrabowski:
the test is fixed by #1305

yohannrub:
Thanks !
Glad to see that this will soon be included in release 2.5.2 / 3.0.1

PR #1308, opened 2021-04-08
SPARKC-647 UDTValue performance fix - reuse CassandraRowMetadata instead of building it for each row

gowa:
# Description

Slow UDTValue related work on saving to Cassandra:
saveToCassandra for 100K rows (about 100Mb) with cores = 1 (1 Spark thread) took 45 seconds
the UDT has about 50 fields.
spark.cassandra.output.batch.size.bytes = 63000
spark.cassandra.output.concurrent.writes = 160
scala case classes were used for both row and a UDT field.
the same results for Java Beans.

asyncprofiler showed that the recalculation of 'metadata' (CassandraRowMetadata) in com.datastax.spark.connector.UDTValue for each row is burning CPU, because metadata is not reused.

At the same time there is a (unused?) class com.datastax.spark.connector.japi.UDTValue that receives CassandraRowMetadata as a constructor parameter.

I decided to align the approaches in those UDTValue classes. 

# How Has This Been Tested?

Unit tests were not executed. The jar (unshaded) was built and profiled on the env. For the same test the timing improved significantly - 9-10 sec instead of 45 sec.

# Checklist:

- [x] I have a ticket in the [OSS JIRA](https://datastax-oss.atlassian.net/browse/SPARKC-647) - SPARKC-647

### DIFF
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala
@@ -6,12 +6,10 @@ import scala.reflect.runtime.universe._
 import com.datastax.driver.core.{UDTValue => DriverUDTValue}
 import com.datastax.spark.connector.types.NullableTypeConverter
 
-final case class UDTValue(columnNames: IndexedSeq[String], columnValues: IndexedSeq[AnyRef])
+final case class UDTValue(metaData: CassandraRowMetadata, columnValues: IndexedSeq[AnyRef])
   extends ScalaGettableData {
   override def productArity: Int = columnValues.size
   override def productElement(i: Int) = columnValues(i)
-
-  override def metaData = CassandraRowMetadata.fromColumnNames(columnNames)
 }
 
 object UDTValue {
@@ -19,11 +17,11 @@ object UDTValue {
   def fromJavaDriverUDTValue(value: DriverUDTValue): UDTValue = {
     val fields = value.getType.getFieldNames.toIndexedSeq
     val values = fields.map(GettableData.get(value, _))
-    UDTValue(fields, values)
+    UDTValue(CassandraRowMetadata.fromColumnNames(fields), values)
   }
 
   def fromMap(map: Map[String, Any]): UDTValue =
-    new UDTValue(map.keys.toIndexedSeq, map.values.map(_.asInstanceOf[AnyRef]).toIndexedSeq)
+    new UDTValue(CassandraRowMetadata.fromColumnNames(map.keys.toIndexedSeq), map.values.map(_.asInstanceOf[AnyRef]).toIndexedSeq)
 
   val TypeTag = implicitly[TypeTag[UDTValue]]
   val Symbol = typeOf[UDTValue].asInstanceOf[TypeRef].sym
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala
@@ -6,23 +6,19 @@ import scala.reflect.runtime.universe._
 import com.datastax.driver.core.{UDTValue => DriverUDTValue}
 import com.datastax.spark.connector.types.NullableTypeConverter
 
-final case class UDTValue(columnNames: IndexedSeq[String], columnValues: IndexedSeq[AnyRef])
+final case class UDTValue(metaData: CassandraRowMetadata, columnValues: IndexedSeq[AnyRef])
   extends ScalaGettableData {
 
-  private var metaData_ : Option[CassandraRowMetadata] = None
+  def this(columnNames: IndexedSeq[String], columnValues: IndexedSeq[AnyRef]) =
+    this(CassandraRowMetadata.fromColumnNames(columnNames), columnValues)
 
-  lazy val metaData : CassandraRowMetadata = metaData_ match {
-    case Some(x) => x
-    case None => CassandraRowMetadata.fromColumnNames(columnNames)
-  }
-
-  def this(metaData: CassandraRowMetadata, columnValues: IndexedSeq[AnyRef]) = {
-    this(metaData.columnNames, columnValues)
-    this.metaData_ = Some(metaData)
-  }
+  def columnNames: IndexedSeq[String] = metaData.columnNames
 
   override def productArity: Int = columnValues.size
   override def productElement(i: Int) = columnValues(i)
+
+  def unapply(t: UDTValue): Some[(IndexedSeq[String],IndexedSeq[AnyRef])] =
+    Some((t.metaData.columnNames,t.columnValues))
 }
 
 object UDTValue {
@@ -46,6 +42,6 @@ object UDTValue {
     }
   }
 
-  def apply(metaData: CassandraRowMetadata, columnValues: IndexedSeq[AnyRef]): UDTValue =
-    new UDTValue(metaData, columnValues)
+  def apply(columnNames: IndexedSeq[String], columnValues: IndexedSeq[AnyRef]): UDTValue =
+    new UDTValue(columnNames, columnValues)
 }
--- a/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala
+++ b/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala
@@ -6,10 +6,25 @@ import scala.reflect.runtime.universe._
 import com.datastax.driver.core.{UDTValue => DriverUDTValue}
 import com.datastax.spark.connector.types.NullableTypeConverter
 
-final case class UDTValue(metaData: CassandraRowMetadata, columnValues: IndexedSeq[AnyRef])
+final case class UDTValue(columnNames: IndexedSeq[String], columnValues: IndexedSeq[AnyRef])
   extends ScalaGettableData {
+
+  @volatile private var metaData_ : Option[CassandraRowMetadata] = None
+
+  def this(metaData: CassandraRowMetadata, columnValues: IndexedSeq[AnyRef]) = {
+    this(metaData.columnNames, columnValues)
+    this.metaData_ = Some(metaData)
+  }
+
   override def productArity: Int = columnValues.size
   override def productElement(i: Int) = columnValues(i)
+
+  override def metaData: CassandraRowMetadata = {
+    metaData_ match {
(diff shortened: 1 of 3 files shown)


spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/UserDefinedType.scala
```diff
@@ -32,6 +33,7 @@ case class UserDefinedType(name: String, columns: IndexedSeq[UDTFieldDef])
   def cqlTypeName = name
 
   val fieldConverters = columnTypes.map(_.converterToCassandra)
+  val metadata = CassandraRowMetadata.fromColumnNames(columnNames)
```
jtgrabowski, line 36: should be private

jtgrabowski (commented):
Hi! Thank you for the contribution. 
Could you remove the additional commit from Russ? Also could you open a second PR against b2.5 with cherry-picked commit?
We also need to have passing tests to merge.

gowa:
@jtgrabowski , I've rebase and force-pushed the code to remove the commit from Russ.
Also, I've added the 'private' keyword and made UDTValue changes more succinct.

I ran sbt test:

`[info] ScalaTest                                                                                                                                                                                                                         [info] Run completed in 16 seconds, 202 milliseconds.                                                                                                                                                                                     [info] Total number of tests run: 335                                                                                                                                                                                                     [info] Suites: completed 31, aborted 0                                                                                                                                                                                                   [info] Tests: succeeded 335, failed 0, canceled 0, ignored 0, pending 0                                                                                                                                                                   [info] All tests passed.                                                                                                                                                                                                                  [info] Passed: Total 493, Failed 0, Errors 0, Passed 493                                                                                                                                                                        [success] Total time: 87 s, completed Apr 9, 2021 4:06:02 PM`

Do I need to run integration tests?

gowa:
@jtgrabowski , travis-ci build was successful: https://travis-ci.com/github/gowa/spark-cassandra-connector/builds/222710219

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala
```diff
@@ -6,24 +6,22 @@ import scala.reflect.runtime.universe._
 import com.datastax.driver.core.{UDTValue => DriverUDTValue}
 import com.datastax.spark.connector.types.NullableTypeConverter
 
-final case class UDTValue(columnNames: IndexedSeq[String], columnValues: IndexedSeq[AnyRef])
+final case class UDTValue(metaData: CassandraRowMetadata, columnValues: IndexedSeq[AnyRef])
```
gowa, line 9: @jtgrabowski , as I changed the case class fields, the constructor and the default apply and unapply methods changed.
Also the getter for columnNames get disappeared. This breaks compatibility. I've seen the usage example for this class in the docs: UDTValue.fromMap . Nothing prohibits this class from being instantiated directly.

Please let me know, if we should take care of compatibility here.
Maybe Option[CassandraRowMetadata] as 3rd parameter could help here.

Thanks

gowa, line 9: @jtgrabowski for the sake of preserving API compatibility, if needed, I can suggest the following change:

````suggestion
final case class UDTValue(columnNames: IndexedSeq[String], columnValues: IndexedSeq[AnyRef])
  extends ScalaGettableData {

  // caching CassandraRowMetadata instance
  @volatile private var metaData_ : Option[CassandraRowMetadata] = None

  //alternative constructor
  def this(metaData: CassandraRowMetadata, columnValues: IndexedSeq[AnyRef]) = {
    this(metaData.columnNames, columnValues)
    this.metaData_ = Some(metaData)
  }

...

  override def metaData: CassandraRowMetadata = {
    metaData_ match {
      case Some(x) => x
      // I think it's ok to assign without double checked locking here:
      // 1. earlier each call of metadata() method resulted in new CassandraRowMetadata instance => returning several different instances should be fine
      // 2. if the caller did not supply CassandraRowMetadata, they are not concerned about performance
      case None => val x = CassandraRowMetadata.fromColumnNames(columnNames); metaData_ = Some(x); x
    }
  }
}

object UDTValue {

...
  // alternative apply method
  def apply(metaData: CassandraRowMetadata, columnValues: IndexedSeq[AnyRef]): UDTValue =
    new UDTValue(metaData, columnValues)
}
````

Comment (jtgrabowski) [reply spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala:9]: It's a good idea, let's stay with the original signature (```final case class UDTValue(columnNames: IndexedSeq[String], columnValues: IndexedSeq[AnyRef])```).
Instead of a volatile var use a lazy val and we should be good to go.

Comment (gowa) [reply spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala]: Hi @jtgrabowski ! I've update the pull request with the corresponding changes - both this one and b2.5 one either.
I've run test and it:test  - they were fine. My internal perf test was ok either.

File: spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala [hunk 3]
```diff
@@ -8,10 +8,21 @@ import com.datastax.spark.connector.types.NullableTypeConverter
 
 final case class UDTValue(columnNames: IndexedSeq[String], columnValues: IndexedSeq[AnyRef])
   extends ScalaGettableData {
+
```
jtgrabowski, line 11: We don't need metaData_ here. How about doing it this way:
```suggestion
final case class UDTValue(columnNames: IndexedSeq[String], columnValues: IndexedSeq[AnyRef])
  extends ScalaGettableData {

  private var metaData_ : Option[CassandraRowMetadata] = None

  lazy val metaData : CassandraRowMetadata = metaData_ match {
    case Some(x) => x
    case None => CassandraRowMetadata.fromColumnNames(columnNames)
  }

  def this(metaData: CassandraRowMetadata, columnValues: IndexedSeq[AnyRef]) = {
    this(metaData.columnNames, columnValues)
    this.metaData_ = Some(metaData)
  }

  override def productArity: Int = columnValues.size
  override def productElement(i: Int) = columnValues(i)
}

object UDTValue {

  def fromJavaDriverUDTValue(value: DriverUDTValue): UDTValue = {
    val fields = value.getType.getFieldNames.map(f => toName(f)).toIndexedSeq
    val values = fields.map(GettableData.get(value, _))
    UDTValue(fields, values)
  }

  def fromMap(map: Map[String, Any]): UDTValue =
    new UDTValue(map.keys.toIndexedSeq, map.values.map(_.asInstanceOf[AnyRef]).toIndexedSeq)

  val TypeTag = implicitly[TypeTag[UDTValue]]
  val Symbol = typeOf[UDTValue].asInstanceOf[TypeRef].sym

  implicit object UDTValueConverter extends NullableTypeConverter[UDTValue] {
    def targetTypeTag = TypeTag
    def convertPF = {
      case x: UDTValue => x
    }
  }

  def apply(metaData: CassandraRowMetadata, columnValues: IndexedSeq[AnyRef]): UDTValue =
    new UDTValue(metaData, columnValues)
}
```

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/UserDefinedType.scala
```diff
@@ -32,6 +32,7 @@ case class UserDefinedType(name: String, columns: IndexedSeq[UDTFieldDef])
   def cqlTypeName = name
 
   val fieldConverters = columnTypes.map(_.converterToCassandra)
+  private val metadata = CassandraRowMetadata.fromColumnNames(columnNames)
```
jtgrabowski, line 35: let's make it lazy

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala
gowa: @jtgrabowski , yes, I wanted initially to go that way, but this will change the auto-generated 'unapply' method for the case class.

earlier one could have used:
```
p match {
 case UDTValue(columnName,columnValues) => ...
}
```
and if we change the case class, then the method allowing match/case to work will be also auto changed.

It seems that in scala it is possible to implement our own 'unapply' for the case class, e.g. to make it identical to what we had before this pull req. If you find it's reasonable to take care about 'unapply', I'll play with the custom 'unapply' method. Please, let me know.

jtgrabowski: Yes, let's add the corresponding `unapply` method.

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/UserDefinedType.scala
gowa: @jtgrabowski I'll push the changes soon. Anyway, it seems that it is not possible to achieve the full binary compatibility with this change. I've recalled the 'copy' method of the case classes. its signature will also change. :(

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala
gowa: @jtgrabowski I'll push the changes soon. Anyway, it seems that it is not possible to achieve the full binary compatibility with this change. I've recalled the 'copy' method of the case classes. its signature will also change. :(

jtgrabowski: that's ok, let's change it anyway.

gowa:
travis build now fails because of https://github.com/databricks/sbt-spark-package/issues/50 ...
https://travis-ci.com/github/gowa/spark-cassandra-connector/jobs/505364029

sbt.ResolveException: unresolved dependency: org.spark-packages#sbt-spark-package;0.2.5: not found

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/UDTValue.scala
gowa: done

spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/types/UserDefinedType.scala
gowa: done

jtgrabowski (approved):

PR #1333, opened 2021-11-13
[SPARKC-670] Make SCC compatible with Spark 3.2.0

alexott:
Due the changes in internals of Spark, SCC 3.1.0 is not compatible with Spark 3.2.0 in
some areas:
- functions `writetime` & `ttl` are crashing
- `DirectJoin` is crashing as well

This PR fixes that problem, and also update some dependencies to match Spark 3.2.0, and
also bumps Java Driver version to 4.13.0

- [X] I have a ticket in the [OSS JIRA](https://datastax-oss.atlassian.net/projects/SPARKC)
- [X] I have performed a self-review of my own code
- [X] Locally all tests pass (make sure tests fail without your patch), some integration tests are failing looks like the problem with CCM start timeout.

PR #1342, opened 2022-04-11
Compatibility with Spark 3.2

bbenshalom:
Continuing @alexott 's work from [1333](https://github.com/datastax/spark-cassandra-connector/pull/1333). There were tests failing and I believe the reason was the new adaptive query execution, which was [enabled by default](https://issues.apache.org/jira/browse/SPARK-33679) in 3.2.

I resolved a few tests by simply handling the case of `AdaptiveSparkPlanExec` (the `BatchScanExec` object was nested within it), and another test by including `spark_catalog` as a valid cassandra catalog name (not 100% sure this is the desired handling of this test, let me know what you think).

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ Currently the following branches are actively supported: 3.1.x ([master](https:/
 
 | Connector | Spark         | Cassandra | Cassandra Java Driver | Minimum Java Version | Supported Scala Versions |
 | --------- | ------------- | --------- | --------------------- | -------------------- | -----------------------  |
+| 3.2       | 3.2           | 2.1.5*, 2.2, 3.x, 4.0 | 4.13             | 8             | 2.12                     |
 | 3.1       | 3.1           | 2.1.5*, 2.2, 3.x, 4.0 | 4.12             | 8             | 2.12                     |
 | 3.0       | 3.0           | 2.1.5*, 2.2, 3.x, 4.0 | 4.12             | 8             | 2.12                     |
 | 2.5       | 2.4           | 2.1.5*, 2.2, 3.x, 4.0 | 4.12             | 8             | 2.11, 2.12               |
--- a/connector/src/it/scala/com/datastax/bdp/spark/search/SearchAnalyticsIntegrationSpec.scala
+++ b/connector/src/it/scala/com/datastax/bdp/spark/search/SearchAnalyticsIntegrationSpec.scala
@@ -498,27 +498,6 @@ class SearchAnalyticsIntegrationSpec extends SparkCassandraITFlatSpecBase with D
     testType("m = '10000000-0000-0000-0000-000000000000'", """"m:10000000\\-0000\\-0000\\-0000\\-000000000000"""")
   }
 
-  it should "work with weird column names" in dseOnly {
-    val df: DataFrame = spark
-      .read
-      .cassandraFormat(weirdTable, ks)
-      .load()
-
-    val filtered = df.filter(df("key") === 1)
-      .filter(df("~~funcolumn") === "hello")
-      .filter(df("MixEdCol") === "world")
-
-    val whereClause = filtered.getUnderlyingCqlWhereClause()
-    val predicates = whereClause.predicates.head
-    val values: String = whereClause.values.head.toString
-    predicates should include("""solr_query""")
-    values should include("funcolumn")
-    values should include("MixEdCol")
-    val rows = filtered.collect
-    rows.head.getAs[String]("~~funcolumn") should be("hello")
-    rows.head.getAs[String]("MixEdCol") should be("world")
-  }
-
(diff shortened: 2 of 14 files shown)


jtgrabowski:
Hi @bbenshalom there is a compilation error in the CI:
```
 [error] /home/runner/work/spark-cassandra-connector/spark-cassandra-connector/connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataSourceSpec.scala:289:29: wrong number of arguments for pattern org.apache.spark.sql.execution.datasources.v2.BatchScanExec(output: Seq[org.apache.spark.sql.catalyst.expressions.AttributeReference],scan: org.apache.spark.sql.connector.read.Scan,runtimeFilters: Seq[org.apache.spark.sql.catalyst.expressions.Expression])
[error]         case a@BatchScanExec(_, _: CassandraInJoin) => a
[error]                             ^
[error] one error found
[error] (connector / IntegrationTest / compileIncremental) Compilation failed
[error] Total time: 23 s, completed Apr 11, 2022 6:49:06 AM
```

bbenshalom:
Hi @jtgrabowski , fixed. Any way for me to re-run the tests workflow myself?

jtgrabowski:
You can run the integration tests locally with `sbt/sbt it:test`, see [this page](https://github.com/datastax/spark-cassandra-connector/blob/master/doc/developers.md#getting-started) for details. Once this PR is merged, you'll be free to run the CI tests on your own. Btw. could you please sign the DataStax Contribution License Agreement https://cla.datastax.com/ ?

bbenshalom:
I signed the license agreement

bbenshalom:
@jtgrabowski The test that's failing in the CI is skipped for me locally ("skipped because not DSE"), I couldn't find in the contribution tips how to define a "DSE mode"... Can you please advise or point me to the right documentation?

jtgrabowski:
Try 
```
export CCM_CASSANDRA_VERSION=5.1.24
sbt/sbt 'it:testOnly com.datastax.bdp.spark.search.SearchAnalyticsIntegrationSpec'
```

jtgrabowski:
It may seam to hang for a couple of minutes. The test needs to download DSE zip. The zip is downloaded to /tmp/ccm-xyz.

bbenshalom:
@jtgrabowski it seems that the current failing test has to do with a [change](https://spark.apache.org/docs/latest/sql-migration-guide.html) in spark 3.2:

> In Spark 3.2, DataFrameNaFunctions.replace() no longer uses exact string match for the input column names, to match the SQL syntax and support qualified column names. Input column name having a dot in the name (not nested) needs to be escaped with backtick `. Now, it throws AnalysisException if the column is not found in the data frame schema. It also throws IllegalArgumentException if the input column name is a nested column. In Spark 3.1 and earlier, it used to ignore invalid input column name and nested column name.

In other words, for the column name `~~funcolumn`, the dataframe filter generated is ``~~funcolumn``. I could just remove backticks from the column names in the code, but it doesn't seem right, since the intention of the Spark change seems to be to avoid "weird" column names. Do you think it would be better to simply remove the test and not support this functionality?

jtgrabowski:
I'm hesitant on just dropping the support for this. Could we match the column name with the backticks instead?

bbenshalom:
@jtgrabowski I tried to match the column names with the backticks while building the cassandra filters, and indeed the execution plan was displayed correctly. However, the test then failed on the collect() method of the DataFrame, again because of the same mismatch. The filter already has the backticks in it when it is passed to `CassandraScanBuilder` (called by spark code)...

Spent quite a bit of time on this and couldn't find a way around it. Open to suggestions...

jtgrabowski:
I didn't have time to look into it yesterday. I'll try to help next week.

jtgrabowski:
@bbenshalom here is a PR that should fix the offending test: https://github.com/bbenshalom/spark-cassandra-connector/pull/1. Please take a look when you get a chance.
I also created this: https://datastax-oss.atlassian.net/browse/SPARKC-677 as the "weird column names" are not fully supported. The problems appear with the older Spark versions too, so I don't think it should block us from moving forward with this PR.

bbenshalom:
@jtgrabowski thanks for the PR. Do you have any plan as to when we can merge & release a new version (even a release candidate)?

jtgrabowski (approved):

jtgrabowski:
We need to do some behind-the-scenes testing before we can release. I'll kick off the performance tests and let you know.

jtgrabowski:
@bbenshalom @alexott thank you for your contribution! This is now merged!

bbenshalom:
@jtgrabowski thank you for your patience and great support! Was a pleasure!

jtgrabowski:
perf results are ok, the release is scheduled for next week :+1:

PR #1353, opened 2022-12-21
SPARKC-695: Fix projection collapse on CassandraDirectJoinStrategy

viirya:
# Description

`CassandraDirectJoinStrategy` doesn't consider the case that there are projections of direct join outputs. For example, `address` is a complex type field in cassandra table which is joined with other table:

```
left.join(right, left("id") === right("id"))
  .select($"address.*")
  .select($"street", $"city")
```

The extractors on `address` will be collapsed. So while building alias map in `CassandraDirectJoinStrategy`, it will fail by:

```
scala.MatchError: address#6337.street AS _extract_street#6354 (of class org.apache.spark.sql.catalyst.expressions.Alias)                                                                                                      
[info]   at org.apache.spark.sql.cassandra.execution.CassandraDirectJoinStrategy$.$anonfun$aliasMap$1(CassandraDirectJoinStrategy.scala:316)         
```

Beside fixing this by providing correct alias map, `reorderPlan` is also incorrect for such case. `reorderPlan` now takes direct join output as final output that replaces original plan's output. However, direct output only contains cassandra table scan outputs and other joining side's output. For above example, the outputs are ['id', 'id', 'address'] so it don't match with original plan's output ['street', 'city'].

This patch also fixes incorrect output issue and verifies it by the added integration test.

# How Has This Been Tested?

Added an integration test and verified it locally.

# Checklist:

### DIFF
--- a/connector/src/it/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinSpec.scala
+++ b/connector/src/it/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinSpec.scala
@@ -98,7 +98,6 @@ class CassandraDirectJoinSpec extends SparkCassandraITFlatSpecBase with DefaultC
                |})""".stripMargin)
         },
         Future {
-          session.execute(s"CREATE TYPE $ks.address (street text, city text, residents set<frozen<tuple<text, text>>>) ")
           session.execute(s"CREATE TYPE $ks.user (address frozen <address>) ")
           session.execute(s"CREATE TABLE $ks.members (id text, user frozen <user>, PRIMARY KEY (id))")
           session.execute(
--- a/connector/src/it/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinSpec.scala
+++ b/connector/src/it/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinSpec.scala
@@ -96,8 +96,6 @@ class CassandraDirectJoinSpec extends SparkCassandraITFlatSpecBase with DefaultC
                |  city: 'New Orleans',
                |  residents:{('sundance', 'dog'), ('cara', 'dog')}
                |})""".stripMargin)
-        },
-        Future {
           session.execute(s"CREATE TYPE $ks.user (address frozen <address>) ")
           session.execute(s"CREATE TABLE $ks.members (id text, user frozen <user>, PRIMARY KEY (id))")
           session.execute(
--- a/connector/src/it/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinSpec.scala
+++ b/connector/src/it/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinSpec.scala
@@ -622,6 +622,14 @@ class CassandraDirectJoinSpec extends SparkCassandraITFlatSpecBase with DefaultC
     left.join(right, left("id") === right("id"))
   }
 
+  it should "work with field extractor after join" in compareDirectOnDirectOff{ spark =>
+    val left = spark.createDataset(Seq(IdRow("test")))
+    val right = spark.read.cassandraFormat("location", ks).load()
+    left.join(right, left("id") === right("id"))
+      .select($"address.*")
+      .select($"street", $"city")
+  }
+
   it should "work on a timestamp PK join" in compareDirectOnDirectOff { spark =>
     val left = spark.createDataset(
       (1 to 100).map(value => TimestampRow(new Timestamp(value.toLong)))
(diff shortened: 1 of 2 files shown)


viirya:
cc @RussellSpitzer

ErickRamirezAU:
@viirya thanks for submitting the PR! If you haven't already:

1. Please [create a Jira ticket](https://datastax-oss.atlassian.net/projects/SPARKC/issues) with details of the issue. Otherwise, please link an existing ticket.
2. Please sign the [Contributor License Agreement](https://cla.datastax.com/).

Thanks again. Cheers! 🍻

connector/src/main/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinStrategy.scala
```diff
@@ -5,7 +5,7 @@ import com.datastax.spark.connector.util.Logging
 import org.apache.spark.sql.{SparkSession, Strategy}
 import org.apache.spark.sql.cassandra.{AlwaysOff, AlwaysOn, Automatic, CassandraSourceRelation}
 import org.apache.spark.sql.cassandra.CassandraSourceRelation._
-import org.apache.spark.sql.catalyst.expressions.{Alias, AttributeReference, ExprId, Expression, NamedExpression}
+import org.apache.spark.sql.catalyst.expressions.{Alias, Attribute, AttributeReference, ExprId, Expression, ExtractValue, NamedExpression}
```
RussellSpitzer, line 8: Is ExtractValue used?

connector/src/main/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinStrategy.scala
```diff
@@ -252,19 +255,25 @@ object CassandraDirectJoinStrategy extends Logging {
...
       case ProjectExec(projectList, child) =>
+        val attrMap = directJoin.output.map {
+          case attr => attr.exprId -> attr
+        }.toMap
+
         val aliases = projectList.collect {
-          case a @ Alias(child: AttributeReference, _) => (child.toAttribute.exprId, a)
+          case a @ Alias(child, _) =>
+            val newAliasChild = child.transform {
+              case attr: Attribute => attrMap.getOrElse(attr.exprId, attr)
```
RussellSpitzer, line 265: So this just descends till we get the root correct?

connector/src/it/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinSpec.scala
```diff
@@ -622,6 +622,14 @@ class CassandraDirectJoinSpec extends SparkCassandraITFlatSpecBase with DefaultC
     left.join(right, left("id") === right("id"))
   }
 
+  it should "work with field extractor after join" in compareDirectOnDirectOff{ spark =>
```
RussellSpitzer, line 625: I would add just one more test with a doubly nested struct, I don't think it has to be full integration test like this, just something to show the getAlias method works on a very nested struct

connector/src/main/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinStrategy.scala
viirya: Yea, it finds and replaces it by the attribute from direct join node.

connector/src/main/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinStrategy.scala
viirya: No, added during development. I'll remove it.

connector/src/it/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinSpec.scala
viirya: Added a doubly nested test.

viirya:
```
[info]   com.datastax.oss.driver.api.core.servererrors.InvalidQueryException: A user type with name 'address' already exists
```

I didn't see this error locally. Actually I didn't have the second 'address' type definition when I was to add the doubly nested test case, but got an error indicating there is no 'address' type. I'm removing it now see if it can pass CI.

RussellSpitzer (approved):
Lgtm +1

jtgrabowski (approved):

jtgrabowski:
@viirya thank you for the contribution!

viirya:
Thanks @RussellSpitzer @jtgrabowski !

PR #1356, opened 2023-05-07
SPARKC-700: Fix complex field extractor after join on CassandraDirectJoinStrategy

viirya:
# Description

We fixed projection collapse on `CassandraDirectJoinStrategy` in https://datastax-oss.atlassian.net/browse/SPARKC-695. But one incorrect assumption during collecting alias mappings is there is only one attribute under the extractor. If users build a struct before the join, and try to extract a field with an extractor, the assert will fail:

The incorrect assumption was fixed in this patch by collecting all attributes under aliased extractor.

Fixes: [SPARKC-700](https://datastax-oss.atlassian.net/browse/SPARKC-700)

# How Has This Been Tested?

# Checklist:

[SPARKC-700]: https://datastax-oss.atlassian.net/browse/SPARKC-700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/connector/src/it/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinSpec.scala
+++ b/connector/src/it/scala/org/apache/spark/sql/cassandra/execution/CassandraDirectJoinSpec.scala
@@ -649,6 +649,15 @@ class CassandraDirectJoinSpec extends SparkCassandraITFlatSpecBase with DefaultC
       .select($"street", $"city")
   }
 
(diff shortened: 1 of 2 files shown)


viirya:
cc @RussellSpitzer

RussellSpitzer (approved):
Looks good to me

jtgrabowski:
Thank you @viirya !

PR #1365, opened 2023-12-20
Fix for SPARKC-699: Projection of Sub Elements

nob13:
# Description

Spark Cassandra Connector was crashing with ClassCastExceptions when selecting UDT fields on pushed down Selections.

The cassandra calls always select whole fields. To my knowledge, cassandra has no possibility to select UDT fields directly. However the the decoder to spark expects only the selected UDT field to be presented. During decoding, it always decodes the first element (which may be of a different type) and thus failing with class cast exception, or in case of same types with a wrong value.

- UnsafeRowReader has now an extra decoder step which fulfils the decoding of projected UDT fields.
- The decoding is done by a new class `UdtProjectionDecoder`
- Added a new Cluster Mode "EXISTING" so that I can run integration tests on a locally running cassandra instance without relying on ccm.

Fixes: [SPARKC-699](https://datastax-oss.atlassian.net/browse/SPARKC-699)

# How Has This Been Tested?

There is a new Integration Test `CassandraDataFrameSelectUdtSpec`

# Checklist:

[SPARKC-699]: https://datastax-oss.atlassian.net/browse/SPARKC-699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataFrameSelectUdtSpec.scala
+++ b/connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataFrameSelectUdtSpec.scala
@@ -3,6 +3,7 @@ package com.datastax.spark.connector.sql
 import com.datastax.spark.connector.SparkCassandraITFlatSpecBase
 import com.datastax.spark.connector.cluster.DefaultCluster
 import com.datastax.spark.connector.cql.CassandraConnector
+import org.apache.spark.sql.Row
 import org.apache.spark.sql.functions.col
 import org.scalatest.Matchers
 import org.scalatest.concurrent.Eventually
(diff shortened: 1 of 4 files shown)


nob13:
Please Note:

- I could only run single integration tests using the cluster mode "EXISTING". I would rely on automatic tests of CI.
- I am not sure if this is the best way to solve the problem, as I am not to deep in the source.
- If this new cluster mode is of no interest, I can also remove it.

Happy holiday season to everyone. I'll be back in January.

nob13:
@jtgrabowski hi, is there any Interest of Datastax in this patch?

jtgrabowski:
Hello @nob13 ! It's possible to select just a part of an UDT in Cassandra. E.g. here https://docs.datastax.com/en/cql-oss/3.3/cql/cql_using/useCreateUDT.html you could do `select basics.nationality from cycling.cyclist_stats ;`
Do you think you could push down the selection to Cassandra?

nob13:
hi @jtgrabowski  I think it's possible to push down this kind of selects in a separate ticket. However this PR addresses the more generic case where you have some collecrtion of UDTs, e.g.

```
CREATE TYPE cycling.basic_info (
  birthday timestamp,
  nationality text,
  weight text,
  height text
);

CREATE TABLE cycling.cyclist_stats ( id uuid PRIMARY KEY, lastname text, basics FROZEN<basic_info>, multiple_basics LIST<FROZEN<basic_info>>);  
```

the following queries are allowed within Spark
```sql
-- Valid but crashes right now and could be pushed down
select basics.birthday from cycling.cyclist_stats; 
-- Valid but crashes right now and as far as I know this cannot be pushed down
select multiple_basics.birthday from cycling.cyclist_stats;
```

The PR fixes both crashes but adds no further pushing down.

connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataFrameSelectUdtSpec.scala
```diff
@@ -0,0 +1,105 @@
...
+
+    session.execute(
+      s"""CREATE TYPE ${ks}.embedded(
+        |        a TEXT,
+        |        b INT
+        |    )""".stripMargin
+    )
+
+    session.execute(
+      s"""CREATE TABLE ${ks}.crash_test(
```
jtgrabowski, line 25: I there a way tuples could be selected as well?

jtgrabowski (commented):
Thanks! This looks good, please see the minor comments.
Also, please sign https://cla.datastax.com/

test-support/src/main/scala/com/datastax/spark/connector/ccm/mode/ExistingModeExecutor.scala
```diff
@@ -0,0 +1,20 @@
+package com.datastax.spark.connector.ccm.mode
+import com.datastax.spark.connector.ccm.CcmConfig
+
+import java.nio.file.{Files, Path}
+
+private[ccm] class ExistingModeExecutor(val config: CcmConfig) extends ClusterModeExecutor {
```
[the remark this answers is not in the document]
nob13: done

connector/src/it/scala/com/datastax/spark/connector/sql/CassandraDataFrameSelectUdtSpec.scala
nob13: Added Support for tuples as well. They were also broken.

nob13:
@jtgrabowski I signed the CLA :)

jtgrabowski (approved):

jtgrabowski:
Thank you @nob13 !

PR #1366, opened 2024-04-22
SPARKC-706: Add basic support for Cassandra vectors

jacek-lewandowski:
# Description

Describe the problem, or state of the project that this patch fixes. Explain
why this is a problem if this isn't obvious.

Example: 
  "When I read from tables with 3 INTS I get a ThreeIntException(). This is a problem because I often want to read from a table with three integers."

How the fix is accomplished, were new parameters or classes added? Why did you
pursue this particular fix?

Example: "I removed the incorrect assertion which would throw the ThreeIntException. This exception was incorrectly added and the assertion is not actually needed."

Fixes: [Put JIRA Reference HERE](https://datastax-oss.atlassian.net/projects/SPARKC)

# How Has This Been Tested?

# Checklist:


driver/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
```diff
@@ -768,6 +769,29 @@ object TypeConverter {
...
+    val elemConverter = implicitly[TypeConverter[T]]
+
+    implicit def elemTypeTag: TypeTag[T] = elemConverter.targetTypeTag
+
+    @transient
+    lazy val targetTypeTag = {
+      implicitly[TypeTag[CqlVector[T]]]
+    }
+
+    def newCollection(items: Iterable[Any]): java.util.List[T] = {
```
jtgrabowski, line 782: Could be private

driver/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
```diff
@@ -768,6 +769,29 @@ object TypeConverter {
...
+      val buf = new java.util.ArrayList[T](dimension)
```
jtgrabowski, line 783: Why java collection here?

jtgrabowski (commented):
Great work! :)

Please have a look at the minor comments and add a note with a large font in the main README stating that the vector type is now supported. Additionally, please add a document showcasing how it can be used (examples taken from tests should be okay here).

connector/src/it/scala/com/datastax/spark/connector/rdd/typeTests/VectorTypeTest.scala
```diff
@@ -0,0 +1,238 @@
...
+  override def cassandraOnly(f: => Unit): Unit = super.cassandraOnly(f)
+
+  override lazy val conn = CassandraConnector(sparkConf)
+
+  val VectorTable = "vectors"
+
+  def createVectorTable(session: CqlSession, table: String): Unit = {
+    session.execute(
+      s"""CREATE TABLE IF NOT EXISTS $ks.$table (
+         |  id INT PRIMARY KEY,
```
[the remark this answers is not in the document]
jacek-lewandowski: Perhaps, why do you ask?

driver/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
jacek-lewandowski: This collection is passed to `CqlVector.getInstance` which expects either Java `List` or array. If we use here a Scala collection and then let `JavaConverters` adjust that, we will have to create one extra wrapper which is a waste in this case.

connector/src/it/scala/com/datastax/spark/connector/rdd/typeTests/VectorTypeTest.scala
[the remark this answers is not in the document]
jtgrabowski: Wondering if that's supported.

doc/2_loading.md
```diff
@@ -184,7 +184,49 @@ val street = address.getString("street")
 val number = address.getInt("number")
 ```
 
-[//]: # (TODO loading vectors)
+### Reading vectors
```
jtgrabowski, line 187: Should we add similar docs to https://github.com/datastax/spark-cassandra-connector/blob/master/doc/14_data_frames.md ?

jtgrabowski (approved):
LGTM!

PR #1367, opened 2024-05-07
SPARKC-710: Update component versions

jacek-lewandowski:
# Description

Describe the problem, or state of the project that this patch fixes. Explain
why this is a problem if this isn't obvious.

Example: 
  "When I read from tables with 3 INTS I get a ThreeIntException(). This is a problem because I often want to read from a table with three integers."

How the fix is accomplished, were new parameters or classes added? Why did you
pursue this particular fix?

Example: "I removed the incorrect assertion which would throw the ThreeIntException. This exception was incorrectly added and the assertion is not actually needed."

Fixes: [Put JIRA Reference HERE](https://datastax-oss.atlassian.net/projects/SPARKC)

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/test-support/src/main/scala/com/datastax/spark/connector/ccm/CcmConfig.scala
+++ b/test-support/src/main/scala/com/datastax/spark/connector/ccm/CcmConfig.scala
@@ -2,7 +2,7 @@ package com.datastax.spark.connector.ccm
 
 import java.io.{File, IOException, InputStream}
 import java.net.InetSocketAddress
-import java.nio.file.{Files, Path, Paths, StandardCopyOption}
+import java.nio.file.{Files, Paths, StandardCopyOption}
 
 import com.datastax.oss.driver.api.core.Version
 import com.datastax.spark.connector.ccm.CcmConfig._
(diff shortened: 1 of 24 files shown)


jtgrabowski (approved):
awesome job @jacek-lewandowski !

test-support/src/main/scala/com/datastax/spark/connector/ccm/CcmConfig.scala
```diff
@@ -116,16 +117,16 @@ object CcmConfig {
...
+  val DSE_V6_8_5: Version = Version.parse("6.8.5")
+  val DSE_V6_8_3: Version = Version.parse("6.8.3")
+  val DSE_V6_7_0: Version = Version.parse("6.7.0")
+  val DSE_V6_0_0: Version = Version.parse("6.0.0")
+  val DSE_V5_1_0: Version = Version.parse("5.1.0")
+  val DSE_V5_0_0: Version = Version.parse("5.0.0")
 
   // C* versions
-  val V4_0_0: Version = Version.parse("4.0.0")
+  val V5_0_0: Version = Version.parse("5.0-alpha1")
```
msmygit, line 128: nit - should we use [`5.0-beta1`](https://github.com/apache/cassandra/releases/tag/cassandra-5.0-beta1) here instead?

PR #1369, opened 2024-07-30
SPARKC-712: Update metrics for read and writes via DSV2

siadat:
# Description

> Describe the problem, or state of the project that this patch fixes. Explain
> why this is a problem if this isn't obvious.
>
> Example: 
>  "When I read from tables with 3 INTS I get a ThreeIntException(). This is a problem because I often want to read from a > table with three integers."

Metrics are not updated for read and writes performed via DSV2. The values returned are all 0s.

> How the fix is accomplished, were new parameters or classes added? Why did you
> pursue this particular fix?
> 
> Example: "I removed the incorrect assertion which would throw the ThreeIntException. This exception was incorrectly added and the assertion is not actually needed."
> 
> Fixes: Put JIRA Reference

InputMetricsUpdater and OutputMetricsUpdater are used in the DSV2 read (including joins and counts) and write paths.

Fixes: [SPARKC-712](https://datastax-oss.atlassian.net/browse/SPARKC-712)

# How Has This Been Tested?

> Almost all changes and especially bug fixes will require a test to be added to either the integration or Unit Tests. Any tests added will be automatically run on travis when the pull request is pushed to github. Be sure to run suites locally as well.

A couple of integration tests were added.

# Checklist:

- [x] I have a ticket in the [OSS JIRA](https://datastax-oss.atlassian.net/projects/SPARKC): https://datastax-oss.atlassian.net/browse/SPARKC-712

[SPARKC-712]: https://datastax-oss.atlassian.net/browse/SPARKC-712

### DIFF
--- a/connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogMetricsSpec.scala
+++ b/connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogMetricsSpec.scala
@@ -1,0 +1,79 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Please see the included license file for details.
+ */
+
+package com.datastax.spark.connector.datasource
+
+import scala.collection.mutable
+import com.datastax.spark.connector._
+import com.datastax.spark.connector.cluster.DefaultCluster
+import com.datastax.spark.connector.cql.CassandraConnector
+import org.scalatest.BeforeAndAfterEach
+import com.datastax.spark.connector.datasource.CassandraCatalog
+import org.apache.spark.scheduler.{SparkListener, SparkListenerTaskEnd}
+import com.datastax.spark.connector.cql.CassandraConnector
+import org.apache.spark.sql.SparkSession
+
+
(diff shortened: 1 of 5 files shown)


siadat:
Hi @jacek-lewandowski Sorry for tagging, I was wondering if I could get your review on this PR. Thanks.

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogMetricsSpec.scala
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Please see the included license file for details.
+ */
```
michaelsembwever: this now needs to be the Apache License header, like other files.

michaelsembwever:
The GHA checks didn't run…  guessing we have a failure in `.github/workflows/main.yml`…

connector/src/it/scala/com/datastax/spark/connector/datasource/CassandraCatalogMetricsSpec.scala
siadat: Thanks for the review. I updated the header. Also rebased the branch.

PR #1371, opened 2024-11-21
PrefetchingResultSetIterator: fix bug w/r/t intermediate zero-row pages

eappere:
# Description

Zero-row pages that aren't the last can sometimes happen, for example when using "filtering". The iterator has a logic bug which make it throw an exception when encountering this.

The fix is rather simple: move maybePrefetch before the "next" call and skip such pages (if -> while).

I didn't mean to open this PR yet, but this is the same logic issue as [NODEJS-668](https://datastax-oss.atlassian.net/browse/NODEJS-668).

# How Has This Been Tested?

Unit test has been added to validate and check the presence of the fix.

# Checklist:

[NODEJS-668]: https://datastax-oss.atlassian.net/browse/NODEJS-668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/connector/src/main/scala/com/datastax/spark/connector/rdd/reader/PrefetchingResultSetIterator.scala
+++ b/connector/src/main/scala/com/datastax/spark/connector/rdd/reader/PrefetchingResultSetIterator.scala
@@ -37,7 +37,10 @@ class PrefetchingResultSetIterator(resultSet: AsyncResultSet, timer: Option[Time
   }
 
   private def maybePrefetch(): Unit = {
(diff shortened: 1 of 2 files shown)


eappere:
Hello, I didn't mean to open this PR just yet (hence the lack of its own JIRA ticket), I'll re-close the issue and reopen it after properly validating the change & opening a JIRA ticket on your tracker.

PR #1378, opened 2025-04-22
Add .asf.yaml

michaelsembwever:
https://cwiki.apache.org/confluence/display/INFRA/Git+-+.asf.yaml+features
And remove a few dangling copyright headers that were missed in 39231410157c50e30ff9c232f114a0eb2d97af5f

 patch by Mick Semb Wever; reviewed by xxx for CASSANALYTICS-15

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,9 +3,7 @@ name: CI
 on:
   # Triggers the workflow on push or pull request events but only for the trunk branch
   push:
-    branches: [ b2.5, b3.0, trunk]
   pull_request:
-    branches: [ b2.5, b3.0, trunk]
 
   workflow_dispatch:
 
@@ -16,15 +14,13 @@ jobs:
       fail-fast: false
       matrix:
         scala:  [2.12.19, 2.13.13]
-        db-version: [3.11.17, 4.0.12, 4.1.4, 5.0-beta1, dse-6.8.44]
+        db-version: [3.11.19, 4.0.17, 4.1.8, 5.0.4, dse-6.8.44]
 
     steps:
       - uses: actions/checkout@v4
 
-      - name: ccm pip installation
-        uses: BSFishy/pip-action@v1
-        with:
-          packages: git+https://github.com/riptano/ccm.git@d74db63d75112908a77b6c80757df9343fdc3338
+      - name: Install ccm via pip
+        run: pip install git+https://github.com/apache/cassandra-ccm.git@trunk
 
       - name: Setup Java
         uses: actions/setup-java@v4
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,11 +1,11 @@
 name: CI
 
 on:
-  # Triggers the workflow on push or pull request events but only for the master branch
+  # Triggers the workflow on push or pull request events but only for the trunk branch
   push:
-    branches: [ b2.5, b3.0, master]
+    branches: [ b2.5, b3.0, trunk]
   pull_request:
-    branches: [ b2.5, b3.0, master]
+    branches: [ b2.5, b3.0, trunk]
 
   workflow_dispatch:
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,6 +20,8 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Install ccm via pip
+        # if cassandra-ccm's trunk breaks this CI, please file a report,
+        #  and temporarily switch this to @cassandra-test or @<sha> where sha is the last known working ccm commit
         run: pip install git+https://github.com/apache/cassandra-ccm.git@trunk
 
       - name: Setup Java
--- a/README.md
+++ b/README.md
@@ -2,14 +2,14 @@
 
 *Lightning-fast cluster computing with Apache Spark&trade; and Apache Cassandra&reg;.*
 
-[![CI](https://github.com/datastax/spark-cassandra-connector/actions/workflows/main.yml/badge.svg?branch=master)](https://github.com/datastax/spark-cassandra-connector/actions?query=branch%3Amaster)
+[![CI](https://github.com/apache/cassandra-spark-connector/actions/workflows/main.yml/badge.svg?branch=trunk)](https://github.com/apache/cassandra-spark-connector/actions?query=branch%3Atrunk)
 
 ## Quick Links
 
 | What       | Where                                                                                                                                                                                                                                                                                                               |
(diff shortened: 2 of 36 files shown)


JeetKunDoug (approved):
+1 - thanks for adding the `.asf.yml` and the copyright notice cleanup.

michaelsembwever:
added another commit to fix some urls and `trunk` branch name.  also to get GHA working again for PRs against `trunk`…

michaelsembwever:
and another commit to get the GHA working…

michaelsembwever:
@JeetKunDoug , need to ask you to quickly review the two small additional commits here 🙏

.github/workflows/main.yml
```diff
@@ -16,15 +14,13 @@ jobs:
...
 
     steps:
       - uses: actions/checkout@v4
 
-      - name: ccm pip installation
-        uses: BSFishy/pip-action@v1
-        with:
-          packages: git+https://github.com/riptano/ccm.git@d74db63d75112908a77b6c80757df9343fdc3338
+      - name: Install ccm via pip
+        run: pip install git+https://github.com/apache/cassandra-ccm.git@trunk
```
JeetKunDoug, line 25: Is there a risk of pulling directly from Trunk breaking something if trunk has some breaking change? Looks like there aren't any tags in the repo to use but maybe tie to a SHA rather than trunk itself? Or maybe we can start tagging CCM?

README.md
```diff
@@ -45,17 +45,17 @@ in your Spark applications.
...
 
 Currently, the following branches are actively supported: 
-3.5.x ([master](https://github.com/datastax/spark-cassandra-connector/tree/master)),
-3.4.x ([b3.4](https://github.com/datastax/spark-cassandra-connector/tree/b3.4)),
-3.3.x ([b3.2](https://github.com/datastax/spark-cassandra-connector/tree/b3.3)),
-3.2.x ([b3.2](https://github.com/datastax/spark-cassandra-connector/tree/b3.2)),
-3.1.x ([b3.1](https://github.com/datastax/spark-cassandra-connector/tree/b3.1)),
-3.0.x ([b3.0](https://github.com/datastax/spark-cassandra-connector/tree/b3.0)) and 
-2.5.x ([b2.5](https://github.com/datastax/spark-cassandra-connector/tree/b2.5)).
+3.5.x ([trunk](https://github.com/apache/cassandra-spark-connector/tree/trunk)),
```
JeetKunDoug, line 52: We should really have a discussion about the number of branches we want to support given the move from Datastax to Apache - this seems excessive and maybe unnecessary, but I'm not sure why we currently have this many (Spark and Scala are sometimes difficult from a compatibility perspective... that said, looking at the compatibility chart, it seems like we may be able to reduce this down to 3.5.1 and 2.5 if 3.5.1 works on older Spark versions (back to 3.0, which most libraries _should_ be able to do I think).

README.md
```diff
@@ -177,7 +177,7 @@ To protect the community, all contributors are required to sign the Apache Softw
 [Tips for Developing the Spark Cassandra Connector](doc/developers.md)
 
 Checklist for contributing changes to the project:
-* Create a [SPARKC JIRA](https://datastax-oss.atlassian.net/projects/SPARKC/issues)
+* Create a [SPARKC JIRA](https://issues.apache.org/jira/projects/CASSANALYTICS)
```
JeetKunDoug: Do we want to add another Jira component before we point folks at CASSANALYTICS? I suppose we can move things over later, so maybe not a big deal.

doc/developers.md
```diff
@@ -73,11 +73,11 @@ git push origin SPARKC-9999-b3.1
 # Repeat for b3.2
```
JeetKunDoug: This should be `Repeat for b3.2, b3.3, b3.4`

JeetKunDoug (changes requested):
A few comments (one that doesn't have anything to do with the review itself, more a policy/procedure question for discussion maybe on Slack) but looking good.

.github/workflows/main.yml
michaelsembwever: there's the `cassandra-test` tag that's mean for CI.

but wouldn't we rather catch failures (even) faster ?  it's not like those making changes to cassandra-ccm will have this repo in mind… (it's usually just cassandra-dtest that's on folks' radar)

README.md
michaelsembwever: agree.   we're free to strip this down to whatever we think is maintainable, and that could well be just one non-trunk version.

README.md
michaelsembwever: pushed a fix

doc/developers.md
michaelsembwever: fixed

.github/workflows/main.yml
JeetKunDoug: So here's the question - are _we_ trying to test CCM or the Spark Connector here? I'd argue that _CCM_ changes breaking this build don't necessarily indicate anything in the connector is broken, just that something in CCM is broken, right? Or is there some case where you'd expect an update to CCM to be backward-incompatible with this library in a way that indicated the _library_ was broken?

JeetKunDoug: (Note I have had the same conversation on the Analytics side, and my plan was to try to put in a nightly build that runs against C* _trunk_ to specifically test for and catch incompatibilities faster, but we otherwise test against released versions)

michaelsembwever: Ideally it would be best to test against both the last stable version of CCM as well as its trunk.  But there are no versions of CCM (currently), and the QA stability benefit of `cassandra-test` over `trunk` is little.

Failures would be a CCM breakage, to be caught early and reported back to CCM.

A separate reason/question between the `cassandra-test` vs `trunk` choice is atomicity of the ccm used within each CI run.  The main benefit I've seen of using the `cassandra-test` tag is to ensure it's only changed in between CI runs, so we don't get mixed results.  (We don't expect ccm devs to be holding back commits to trunk and being conscious of what CI is doing, but they need to be when re-tagging `cassandra-test`.)  While I doubt anyone will be checking GHA runs from this repo before re-tagging `cassandra-test`, it might still have some benefit 🤷 

what's the verdict ?

JeetKunDoug: So why don't we do this - leave it at trunk but document in a comment if CCM breaks something it's reasonable to pin it to a previous SHA until CCM is fixed? That way we at least see the breakage but then have the option to unblock changes here while CCM gets fixed (or we just go fix it, given if it fails nobody else noticed it was broken in the first place I guess).

michaelsembwever: done.

PR #1380, opened 2025-06-23
CASSANALYTICS-68: Optimise byte array to hex string conversion

architch:
# Description
This change optimises byte array to hex string conversion 
Current implementation uses string formatting which needs to parse the format for every byte. This takes a lot of time and CPU.
New implementation uses bit operations.

Below code demonstrates the time taken by the two implementations (>600ms to <20ms for 1Million bytearray)

```
//create random byte array of 1 Million bytes
val byteArray = (1 to 1000000).map(x => ((x%128). * ( if x%2 == 0 then -1 else 1)).toByte).toArray

// Time taken by current code ~ 800ms
var start = System.currentTimeMillis();
var hexString = "0x" + byteArray.map("%02x" format _).mkString
var end = System.currentTimeMillis()
print("time taken : " + (end - start)) // time taken : 854


//Time taken by new code ~ 15 ms 
start = System.currentTimeMillis();
hexString = byteArrayToHexString(byteArray)
end = System.currentTimeMillis()
print("time taken : " + (end - start)) // time taken : 14
```	

Why pursue this particular fix?
We found lots of threads busy/occupied  parsing the format while converting the byteArray to String in our application. 

Fixes: [CASSANALYTICS-68](https://issues.apache.org/jira/browse/CASSANALYTICS-68)

# How Has This Been Tested?

Unit Tests.

# Checklist:

- [x] I have a ticket in the [JIRA](https://issues.apache.org/jira/projects/CASSANALYTICS)

### DIFF
--- a/driver/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
+++ b/driver/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
@@ -270,7 +270,18 @@ object TypeConverter {
     }
   }
 
-  private def byteArrayToString (x: Array[Byte])  = "0x" + x.map("%02x" format _).mkString
+  private val HEX_CHARS = Array(
+    '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'
+  )
+  private def byteArrayToString (x: Array[Byte]) : String = {
+    val result = new StringBuilder(x.length * 2) 
+    x.foreach { b =>
+      val v = b & 0xFF
+      result.append(HEX_CHARS(v >>> 4))
+      result.append(HEX_CHARS(v & 0x0F))
+    }
+    "0x" + result.toString()
+  }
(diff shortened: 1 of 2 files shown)


driver/src/main/scala/com/datastax/spark/connector/types/TypeConverter.scala
```diff
@@ -270,7 +270,18 @@ object TypeConverter {
     }
   }
 
-  private def byteArrayToString (x: Array[Byte])  = "0x" + x.map("%02x" format _).mkString
+  private val HEX_CHARS = Array(
+    '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'
+  )
+  private def byteArrayToString (x: Array[Byte]) : String = {
+    val result = new StringBuilder(x.length * 2)
```
nob13: Nit: Why not using `x.length * 2 + 2` and already inserting `0x` ? Then you can return result.toString()

nob13 (approved):

